PR #4, opened 2024-05-14
CVPN-555: Add support for plugins to send a reply packet

kp-mariappan-ramasamy:
## Description
Add support for plugins to send a reply packet

## Motivation and Context

Extending plugin functionality by returning a reply packet

## How Has This Been Tested?
All unit test pass.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/io/inside/tun.rs
+++ b/lightway-server/src/io/inside/tun.rs
@@ -49,6 +49,7 @@ impl InsideIO for Tun {
 impl InsideIOSendCallback<ConnectionState> for Tun {
     fn send(&self, mut buf: BytesMut, state: &mut ConnectionState) -> IOCallbackResult<usize> {
         let Some(client_ip) = state.ip else {
+            metrics::tun_rejected_packet_no_client_ip();
             // Ip address not found, dropping the packet
             return IOCallbackResult::Ok(buf.len());
         };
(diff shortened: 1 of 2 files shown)


lightway-server/src/io/inside/tun.rs
```diff
@@ -48,14 +47,8 @@ impl InsideIO for Tun {
...
 impl InsideIOSendCallback<ConnectionState> for Tun {
-    fn send(
-        &self,
-        session_id: SessionId,
-        mut buf: BytesMut,
-        state: &mut ConnectionState,
-    ) -> IOCallbackResult<usize> {
+    fn send(&self, mut buf: BytesMut, state: &mut ConnectionState) -> IOCallbackResult<usize> {
         let Some(client_ip) = state.ip else {
-            warn!(?session_id, "Unable to find client ip");
```
xv-ian-c, line 58: maybe the warning is still useful without the session id?

lightway-core/src/plugin.rs
```diff
@@ -42,6 +42,9 @@ pub enum PluginResult {
     /// [`Plugin`] dropped the packet
     #[error("Plugin drops the packet")]
     Drop,
+    /// [`Plugin`] dropped the packet and returned a reply packet to send back
+    #[error("Plugin drops the packet with reply packet")]
+    DropWithReply(BytesMut),
```
xv-ian-c, line 49: From the implementation it appears that this is an inside plugin only feature? Is it worth spelling that out here?

And then for the two "We cannot handle reply packet. Ignore it" comments they could also be clarified like "Outside plugins cannot drop with reply" or something similar?

lightway-server/src/io/inside/tun.rs
kp-mariappan-ramasamy: Thanks for reminding. I wanted to create metrics and missed it. Updated it now

PR #15, opened 2024-06-03
CVPN-398: lightway-client: allow all `RootCertificate` variants in `ClientConfig`

xv-ian-c:
## Description

Change `ClientConfig::ca_cert: PathBuf` to `root_ca_cert: RootCertificate<'cert>`

To support the lifetime I switched from `derivative` (which [appears to be unmaintained](https://github.com/mcarton/rust-derivative/issues/117)) to `educe` which seemed as good as any out of the set of replacements. Doing this caused me to notice a couple of things about the debug output of the `ClientConfig` (printed at start of day) which weren't quite right so I fixed them here too.

## Motivation and Context

This will allow consumers to opt into `RootCertificate::PemFileOrDirectory` or `RootCertificate::PemBuffer` as they like.

## How Has This Been Tested?

Existing test cases ensure no regressions.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -14,18 +14,20 @@ use lightway_app_utils::{
 use lightway_core::{
     ipv4_update_destination, ipv4_update_source, BuilderPredicates, ClientContextBuilder,
     ClientIpConfig, ConnectionError, ConnectionType, Event, IOCallbackResult, InsideIpConfig,
-    OutsidePacket, RootCertificate, State,
+    OutsidePacket, State,
 };
 
 // re-export so client app does not need to depend on lightway-core
 pub use lightway_core::{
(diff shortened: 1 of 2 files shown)


lightway-client/src/lib.rs
```diff
@@ -183,11 +186,9 @@ async fn handle_events(mut stream: EventStream, keepalive: Keepalive) {
     }
 }
 
-pub async fn client(config: ClientConfig) -> Result<()> {
+pub async fn client(config: ClientConfig<'_>) -> Result<()> {
     println!("Client starting with config:\n{:#?}", &config);
```
kp-mariappan-ramasamy, line 190: I wonder whether we should have used Display trait instead of Debug here.

xv-ian-c: Yeah, I was thinking the same thing while typing these things in.

`Display` is not `derive`-able at all so it would be a totally manual implementation. That might not be a bad thing though (except we'd inevitably forget to update it when we added fields...)

PR #22, opened 2024-06-18
CVPN-956: removing log noise

kp-stanislav-baiduzhyi:
## Description
Before: both log line and metric are emitted when the traffic is sent faster than receiver can accept it.
After: only metric is emitted.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/metrics.rs
+++ b/lightway-core/src/metrics.rs
@@ -1,5 +1,4 @@
 use metrics::counter;
-use tracing::warn;
 use wolfssl::ProtocolVersion;
 
 const METRIC_CONNECTION_ALLOC_FRAG_MAP: &str = "conn_alloc_frag_map";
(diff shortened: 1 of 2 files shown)


lightway-core/src/metrics.rs
```diff
@@ -21,7 +20,6 @@ pub(crate) fn wolfssl_appdata(tls_version: &ProtocolVersion) {
 }
 
 /// A call to [`crate::io::InsideIOSendCallback::send`] failed
-pub(crate) fn inside_io_send_failed(err: std::io::Error) {
-    warn!(%err, "Failed to send to inside IO");
+pub(crate) fn inside_io_send_failed() {
```
xv-ian-c, line 23: Oh. This might be more trickier than we expected, since this removes reporting of all send errors, not just those  due to the send buffer filling up.

I can see two ways we could address this:
* Enhance the return value at whichever point generates the actual error such that we can match on it here.
* Consume the error at the point where it occurs and generate a new more specific metric for the "congestion" case, such that `IOCallbackResult::Ok` is what the higher layers see. I don't think we need to introduce `IOCallbackResult::Dropped` or anything, but that's an option too.

My initial inclination is towards the latter case.

PR #25, opened 2024-06-24
CVPN-945: Migrate old docs from the C impl

xv-raihaan-m:
## Description
Moves the old documentation from the C implementation.

I've removed the question regarding why C instead of Rust since we now in fact have a Rust implementation.

The other change is to update the C functions to the equivalent Rust traits in design overview.

## Motivation and Context
This is to ensure we do not have missing documentation and have to re-do work already done

## How Has This Been Tested?
docs only change

## Types of changes

## Checklist:

### DIFF
--- a/docs/design_overview.md
+++ b/docs/design_overview.md
@@ -1,0 +1,46 @@
+# Design Overview
+
+## lightway-core
+
+lightway-core is a small, multi-platform, Rust library that encapsulates the
+encryption and processing of IP packets.
+
+On its own, lightway-core is not an executable application. Instead, it is a
+*purposefully* simple library. Intentionally, lightway-core is opinionated
+about how it works and the scope it controls, and very agnostic about
+everything else. The core use case of this library is as part of a
+high-performance, always-on VPN application, which necessarily entails
+deferring items like "how do I actually send UDP packets?" to the host
+application, which can use the best API for the platform, be it a Windows
+desktop or an iPhone.
+
+## lightway-client
+
+lightway-client is a Linux implementation for a fully working Lightway client with both TCP and UDP support.
+
+## lightway-server
+
+lightway-server is a Linux implementation for a fully working Lightway server with both TCP and UDP support.
+
+# Terminology
+
+Some people may prefer to see these terms in context, see [What does it actually do?](#what-does-it-actually-do)
+
+## Inside
+Refers to data that will be wrapped or has already been unwrapped by lightway-core. This corresponds to data coming to / from the tun device.
+## Outside
+Refers to data wrapped by lightway-core. This corresponds to data coming to / from a network socket.
+## Context
+lightway-core attributes that may be shared across multiple connections
+## Connection
+lightway-core attributes that reflect a single wrapped data path between a client and server.
+
+# What does it actually do?
+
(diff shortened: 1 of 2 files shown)


docs/design_overview.md
```diff
@@ -0,0 +1,46 @@
...
+application, which can use the best API for the platform, be it a Windows
+desktop or an iPhone.
+
+## lightway-client
+
+Lightway client is a Linux implementation for a fully working Lightway-client with both TCP and UDP support. This client is used for the e2e testing of lightway-core.
+
+## lightway-server
+
+Lightway client is a Linux implementation for a fully working Lightway-client with both TCP and UDP support. This client is used for the e2e testing of lightway-core.
```
kp-stanislav-baiduzhyi: Title says `server` but the text only talks about client 🤔

docs/design_overview.md
```diff
@@ -0,0 +1,46 @@
...
+about how it works and the scope it controls, and very agnostic about
+everything else. The core use case of this library is as part of a
+high-performance, always-on VPN application, which necessarily entails
+deferring items like "how do I actually send UDP packets?" to the host
+application, which can use the best API for the platform, be it a Windows
+desktop or an iPhone.
+
+## lightway-client
+
+Lightway client is a Linux implementation for a fully working Lightway-client with both TCP and UDP support. This client is used for the e2e testing of lightway-core.
```
xv-ian-c: Although it is true that the client is used for e2e testing it's actually far more capable and is actually a completely capable/deployable client implementation.

I think that can be resolved by simply dropping the second sentence.

Same goes for the server block below.

docs/design_overview.md
```diff
@@ -0,0 +1,46 @@
...
+
+One direction is the "inside path". The host application passes data to lightway-core for wrapping via `Connection.inside_data_received`, and then lightway-core will call the host application's `outside_io.send` one or more times with the appropriately encrypted data, which the host is then responsible for transmitting appropriately. Where "appropriately" normally means "send these packets over the internet, client->server or server->client depending on which side of the connection we are on."
+
+The other direction is the "outside path". The host application passes data wrapped by lightway-core to the appropriate library function `Connection.outside_data_received`; lightway-core will then call the host application's `inside_io.send` one or more times with the appropriately unwrapped data, which the host can then deliver the data into the tun device.
+
+Of course, the devil is in the details, and there are a lot of details here. For an example of how the above works you can see [IP Translation](./ip_translation.md#packet-flow-steps-as-marked-yellow-in-above-picutre)
```
xv-ian-c: There is a typo in the anchor here "picutre" which is actually because there is a typo in the text. Would you mind fixing please?

docs/design_overview.md
xv-raihaan-m: Good eye fixed thanks!

docs/design_overview.md
xv-raihaan-m: Didn't notice it because autocomplete wrote it 😅 . Thanks Ian!

docs/design_overview.md
xv-raihaan-m: Yeah I've dropped the second sentence

PR #26, opened 2024-06-24
Earthly: copy artifacts locally from build container

kp-mariappan-ramasamy:
## Description

Update Earthy target to copy binaries locally after building

## Motivation and Context

It makes it easier to use the binaries built using Earhtly.

## How Has This Been Tested?
Yes, verified the binaries copied locally.

```
❯ tree target/
target/
└── x86_64-unknown-linux-gnu
    └── release
        ├── lightway-client
        └── lightway-server

3 directories, 2 files
```

## Types of changes

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -47,8 +47,8 @@ build:
 
     DO lib-rust+CARGO --args="build --release --target=$target" --output="$target/release/lightway-(client|server)$"
 
-    SAVE ARTIFACT ./target/$target/release/lightway-client
-    SAVE ARTIFACT ./target/$target/release/lightway-server
+    SAVE ARTIFACT ./target/$target/release/lightway-client AS LOCAL ./target/$target/release/
+    SAVE ARTIFACT ./target/$target/release/lightway-server AS LOCAL ./target/$target/release/
 
 # build-arm64 build for arm64. Support building from an amd64 or arm64 host
 build-arm64:


Earthfile
```diff
@@ -47,8 +47,8 @@ build:
 
     DO lib-rust+CARGO --args="build --release --target=$target" --output="$target/release/lightway-(client|server)$"
 
-    SAVE ARTIFACT ./target/$target/release/lightway-client
-    SAVE ARTIFACT ./target/$target/release/lightway-server
+    SAVE ARTIFACT ./target/$target/release/lightway-client AS LOCAL ./target/$target/release/
+    SAVE ARTIFACT ./target/$target/release/lightway-server AS LOCAL ./target/$target/release/
```
xv-ian-c, line 51: FYI you don't actually _need_ to do this to get access to the artifacts, you can just use `earthly -a '+build/*' foo/` to copy artifacts to `foo/` after build:
```console
$ earthly -a '+build/*' foo/
...
 Local Output Summary 🎁 (single artifact)
————————————————————————————————————————————————————————————————————————————————

Artifact +build/lightway-client output as foo/lightway-client
Artifact +build/lightway-server output as foo/lightway-server


========================== 🌍 Earthly Build  ✅ SUCCESS ==========================
$ ls foo/
lightway-client*  lightway-server*
```

But I guess it is more convenient if it happens by default?

And I suppose having it produce the same path as a local `cargo build` is useful if you have scripts/finger macros which you want to work either way?

kp-mariappan-ramasamy: Yes, I think its much easier for new comers/testers.
This was reported by Daniel, who was proof reading the documentation/README
We always tell earthly will make generating binaries easier. So there should be a default way to copy binaries out from build container. IMHO this is must have

PR #31, opened 2024-06-26
tests: Add test cases with a single Tokio worker thread

xv-ian-c:
The hope is that this will help catch potential deadlocks or other problems/interactions which do not manifest if there are multiple worker threads.

To facilitate this these tests enable PMTUD and keepalive so we have the large set of tasks.

Also enable io_uring in one of the tests but not the other, in principal this is independent of the tunnel mode so this allows us to test both io_uring and tun modes without exploding the test matrix any further.

### DIFF
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -12,6 +12,9 @@ TEST:
     FUNCTION
     FROM ./e2e+build-container
 
+    ARG CLIENT_TOKIO_WORKER_THREADS
+    ARG SERVER_TOKIO_WORKER_THREADS
+
     ARG MODE
     ARG SERVER_PORT
     ARG SERVER_EXTRA_ARGS=""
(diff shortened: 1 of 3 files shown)


tests/Earthfile
```diff
@@ -102,6 +105,14 @@ run-udp-keepalive-test:
 run-tcp-keepalive-test:
     DO +TEST --MODE=tcp --SERVER_PORT=443 --CLIENT_EXTRA_ARGS="--keepalive-interval=2s --keepalive-timeout=6s"
 
+# run-udp-single-threaded-test runs e2e test of UDP with server and client using a single Tokio worker thread
```
kp-stanislav-baiduzhyi, line 108: Might be worth adding to the comment that `--enable-tun-iouring` is intentionally here unlike the next test

xv-ian-c: This is covered in the commit message which I believe is sufficient.

PR #37, opened 2024-07-04
CVPN-939 Sync Connection State Enum Values with Lightway C

xv-thomas-leong:
## Description
Explicitly specify the connection state enum values to maintain consistency between Lightway C and Rust.

State enum definition in Lightway C (under `he_conn_state`) can be found here:
https://github.com/expressvpn/lightway-core/blob/main/include/he.h

## How Has This Been Tested?

CI tested

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -63,24 +63,25 @@ const WOLF_TICK_DTLS13_QUICK_TIMEOUT_DIVISOR: u32 = 4;
 
 /// Connection state
 #[derive(Debug, Clone, Copy, PartialEq)]
+#[repr(u8)]
 pub enum State {
     /// Secure connection is being established.
-    Connecting,
+    Connecting = 2,
 
     /// Secure connection is established
-    LinkUp,
+    LinkUp = 6,
 
     /// Connection is established, client is authenticating
-    Authenticating,
+    Authenticating = 5,
     // Configuring,
     /// Tunnel is online
-    Online,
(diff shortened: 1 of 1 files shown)


lightway-core/src/connection.rs
```diff
@@ -63,24 +63,25 @@ const WOLF_TICK_DTLS13_QUICK_TIMEOUT_DIVISOR: u32 = 4;
...
-    Online,
+    Online = 7,
 
     /// Disconnect is in progress
-    Disconnecting,
+    Disconnecting = 4,
 
     /// Connection has been disconnected
-    Disconnected,
+    Disconnected = 1,
```
kp-mariappan-ramasamy, line 84: We might as well arrange it in ascending order ?

xv-thomas-leong: Sure. fixed in fcb100d

xv-ian-c: Personally I'd rather it was in the order of the expected normal flow through (which I think is what it was before), the fact that the C version has numbers which don't order to match that is unfortunate but it is what it is.

kp-mariappan-ramasamy: Hmmm, its true. We have even holes in between the numbers.
@xv-thomas-leong sorry for conflicting comment, can we revert to previous commit ?

xv-thomas-leong: @kp-mariappan-ramasamy Done.

PR #38, opened 2024-07-05
CVPN-939 Platform Specific Customizations

xv-thomas-leong:
## Description
This PR prepares Lightway to be platform independent by accepting customizations.

## How Has This Been Tested?

CI tested. 

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -13,8 +13,8 @@ use lightway_app_utils::{
 };
 use lightway_core::{
     ipv4_update_destination, ipv4_update_source, BuilderPredicates, ClientContextBuilder,
-    ClientIpConfig, ConnectionError, ConnectionType, Event, IOCallbackResult, InsideIpConfig,
-    OutsidePacket, State,
+    ClientIpConfig, ConnectionError, ConnectionType, Event, EventCallback, IOCallbackResult,
+    InsideIpConfig, OutsidePacket, State,
 };
 
 // re-export so client app does not need to depend on lightway-core
@@ -60,7 +60,7 @@ impl std::fmt::Debug for ClientConnectionType {
 
 #[derive(educe::Educe)]
 #[educe(Debug)]
-pub struct ClientConfig<'cert> {
+pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
     /// Connection mode
     pub mode: ClientConnectionType,
 
@@ -136,6 +136,10 @@ pub struct ClientConfig<'cert> {
     /// Specifies if the program responds to INT/TERM signals
     pub exit_on_ctrlc: bool,
 
+    /// Allow injection of a custom handler for event callback
+    #[educe(Debug(ignore))]
+    pub event_handler: Option<A>,
+
     /// File path to save wireshark keylog
     #[cfg(feature = "debug")]
     pub keylog: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


lightway-client/src/lib.rs
```diff
@@ -367,19 +367,28 @@ pub async fn client(config: ClientConfig<'_>) -> Result<()> {
         }
     });
 
+    #[cfg(target_os = "linux")]
```
kp-mariappan-ramasamy: It's better to add a cli config like "disable_signal_handling` and do all this magic inside of if check instead of relying on platform check directly.

It gives the client apps to choose the behavior and remove the hardcoded platform behaviour needed in our library.

lightway-client/src/lib.rs
```diff
@@ -367,19 +367,28 @@ pub async fn client(config: ClientConfig<'_>) -> Result<()> {
...
+        ctrlc_rx.recv().await;
+        #[cfg(not(target_os = "linux"))]
+        futures::future::pending::<anyhow::Error>().await;
+    };
+
     tokio::select! {
         Some(_) = keepalive_task => Err(anyhow!("Keepalive timeout")),
         io = outside_io_loop => Err(anyhow!("Outside IO loop exited: {io:?}")),
         io = inside_io_loop => Err(anyhow!("Inside IO loop exited: {io:?}")),
-        _ = ctrlc_rx.recv() => {
```
kp-mariappan-ramasamy, line 379: If you add a cli config to disable signal handling, we can ignore this select case by using this tokio syntax:
```rust
_ = ctrlc_rx.recv(), if !config.disable_signal => {
  ...
}

File: lightway-client/src/lib.rs (hunk 1 above)
Comment (xv-thomas-leong) [reply lightway-client/src/lib.rs]: I think your suggestion make sense since we don't have other platform specific interruption behavior implemented. Will make changes according to your other comment.

File: lightway-client/src/lib.rs (hunk 2 above)
Comment (xv-ian-c) [reply lightway-client/src/lib.rs:379]: There is also [`OptionFuture`](https://docs.rs/futures/latest/futures/future/struct.OptionFuture.html), if `ctrlc_rx` is one  of those those with `ctrlc_rx.recv()` inside (or not) then I think it becomes `Some(_) = that_option_thing => { ... } `. You then don't need the `futures::future::pending` because you just have a `None` instead in that case.

I think it's equivalent to @kp-mariappan-ramasamy 's suggestion and I don't really know which is better. Might be fewer conditionals with the `OptionFuture` approach.

I think `.recv` is cancellation safe (see [tokio docs](https://docs.rs/tokio/latest/tokio/macro.select.html#cancellation-safety)), so no need to worry about that?

File: lightway-client/src/lib.rs (hunk 1 above)
Comment (xv-ian-c) [reply lightway-client/src/lib.rs]: Making it an option presupposes that `ctrlc` crate at least builds on those other platforms even if it cannot be used. It supports Unix-like and Windows though so I guess this is sufficient.

BTW, I'd invert the option name `handle_ctrlc`  or `exit_on_ctrlc` or something.

Comment (xv-thomas-leong) [reply lightway-client/src/lib.rs]: `exit_on_ctrlc` sounds good. Will make a change 🙏

File: lightway-core/src/connection/event.rs [hunk 3]
```diff
@@ -1,7 +1,7 @@
 use crate::{SessionId, State};
 
 /// A lightway event
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
xv-ian-c: I don't see where this new capability is used, is it a left over from a previous approach?

xv-thomas-leong: It isn't needed anymore. Thanks for noticing. Fixed in 1ebb877

PR #39, opened 2024-07-05
CVPN-939 Non-Linux Tun Support

xv-thomas-leong:
## Description

This PR introduces tun support for non-linux platforms, as long as the fd of the tun is supplied. Platform specific logic for creating tun/retrieving tun fd is out of the scope of this PR.

## Motivation and Context

The tun IO crate of choice does not support platform outside of Linux. This PR addresses this by introducing a "TokioRawTun" module which should be reusable across platforms.

## How Has This Been Tested?

Manual tested.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/Cargo.toml
+++ b/lightway-app-utils/Cargo.toml
@@ -22,6 +22,7 @@ serde = "1.0.189"
 serde_with = "3.4.0"
 tokio = { version = "1.33.0", features = ["macros", "rt", "sync", "time"], optional = true }
 tokio-stream = { version = "0.1.14", optional = true }
+libc = "0.2.152"
 clap = { version = "4.4.7", features = ["derive"] }
 tracing-subscriber = { version = "0.3.17", features = ["json"] }
 tokio-util = "0.7.10"
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 #[cfg(feature = "tokio-tun")]
 use tokio_tun::Tun as TokioTun;
 
+use crate::tokio_raw_tun::TokioRawTun;
 #[cfg(feature = "io-uring")]
 use crate::IOUring;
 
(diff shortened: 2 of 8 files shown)


lightway-app-utils/Cargo.toml
```diff
@@ -31,7 +32,7 @@ bytes = "1.5.0"
 tokio-eventfd = "0.2.1"
 dashmap = "6.0.1"
 thiserror = "1.0.57"
-tokio-tun = "0.11.2"
+tokio-tun = { version = "0.11.2", optional = true}
```
xv-ian-c, line 36: A more common way to reflect the platform specific nature of a dep is to put it into a target specific dep block:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
```

block.

In this case given `io-uring` is already a feature I'm not sure which approach is better here.

lightway-app-utils/src/tun.rs
```diff
@@ -149,6 +162,59 @@ impl AsRawFd for TunDirect {
...
 
+pub struct TunRaw {
+    tun: TokioRawTun,
+    mtu: usize,
+}
+
+impl TunRaw {
+    /// Create `TunRaw` struct
+    pub fn new(fd: RawFd, mtu: Option<i32>) -> Result<Self> {
+        let mtu = mtu.unwrap_or(1350) as usize;
```
xv-ian-c, line 173: I think only the caller who produced the `RawFd` knows the correct MTU to use.

So I think it makes sense to force them to tell us by making it non-optional.

lightway-client/src/main.rs
```diff
@@ -54,7 +54,10 @@ async fn main() -> Result<()> {
         root_ca_cert,
         outside_mtu: config.outside_mtu,
         inside_mtu: config.inside_mtu,
+        #[cfg(feature = "linux-tun")]
         tun_name: config.tun_name,
+        #[cfg(not(feature = "linux-tun"))]
+        tun_fd: -1, // A placeholder until tun_fd retrival is implemented
```
xv-ian-c, line 60: This smells a bit like some sort of enum ought to be exposed through to `main.rs`.

I wonder if a field of type `lightway_app_utils::Tun` could be made to work here? Might need a bit of rectoring but might not be too bad.

Or if`io::inside::Tun` were available publicly that could be passed here.

PR #42, opened 2024-07-10
CVPN-939: use tun2 instead of tokio-tun

kp-mariappan-ramasamy:
## Description

Change tokio-tun to tun2 for tunnel management.

## Motivation and Context

tokio-tun is not cross platform supported. 

## How Has This Been Tested?
Verified in unit tests and running traffic manaully with UDP and TCP connections

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -3,7 +3,7 @@ mod io;
 mod keepalive;
 
 use crate::io::inside::InsideIO;
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use bytes::BytesMut;
 use bytesize::ByteSize;
 use keepalive::Keepalive;
@@ -33,8 +33,8 @@ use std::{
 };
 use tokio::{
     net::{TcpStream, UdpSocket},
+    sync::mpsc::Receiver,
     task::JoinHandle,
-    sync::mpsc::Receiver
 };
 use tokio_stream::StreamExt;
 use tracing::info;
@@ -211,11 +211,15 @@ pub async fn client<A: 'static + Send + EventCallback>(
     let (connection_type, outside_io): (ConnectionType, Arc<dyn io::outside::OutsideIO>) =
         match config.mode {
             ClientConnectionType::Datagram(maybe_sock) => {
-                let sock = io::outside::Udp::new(&config.server, maybe_sock).await?;
+                let sock = io::outside::Udp::new(&config.server, maybe_sock)
+                    .await
+                    .context("Outside IO UDP")?;
                 (ConnectionType::Datagram, sock)
             }
             ClientConnectionType::Stream(maybe_sock) => {
-                let sock = io::outside::Tcp::new(&config.server, maybe_sock).await?;
+                let sock = io::outside::Tcp::new(&config.server, maybe_sock)
+                    .await
+                    .context("Outside IO TCP")?;
                 (ConnectionType::Stream, sock)
             }
         };
@@ -243,7 +247,8 @@ pub async fn client<A: 'static + Send + EventCallback>(
             #[cfg(feature = "io-uring")]
             iouring,
         )
-        .await?,
+        .await
+        .context("Tun creation")?,
     );
 
     let (event_cb, event_stream) = EventStreamCallback::new();


lightway-app-utils/src/tun.rs
```diff
@@ -78,37 +88,56 @@ impl AsRawFd for Tun {
...
+        let fd = tun.as_ref().as_raw_fd();
+        let mtu = tun.as_ref().mtu()?;
 
-        Ok(TunDirect { tun, mtu })
+        let (read_half, write_half) = tokio::io::split(tun);
+        let read_half = Arc::new(Mutex::new(read_half));
+        let (send_q_tx, send_q_rx) = bounded(Self::CHANNEL_SIZE);
+
+        thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
```
xv-ian-c: A new thread and the associated communication channel and lock etc concerns me, have you measured the performance impact?

In the I/O uring case is there now 2 layers of threads and channels?

lightway-app-utils/src/tun.rs
```diff
@@ -124,25 +153,38 @@ impl TunDirect {
...
-            Ok(nr) => IOCallbackResult::Ok(nr),
-            Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
-                IOCallbackResult::WouldBlock
+    pub fn try_send(&self, buf: BytesMut) -> IOCallbackResult<usize> {
+        let buf_size = buf.len();
+        let try_send_res = self.send_q_tx.try_send(buf);
+        match try_send_res {
+            Ok(()) => IOCallbackResult::Ok(buf_size),
+            Err(e) if e.is_full() => IOCallbackResult::WouldBlock,
+            Err(e) => {
```
xv-ian-c: Isn't this only the result of sending to the channel, it's losing any actual IO error from inside the thread.

kp-mariappan-ramasamy: There is another error which will be raised when channel is closed:
https://docs.rs/async-channel/latest/async_channel/enum.TrySendError.html#variant.Full

But you are right, we are missing the error inside send thread. I have handled it to close the thread and thus activate above SendErrror::Full error

lightway-app-utils/src/tun.rs
kp-mariappan-ramasamy: I have verified that there is no performance impact due to this.

No, io-uring and direct are parallel things. Both now uses one thread to send receive packet

lightway-client/src/lib.rs
```diff
@@ -136,7 +137,7 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
     pub outside_plugins: PluginFactoryList,
 
     /// Specifies if the program responds to INT/TERM signals
-    pub exit_on_ctrlc: bool,
+    pub stop_signal: Receiver<()>,
```
kp-mariappan-ramasamy, line 141: Will merge with previous commit if this approach of stop_signal seems sane

lightway-client/src/lib.rs
```diff
@@ -390,21 +391,23 @@ pub async fn client<A: 'static + Send + EventCallback>(
...
-    let (ctrlc_tx, mut ctrlc_rx) = tokio::sync::mpsc::channel(1);
-    if config.exit_on_ctrlc {
+    let (ctrlc_tx, ctrlc_rx) = tokio::sync::mpsc::channel(1);
+    if config.exit_on_ctrlc.is_none() {
         ctrlc::set_handler(move || {
             ctrlc_tx.blocking_send(()).expect("CtrlC handler failed");
         })?;
     }
 
+    let mut stop_signal = config.exit_on_ctrlc.unwrap_or( ctrlc_rx);
```
xv-ian-c: The commit message is not very clear on what the desired semantics here are, but I think it is to allow the app to provide an alternative signal handling channel which the we will await on here?

I think this would be clearer as:
```
let mut stop_signal = match config.exit_on_ctrlc {
    Some(rx) => rx,
    None => {
        let (ctrlc_tx, ctrlc_rx) = tokio::sync::mpsc::channel(1);
        ctrlc::set_handler(move || {
            ctrlc_tx.blocking_send(()).expect("CtrlC handler failed");
        })?;
       ctrlc_rx
    }
}
```

lightway-client/src/lib.rs
```diff
@@ -390,21 +391,23 @@ pub async fn client<A: 'static + Send + EventCallback>(
...
         Some(_) = keepalive_task => Err(anyhow!("Keepalive timeout")),
         io = outside_io_loop => Err(anyhow!("Outside IO loop exited: {io:?}")),
         io = inside_io_loop => Err(anyhow!("Inside IO loop exited: {io:?}")),
-        _ = ctrlc_rx.recv(), if config.exit_on_ctrlc => {
-            info!("sigint/sigterm received, gracefully shutting down");
+        _ = stop_signal.recv() => {
+            info!("gracefully shutting down");
             let _ = conn.lock().unwrap().disconnect();
-            Err(anyhow!("sigint/sigterm received"))
+            Ok(())
```
xv-ian-c, line 407: I spotted this on Friday but didn't get as far as a PR, this is a good change, we should do the same in the server too

lightway-client/src/lib.rs
```diff
@@ -136,7 +137,7 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
     pub outside_plugins: PluginFactoryList,
 
     /// Specifies if the program responds to INT/TERM signals
-    pub exit_on_ctrlc: bool,
+    pub exit_on_ctrlc: Option<Receiver<()>>,
```
xv-ian-c: I think the comment and probably the field name need to be updated to reflect this change.

lightway-client/src/lib.rs
xv-ian-c: Ah, that might invalidate one or more of my previous comments.

lightway-app-utils/src/tun.rs
```diff
@@ -78,37 +88,56 @@ impl AsRawFd for Tun {
...
+        let tun = tun2::create_as_async(&tun2_config)?;
+        let fd = tun.as_ref().as_raw_fd();
+        let mtu = tun.as_ref().mtu()?;
 
-        Ok(TunDirect { tun, mtu })
+        let (read_half, write_half) = tokio::io::split(tun);
+        let read_half = Arc::new(Mutex::new(read_half));
+        let (send_q_tx, send_q_rx) = bounded(Self::CHANNEL_SIZE);
+
+        thread::spawn(move || {
```
xv-ian-c: Does this really need a dedicated thread? Would `tokio::spawn` not work just as well and use the existing Tokio threadpool?

lightway-app-utils/src/tun.rs
```diff
@@ -71,35 +82,46 @@ impl AsRawFd for Tun {
...
-        Ok(TunDirect { tun, mtu })
+    /// Create a new `Tun` struct
+    pub fn new(config: TunConfig) -> Result<Self> {
+        let tun = tun2::create_as_async(&config)?;
+        let fd = tun.as_ref().as_raw_fd();
+        let mtu = tun.as_ref().mtu()?;
+
+        let (read_half, write_half) = tokio::io::split(tun);
+        let read_half = Arc::new(Mutex::new(read_half));
+        let (send_q_tx, send_q_rx) = bounded(Self::CHANNEL_SIZE);
```
xv-ian-c: It's non-obvious (especially since the [docs](https://docs.rs/tun2/latest/tun2/) are generated without the async feature) but it is possible (I think, I've not tried) to get a sync write method out of the `AsyncDevice` which you have here, via the [`as_ref`](https://github.com/tun2proxy/rust-tun/blob/v2/src/async/unix_device.rs#L33) impl (or the `as_mut`) one.

If I'm reading the code correctly I think something like `tun.as_ref().write()` or `tun.as_mut().write()` would do the trick (this is what is inside the async `poll_write` method).

If I'm right then the need for this channel and the thread goes away I think and you don't need to split either?

lightway-app-utils/src/tun.rs
```diff
@@ -136,13 +138,13 @@ impl TunDirect {
 
     /// MTU of Tun
     pub fn mtu(&self) -> usize {
```
xv-ian-c, line 123: should this have become u16? (the field is u16 here and the next commits changes other apis from u32 to u16)

kp-mariappan-ramasamy: I have tried that earlier and it went deeper to use "as usize" in many places. So left it

PR #49, opened 2024-07-24
Correct TLS key rotation and generate events/metrics for start and complete

xv-ian-c:
## Description

Previously we were calling `wolfssl::Session::try_trigger_update_key()` (which binds [`wolfSSL_update_keys`](https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_update_keys)) but not calling `wolfssl::Session::is_update_keys_pending()` (which binds [`wolfSSL_key_update_response`](https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_key_update_response)) to detect completion. We now do so and therefore generate events for both the start and end of a rotation, which we use to produce metrics.

Furthermore the implementation of `wolfssl::Session::try_trigger_update_key()` was incorrect and was not actually starting a rotation, this was addressed in https://github.com/expressvpn/wolfssl-rs/pull/162 and a bump to consume that is included here.

Following the lead of the C version of lightway we also resume ticking while the rotation is pending, this helps detect connections which have failed during rotation.

TODO:
- [x] update with merge/release of https://github.com/expressvpn/wolfssl-rs/pull/162

## Motivation and Context

This is useful to monitor for issues with key rotation.

## How Has This Been Tested?

I've run both TCP and UDP connections in the local containerized setup with `--log-level trace --key-update-interval 3s` and observed the rotations in the logs.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/connection_manager.rs
+++ b/lightway-server/src/connection_manager.rs
@@ -143,6 +143,9 @@ fn handle_finalize_session_rotation(conn: &Weak<Connection>, old: SessionId, new
 
 #[instrument(level = "trace", skip_all)]
 fn handle_tls_keys_update_start(conn: &Weak<Connection>) {
+    metrics::connection_key_update_start();
+
+    // For UDP connections begin a session ID rotation
     let Some(conn) = conn.upgrade() else {
         info!("Connection has gone away, stopping");
         return;
@@ -151,6 +154,11 @@ fn handle_tls_keys_update_start(conn: &Weak<Connection>) {
     if conn.connection_type().is_datagram() {
         conn.begin_session_id_rotation();
     }
+}
+
+#[instrument(level = "trace", skip_all)]
+fn handle_tls_keys_update_complete() {
+    metrics::connection_key_update_complete();
 }
 
 #[instrument(level = "trace", skip_all)]
@@ -163,7 +171,7 @@ async fn handle_events(mut stream: EventStream, conn: Weak<Connection>) {
                 handle_finalize_session_rotation(&conn, old, new);
             }
             Event::TlsKeysUpdateStart => handle_tls_keys_update_start(&conn),
-            Event::TlsKeysUpdateCompleted => {}
+            Event::TlsKeysUpdateCompleted => handle_tls_keys_update_complete(),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


lightway-core/src/connection.rs
kp-mariappan-ramasamy, line 1: Typo in commit message of e6e51fb862e6ac4744a8c44fdf942b33de02056a
`neglibible`

lightway-core/src/connection.rs
```diff
@@ -1024,36 +1020,22 @@ impl<AppState: Send> Connection<AppState> {
     // Trigger a periodic key update for TLS/DTLS 1.3 server
     // connections.
     fn maybe_update_tls_keys(&mut self) -> ConnectionResult<()> {
-        // Only if online
-        if !matches!(self.state, State::Online) {
-            return Ok(());
+        // Only for TLS/DTLS 1.3
+        match self.session.version() {
```
kp-mariappan-ramasamy: We could also consider using api's added for TLS1.3 checks in wolfssl:
https://github.com/expressvpn/wolfssl-rs/pull/162/commits/865b69a12bf390980f19ab25deb2fc0ef0827e8b#diff-b61040fb09ab9b344fe843e53557d6c6976bae483c11a2e101a8f1e02198d902R142

lightway-core/src/connection/key_update.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /// Check if a key update has completed.
+    ///
+    /// Implicitly transitions to `Self::Waiting` if required.
+    ///
+    /// The `still_pending` argument is the result of calling
+    /// `wolfssl::Session::is_update_keys_pending()`.
+    pub fn complete(&mut self, still_pending: bool) -> bool {
```
kp-mariappan-ramasamy: This is a nice refactor.
One minor thing is this api receiving `still_pending` and return it back seems little un-intuitive.

WDYT of making this api `pub fn complete(&mut self)  {}` and call based on `self.session.is_update_keys_pending()`'s output
This will only be a state transition api

lightway-core/src/connection.rs
xv-ian-c: They are private :-(

lightway-core/src/connection/key_update.rs
xv-ian-c: Done!

lightway-core/src/connection.rs
kp-mariappan-ramasamy: That's sad. It should have been a pub function.
Its a useful one.

PR #52, opened 2024-07-26
sockopt: make sockopt file mac compatible

xv-raihaan-m:
## Description
Makes more of the codebase macOS compatible

## Motivation and Context
My main machine is a macbook and this allows running some tests locally without starting earthly

## How Has This Been Tested?
This file no longer errors out during compilation locally

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp/sockopt.rs
+++ b/lightway-client/src/io/outside/udp/sockopt.rs
@@ -53,12 +53,27 @@ pub(super) fn get_ip_mtu_discover(sock: &tokio::net::UdpSocket) -> std::io::Resu
     let mut value: MaybeUninit<libc::c_int> = MaybeUninit::uninit();
     let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
 
+    let level: i32;
+    let optname: i32;
+
+    #[cfg(target_os = "macos")]
+    {
+        level = libc::IPPROTO_IP;
+        optname = libc::IP_DONTFRAG;
+    }
+
+    #[cfg(not(target_os = "macos"))]
+    {
+        level = libc::SOL_IP;
+        optname = libc::IP_MTU_DISCOVER;
+    }
+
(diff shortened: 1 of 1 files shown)


lightway-client/src/io/outside/udp/sockopt.rs
```diff
@@ -54,6 +54,18 @@ pub(super) fn get_ip_mtu_discover(sock: &tokio::net::UdpSocket) -> std::io::Resu
     let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
 
     // SAFETY: `getsockopt` requires an fd and a valid buffer of `c_int` size
```
xv-ian-c: Please make sure every `unsafe` block has an appropriate `SAFETY` comment. With this change the one below (the original one which is now the `#[cfg(not(target_os = "macos"))]` one) does not.

xv-raihaan-m: Addressed this by moving the level / optname difference into a single unsafe block

lightway-client/src/io/outside/udp/sockopt.rs
```diff
@@ -55,10 +55,25 @@ pub(super) fn get_ip_mtu_discover(sock: &tokio::net::UdpSocket) -> std::io::Resu
...
+        #[cfg(target_os = "macos")]
+        {
+            level = libc::IPPROTO_IP;
+            optname = libc::IP_DONTFRAG;
+        }
+
+        #[cfg(not(target_os = "macos"))]
+        {
+            level = libc::SOL_IP;
+            optname = libc::IP_MTU_DISCOVER;
```
xv-ian-c: In general the `unsafe` block should be as small as possible, such that the scope of the `SAFETY` comment clearly applies to the minimum number of unsafe operations.

So I think these declarations should be above, outside the unsafe block (same in the `setsockopt` case)

PR #53, opened 2024-07-26
CVPN-1144: add network change behaviours

xv-raihaan-m:
## Description
Adds a way to notify the client of a network change.

On TCP: We are expected to disconnect as there is no way to recover
On UDP: We probe the connection via keepalives to ensure it is still usable

## Motivation and Context
To support floating connections properly

## How Has This Been Tested?
Tested on an android device

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -5,6 +5,7 @@ pub mod keepalive;
 use anyhow::{anyhow, Context, Result};
 use bytes::BytesMut;
 use bytesize::ByteSize;
+use futures::future::OptionFuture;
 use keepalive::Keepalive;
 use lightway_app_utils::{
     args::Cipher, connection_ticker_cb, ConnectionTicker, ConnectionTickerState, DplpmtudTimer,
@@ -29,12 +30,12 @@ use pnet::packet::ipv4::Ipv4Packet;
 use std::path::PathBuf;
 use std::{
     net::Ipv4Addr,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Weak},
     time::Duration,
 };
 use tokio::{
     net::{TcpStream, UdpSocket},
-    sync::oneshot::Receiver,
+    sync::{mpsc, oneshot},
     task::{JoinHandle, JoinSet},
 };
 use tokio_stream::StreamExt;
@@ -58,6 +59,12 @@ impl std::fmt::Debug for ClientConnectionType {
             Self::Datagram(_) => f.debug_tuple("Datagram").finish(),
         }
     }
+}
+
+#[derive(Debug)]
+pub enum ClientResult {
+    UserDisconnect,
+    NetworkChange,
 }
 
 #[derive(educe::Educe)]
@@ -143,7 +150,13 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
 
     /// Specifies if the program responds to INT/TERM signals
     #[educe(Debug(ignore))]
-    pub stop_signal: Receiver<()>,
+    pub stop_signal: oneshot::Receiver<()>,
+
+    /// Signal for notifying a network change event
+    /// network change being defined as a change in
+    /// wifi networks or a change of network interfaces
+    #[educe(Debug(ignore))]
+    pub network_change_signal: Option<mpsc::Receiver<()>>,
 
     /// Allow injection of a custom handler for event callback
     #[educe(Debug(ignore))]
@@ -297,10 +310,48 @@ pub async fn inside_io_task<T: Send + Sync>(
     }
 }
 
+async fn handle_network_change(
+    keepalive: Keepalive,
+    mut network_change_signal: mpsc::Receiver<()>,
+    weak: Weak<Mutex<lightway_core::Connection<ConnectionState>>>,
+) -> ClientResult {
+    while (network_change_signal.recv().await).is_some() {
+        let Some(conn) = weak.upgrade() else {
+            return ClientResult::UserDisconnect;
+        };
+        let conn_type = conn.lock().unwrap().connection_type();
+        match conn_type {
+            ConnectionType::Datagram => {
+                info!("sending keepalives due to network change ..");
+                keepalive.network_changed().await;
+            }
+            ConnectionType::Stream => {
+                info!("client shutting down due to network change ..");
+                let _ = conn.lock().unwrap().disconnect();
+                return ClientResult::NetworkChange;
+            }
+        }
+    }
+    ClientResult::UserDisconnect
+}
+
+fn validate_client_config<A: 'static + Send + EventCallback>(
+    config: &ClientConfig<'_, A>,
+) -> Result<()> {
+    if config.network_change_signal.is_some() && config.keepalive_interval.is_zero() {
+        return Err(anyhow!(
+            "Keepalive interval cannot be zero when network change signal is set"
+        ));
+    }
+    Ok(())
+}
+
 pub async fn client<A: 'static + Send + EventCallback>(
     mut config: ClientConfig<'_, A>,
-) -> Result<()> {
+) -> Result<ClientResult> {
     println!("Client starting with config:\n{:#?}", &config);
+
+    validate_client_config(&config)?;
 
     let mut join_set = JoinSet::new();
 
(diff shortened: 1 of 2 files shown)


lightway-client/src/lib.rs
```diff
@@ -316,7 +351,16 @@ pub async fn client<A: 'static + Send + EventCallback>(
     ticker_task.spawn(Arc::downgrade(&conn));
     pmtud_timer_task.spawn(Arc::downgrade(&conn));
 
+    if config.network_change_signal.is_some() {
```
kp-mariappan-ramasamy: You could use `if let` syntax to get rid of unwrap. Example on 
https://github.com/expressvpn/lightway/pull/53/commits/eeeb01984b39f982ce3e30ddd9d8c4714ca16c28#diff-76cc785c1d261df977f03e710141e50595aa4dbe21f36ea29edccc143ff96c3cR370

```rust
if let Some(network_change) = config.network_change_signal {
    tokio::spawn(handle_network_change(
            keepalive.clone(),
            .network_change,
            Arc::downgrade(&conn),
        ));
}
```

lightway-client/src/lib.rs
```diff
@@ -316,7 +351,16 @@ pub async fn client<A: 'static + Send + EventCallback>(
...
+        tokio::spawn(handle_network_change(
+            keepalive.clone(),
+            config.network_change_signal.unwrap(),
+            Arc::downgrade(&conn),
+        ));
+    }
+
     let outside_io_loop_conn = conn.clone();
+    let outside_io_loop_keepalive = keepalive.clone();
```
kp-mariappan-ramasamy: We do not need to clone here, since this is the last keepalive used.

lightway-client/src/lib.rs
```diff
@@ -140,7 +145,13 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
...
     /// Specifies if the program responds to INT/TERM signals
     #[educe(Debug(ignore))]
-    pub stop_signal: Receiver<()>,
+    pub stop_signal: OneShotReceiver<()>,
+
+    /// Signal for notifying a network change event
+    /// network change being defined as a change in
+    /// wifi networks or a change of network interfaces
+    #[educe(Debug(ignore))]
+    pub network_change_signal: Option<MpscReceiver<()>>,
```
kp-mariappan-ramasamy: The current code implicitly requires `keepalive_interval` to be valid non zero value. 
Otherwise, keepalive task will not be created and thus this signal will fail:

Ref: https://github.com/expressvpn/lightway/blob/main/lightway-client/src/keepalive.rs#L81

lightway-client/src/lib.rs
xv-raihaan-m: Fixed, good eye. Holdover from a previous iteration

lightway-client/src/lib.rs
xv-raihaan-m: That is correct, I'm unsure of how best to manage these 2 requirements. Should we cause an error if the signal is set without a timer value? Thinking out loud here

xv-ian-c: Best way would be to leverage the type system to make it impossible to specify one without the other, e.g. by making them a struct so you have to provide neither or both, but there is no `std::time::NonZeroDuration` which you would need to use for the interval in such a struct (otherwise the invariant isn't enforced anymore).  I can't think of another way to use the type system here.

I think returning a new error from `client()` if `network_change_signal` is set but the interval is not > 0s is fine.

Or if keepalive is not enabled then maybe network change should just become a synonym for "stop_signal"? For TCP that makes 100% sense, I'm less sure for the UDP case.

xv-ian-c: We are talking here about the link to `keepalive_interval`, but really `keepalive_timeout` is the one which is strictly required for the oneshot case to be used on network change.

Then you also need to arrange for the `keepalive_manager` to start if either of `keepalive_interval` or (new) the network change hook is set. And be sure the manager is happy to start with keepalive_interval == 0.

xv-raihaan-m: I've made the requirement to have `keepalive_interval` be non-zero when the handler is set. Perhaps that should extend to the timeout as well. I think it makes more sense to have them all be non zero even in the oneshot case. So that the client can send multiple keepalives before timing out.

lightway-client/src/keepalive.rs
```diff
@@ -172,11 +191,24 @@ async fn keepalive<CONFIG: SleepManager, CONNECTION: Connection>(
...
                         // flowing only in one direction.
                         continue
-                    }
+                    },
                     Message::ReplyReceived => {
+                        state = if config.continuous() {
+                            State::Waiting
+                        } else {
+                            tracing::info!("reply received turning off network change keepalives");
+                            State::Offline
```
xv-ian-c: `::Offline` is declared as:
```
        // Not yet online
        Offline,
```

At a minimum I think we should change the comment, but perhaps it would be best to rename as `::Inactive` to properly capture the new semantics?

lightway-client/src/lib.rs
```diff
@@ -29,12 +29,13 @@ use pnet::packet::ipv4::Ipv4Packet;
...
     net::Ipv4Addr,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, Weak},
     time::Duration,
 };
 use tokio::{
     net::{TcpStream, UdpSocket},
-    sync::oneshot::Receiver,
+    sync::mpsc::Receiver as MpscReceiver,
+    sync::oneshot::Receiver as OneShotReceiver,
```
xv-ian-c: FWIW my preference  is to import the modules in this case (`use tokio::sync::{onshot, mpsc}`) and then refer to `oneshot::Receive` and `oneshot::channel` etc.

I think this is fine too if you prefer it though.

lightway-client/src/lib.rs
```diff
@@ -297,11 +304,47 @@ pub async fn inside_io_task<T: Send + Sync>(
...
+    mut network_change_signal: MpscReceiver<()>,
+    weak: Weak<Mutex<lightway_core::Connection<ConnectionState>>>,
+) {
+    while (network_change_signal.recv().await).is_some() {
+        let Some(conn) = weak.upgrade() else {
+            return;
+        };
+        if matches!(
+            conn.lock().unwrap().connection_type(),
+            ConnectionType::Datagram
```
xv-ian-c: FYI there is an `.is_datagram()` helper on `ConnectionType` so this can just be
```
if  conn.lock().unwrap().connection_type().is_datagram()
```

But given there is an else covering the only other variant a more Rustish way to write this would be:

```rust
match conn.lock().unwrap().connection_type() {
     ConnectionType::Datagram => {
            info!("sending keepalives due to network change ..");
            keepalive.network_changed().await;
     },
     ConnectionType::Stream => {
            info!("client shutting down due to network change ..");
            let _ = conn.lock().unwrap().disconnect();
            return;
     },
}
```

lightway-client/src/lib.rs
```diff
@@ -308,10 +314,10 @@ async fn handle_network_change(
     keepalive: Keepalive,
     mut network_change_signal: MpscReceiver<()>,
     weak: Weak<Mutex<lightway_core::Connection<ConnectionState>>>,
-) {
+) -> Result<()> {
```
xv-ian-c: I think you should make this `Result<ClientResult>` instead of smuggling `ClientResult::NetworkChange` through as an error and putting it back together further up the stack.

At that point maybe there is no `Err` case at all, in which case this can just return `ClientResult` itself.

lightway-client/src/main.rs
```diff
@@ -112,5 +112,8 @@ async fn main() -> Result<()> {
         keylog: config.keylog,
     };
 
-    client(config).await
+    match client(config).await {
```
xv-ian-c: I think this might be:
```rust
client(config).await.map(|_| ())
```
or
```rust
client(config).await.or(Ok(()))
```

lightway-client/src/keepalive.rs
```diff
@@ -28,11 +28,15 @@ pub trait SleepManager: Send {
...
     fn sleep_for_timeout(&self) -> impl std::future::Future<Output = ()> + std::marker::Send;
+    fn continuous(&self) -> bool;
+    fn network_change_signal_set(&self) -> bool;
 }
 
 pub struct Config {
     pub interval: Duration,
     pub timeout: Duration,
+    pub continuous: bool,
+    pub network_change_signal_set: bool,
```
xv-ian-c: This is unused by the end of the series, the usages are removed in the commit after this one.

lightway-client/src/lib.rs
xv-ian-c: ISTR that allowing `keepalive_interval` without `keepalive_timeout` was deliberate but I can't remember/find the reasoning. Something to do with best effort non-fatal attempts to keep any NAT gateways fresh maybe...

I think what you have now is fine. The other valid choice is to make it the callers responsibility since `Keepalive::network_changed` will correctly (given a 0 interval) drop the event on the floor if no timeout is specified (`if let Some(tx) = &self.tx` will be false in that case).

xv-raihaan-m: I think we can leave it as is. On network change if the caller doesn't want to timeout the keepalive operation for some reason that should be fine on our end.

lightway-client/src/keepalive.rs
xv-raihaan-m: Fixed this

lightway-client/src/lib.rs
xv-raihaan-m: Took this advice and made the change.

lightway-client/src/lib.rs
xv-raihaan-m: Made a slight modification to prevent an issue with borrowing. But it closely matches this change

lightway-client/src/lib.rs
xv-raihaan-m: Fixed this as well

PR #55, opened 2024-07-29
lightway-{client,server}: Failure to send Ctrl-C signal is not fatal

xv-ian-c:
Failing to send on a oneshot channel can only happen if the other end has died, which is already fatal (or soon will be).

Panicking in the Ctrl-C handler "will not be caught and will cause the signal handler thread to stop" according to [`ctrlc::set_handler` docs][], so the `.expect()` is not very useful.

[`ctrlc::set_handler` docs]: https://detegr.github.io/doc/ctrlc/fn.set_handler.html

### DIFF
--- a/lightway-client/src/main.rs
+++ b/lightway-client/src/main.rs
@@ -73,8 +73,8 @@ async fn main() -> Result<()> {
     let (ctrlc_tx, ctrlc_rx) = tokio::sync::oneshot::channel();
     let mut ctrlc_tx = Some(ctrlc_tx);
     ctrlc::set_handler(move || {
-        if let Some(tx) = ctrlc_tx.take() {
-            tx.send(()).expect("CtrlC handler failed");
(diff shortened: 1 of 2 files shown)


lightway-client/src/main.rs
```diff
@@ -73,8 +73,8 @@ async fn main() -> Result<()> {
     let (ctrlc_tx, ctrlc_rx) = tokio::sync::oneshot::channel();
     let mut ctrlc_tx = Some(ctrlc_tx);
     ctrlc::set_handler(move || {
-        if let Some(tx) = ctrlc_tx.take() {
-            tx.send(()).expect("CtrlC handler failed");
```
kp-mariappan-ramasamy, line 77: Interesting! Curious how did you hit or found this issue ?

xv-ian-c: I just spotted it while looking for something else.

PR #57, opened 2024-08-01
CVPN-940: Make more client apis public

kp-mariappan-ramasamy:
## Description

This PR is a refactor to expose many modules as possible to external client applications.
There is no other functional change in this PR. 

## Motivation and Context

Making different modules public will give freedom to external client application developers to reuse modules if required
without re-implementing

## How Has This Been Tested?
Ran unittests and E2E

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -35,7 +35,7 @@ use std::{
 use tokio::{
     net::{TcpStream, UdpSocket},
     sync::oneshot::Receiver,
-    task::JoinHandle,
+    task::{JoinHandle, JoinSet},
 };
 use tokio_stream::StreamExt;
 use tracing::info;
@@ -298,6 +298,8 @@ pub async fn client<A: 'static + Send + EventCallback>(
 ) -> Result<()> {
     println!("Client starting with config:\n{:#?}", &config);
 
+    let mut join_set = JoinSet::new();
+
     let (connection_type, outside_io): (ConnectionType, Arc<dyn io::outside::OutsideIO>) =
         match config.mode {
             ClientConnectionType::Datagram(maybe_sock) => {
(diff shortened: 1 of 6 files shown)


lightway-client/src/lib.rs
```diff
@@ -207,6 +207,91 @@ async fn handle_events<A: 'static + Send + EventCallback>(
     }
 }
 
+pub async fn outside_io_loop_fn(
```
xv-ian-c: Is the `_fn` bit required -- maybe a clash with the let outside_io_loop = ... outside_io_loop`? I'd have thought shadowing would have allowed that.

I'd prefer `process_outside_io` or `outside_io_task` or something to the `_fn` suffix..

lightway-client/src/lib.rs
```diff
@@ -159,21 +159,23 @@ fn debug_fmt_plugin_list(
...
 }
 
-pub struct ConnectionState {
+pub struct ConnectionState<T: Send + Sync> {
     /// Handler for tick callbacks.
     pub ticker: ConnectionTicker,
     /// InsideIpConfig received from server
     pub ip_config: Option<InsideIpConfig>,
+    /// Other extended state
+    pub extended: Option<T>,
```
xv-ian-c: Instead of making it an option external clients should use `T=()` if they have nothing to add. In fact making that the default for `T` would be reasonable IMHO.

xv-ian-c: This would also mean that `client()` would need to take a `T` so you can initialize the field..

lightway-client/src/lib.rs
kp-mariappan-ramasamy: FTR, I was not very happy with the `_fn` name either, when i write it :stuck_out_tongue_closed_eyes: 
I went with `outside_io_task` which seems nice

lightway-client/src/lib.rs
kp-mariappan-ramasamy: Using T instead of Option<T> sounds nice. Updated it.

> This would also mean that client() would need to take a T so you can initialize the field..

for this second comment, I did not address it. 
Because if a external app chose to use client() api to run, then all the core callbacks are consumed with lightway_client itself. So it was not very useful for external app to set anything other than ()
PLMK if I miss something.

xv-ian-c: Ah, ok!

PR #61, opened 2024-08-06
CVPN-942 Restart DTLS Connection on Double Handshake Error

xv-thomas-leong:
## Description

The observed behaviour in client when this error happens is that the
connection will hang indefinitely until it is manually restarted. There
is nothing that the client can do to resurrect the failing state.

Handling the error as fatal will trigger early termination of the
client, which would allow a faster restart.

## How Has This Been Tested?

CI Tested

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -194,6 +194,7 @@ impl ConnectionError {
                     Goodbye => true,
                     WireError(_) => true,
                     WolfSSL(wolfssl::Error::Fatal(ErrorKind::DomainNameMismatch)) => true,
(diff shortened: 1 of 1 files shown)


Cargo.lock
expressvpn-andre-l, line 1: I had to compare the commit hashes at https://github.com/expressvpn/wolfssl-rs/commits/main/ to understand why you bumped WolfSSL.

Could you explain in the commit message why the bump was needed? i.e. Need for WolfSSL to expose a new error type

xv-thomas-leong: Thanks for the comment. Addressed in 8a3b126

expressvpn-andre-l: Thanks! Resolving the thread

PR #62, opened 2024-08-07
server: Detect local address for new connections

xv-ian-c:
## Description

Add the local address (i.e. the address the client connected to) to the server's `ConnectionState` for use in other parts of the system (e.g. authentication, internal IP allocation, although neither of these are tackled here).

## Motivation and Context

On a server with multiple interfaces I would like to configure to listen on `0.0.0.0:PORT` to support connections to any IP address. The listening part works as is (in fact the e2e tests are configured that way), but it is not possible to know which address the client connected to.

## How Has This Been Tested?

I've manually tested using the local test setup that I can see the correct address for both UDP and TCP.

The existing e2e tests will validate no regressions.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/io/outside/udp.rs
+++ b/lightway-server/src/io/outside/udp.rs
@@ -275,6 +275,7 @@ impl Server for UdpServer {
 
                     let Some(addr) = sock_addr.as_socket() else {
                         // Since we only bind to IP sockets this shouldn't happen.
+                        metrics::udp_recv_invalid_addr();
                         return Err(std::io::Error::new(
                             std::io::ErrorKind::InvalidInput,
                             "failed to convert local addr to socketaddr",
@@ -302,7 +303,10 @@ impl Server for UdpServer {
                                         _ => None,
                                     }
                                 }) else {
-                                    // Since we have a bound socket this shouldn't happen.
+                                    // Since we have a bound socket
+                                    // and we have set IP_PKTINFO
+                                    // sockopt this shouldn't happen.
+                                    metrics::udp_recv_missing_pktinfo();
                                     return Err(std::io::Error::new(
                                         std::io::ErrorKind::Other,
                                         "recvmsg did not return IP_PKTINFO",
--- a/lightway-server/src/metrics.rs
+++ b/lightway-server/src/metrics.rs
@@ -28,6 +28,8 @@ const METRIC_UDP_SESSION_ROTATION_FINALIZED: &str = "udp_session_rotation_finali
 const METRIC_UDP_SESSION_ROTATION_ATTEMPTED_VIA_REPLAY: &str =
     "udp_session_rotation_attempted_via_replay";
 const METRIC_UDP_RECV_TRUNCATED: &str = "udp_recv_truncated";
+const METRIC_UDP_RECV_INVALID_ADDR: &str = "udp_recv_invalid_addr";
+const METRIC_UDP_RECV_MISSING_PKTINFO: &str = "udp_recv_missing_pktinfo";
 
 // Connection performance
 const METRIC_TO_LINK_UP_TIME: &str = "to_link_up_time";
@@ -192,6 +194,14 @@ pub(crate) fn udp_no_header() {
 
 pub(crate) fn udp_recv_truncated() {
     counter!(METRIC_UDP_RECV_TRUNCATED).increment(1);
+}
+
+pub(crate) fn udp_recv_invalid_addr() {
+    counter!(METRIC_UDP_RECV_INVALID_ADDR).increment(1);
+}
+
+pub(crate) fn udp_recv_missing_pktinfo() {
+    counter!(METRIC_UDP_RECV_MISSING_PKTINFO).increment(1);
 }
 
 /// Fatal TLS error for [`lightway_core::Connection`].
--- a/docs/logs_and_metrics.md
+++ b/docs/logs_and_metrics.md
@@ -40,6 +40,8 @@ Lightway server also supports metrics to monitor. The following are the metrics 
 | udp_conn_recovered_via_session | server | Counter | Counts UDP connections which have been recovered using the session ID (which indicates that the client’s IP address changed) |
 | udp_session_rotation_attempted_via_replay | server | Counter | Counts UDP rotation attempted using duplicated packets. i.e An attack<br>ie. Some adversary capture and replay packets from different IP address.<br><br>There is also a possibility that counter is incremented due to aggressive connect.<br> |
 | udp_recv_truncated | server | Counter | Counts occurrences of UDP packet truncation on receive |
+| udp_recv_invalid_addr | server | Counter | Counts failures to retrieve a valid socket address from `recvmsg` syscall |
+| udp_recv_missing_pktinfo | server | Counter | Counts failures to find a valid `PKTINFO` control message in `recvmsg` result |
 | udp_bad_packet_version | server | Counter | Counts UDP packets where the the version in the wire protocol header was not a version supported by the server |
 | udp_rejected_session | server | Counter | Counts UDP packets which were rejected due to the session id in the wire protocol header not being recognised |
 | udp_parse_wire_failed | server | Counter | Counts UDP packets which could not be parsed. Indicates plugin ingress chain failed |


lightway-server/src/io/outside/udp.rs
```diff
@@ -179,9 +180,31 @@ impl Server for UdpServer {
...
+                    unsafe {
+                        buf.set_len(len)
+                    };
+
+                    let Some(addr) = addr.as_socket() else {
+                        // Since we only bind to IP sockets this shouldn't happen.
+                        tracing::debug!("Failed to convert local addr to socketaddr");
+                        return Err(std::io::Error::new(
+                            std::io::ErrorKind::InvalidInput,
+                            "invalid argument",
```
kp-mariappan-ramasamy: In case this happens, this error message will not be very helpful. Can we consider adding reason like the debug log above ?

lightway-server/src/io/outside/udp.rs
```diff
@@ -226,6 +259,28 @@ impl Server for UdpServer {
...
                     };
 
+                    #[allow(unsafe_code)]
+                    let Some(local_addr) =
+                        // SAFETY: The call to `recvmsg` above updated
+                        // the control buffer length field.
+                        unsafe { control.iter(control_len) }.find_map(|cmsg| {
+                            match cmsg {
+                                cmsg::Message::IpPktinfo(pi) => {
+                                    let ipv4 = std::net::Ipv4Addr::from(pi.ipi_spec_dst.s_addr.to_ne_bytes());
```
kp-mariappan-ramasamy: Is it not `.to_be_bytes()` ?

lightway-server/src/io/outside/udp.rs
```diff
@@ -46,9 +50,31 @@ impl OutsideIOSendCallback for UdpSocket {
     }
 }
 
+#[allow(unsafe_code)]
+fn socket_enable_pktinfo(sock: &tokio::net::UdpSocket) -> Result<()> {
```
kp-mariappan-ramasamy: We also have lightway-client::io::udp::sockopt module.
We should soon consider moving this socket options set/get to common place.

lightway-server/src/io/outside/udp.rs
```diff
@@ -167,13 +207,90 @@ impl UdpServer {
 #[async_trait]
 impl Server for UdpServer {
     async fn run(&mut self) -> Result<()> {
-        info!("Accepting traffic on {}", self.sock.local_addr()?);
+        info!("Accepting traffic on {}", self.local_addr);
         loop {
             let mut buf = BytesMut::with_capacity(MAX_OUTSIDE_MTU);
 
-            let (_len, addr) = self.sock.recv_buf_from(&mut buf).await?;
+            let (addr, local_addr) = self
```
kp-mariappan-ramasamy, line 221: Should we consider skipping this cmsg parsing for udp server, which are not binded to `0.0.0.0` ?

lightway-server/src/connection.rs
```diff
@@ -22,6 +22,8 @@ pub(crate) struct ConnectionState {
     ticker: ConnectionTicker,
     // The backend IP (from IP pool) associated with this connection
     pub(crate) ip: Option<Ipv4Addr>,
+    // The local IP which the client has connected to
```
kp-mariappan-ramasamy:
```suggestion
    // The local SocketAddr which the client has connected to
```
or 
```suggestion
    // The local IP:port which the client has connected to
```
 ?

lightway-server/src/connection.rs
```diff
@@ -22,6 +22,8 @@ pub(crate) struct ConnectionState {
     ticker: ConnectionTicker,
     // The backend IP (from IP pool) associated with this connection
     pub(crate) ip: Option<Ipv4Addr>,
```
kp-mariappan-ramasamy: I think we could also rename this field appropriately, now that we have one more IP(:port) field in this struct

lightway-server/src/io/outside/udp.rs
xv-ian-c: Good question. [`Ipv4Addr::from`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#impl-From%3C%5Bu8;+4%5D%3E-for-Ipv4Addr) doesn't really specify the byte order. If I use big endian then I get `local_addr=1.0.16.172:27690` instead of the expected `local_addr=172.16.0.1:27690` which I get with native endianess.

The [from for `u32` method](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.from-2) does say host byte order. Probably I should be using that anyway since `pi.ipi_spec_dst.s_addr` is (via some type aliases) a `u32` anyway.

lightway-server/src/io/outside/udp.rs
xv-ian-c: agreed. will look into that.

lightway-server/src/io/outside/udp.rs
xv-ian-c: I'll see how much plumbing that would take, would be nice to avoid the overhead I guess.

lightway-server/src/io/outside/udp.rs
xv-ian-c: actually, `s_addr` from the kernel is in network byte order (be), so `From::<u32>` doesn't work directly.

There is no `to_ne` on u32, only `to_ne_bytes` so I think we have to use that and assume that `Ipv4Addr::from::<[u8; 4]()` accepts native endian bytes like most other `Ipv4Addr::from` functions and `Ipv4Addr::from_bits` do.

lightway-server/src/io/outside/udp.rs
xv-ian-c: I'll move the debug message wording into the error.

lightway-server/src/connection.rs
xv-ian-c: I went with the latter, since the `SocketAddr` is already the type of the field.

lightway-server/src/connection.rs
xv-ian-c: Good idea. I went with `internal_ip`

lightway-server/src/io/outside/udp.rs
```diff
@@ -167,13 +219,110 @@ impl UdpServer {
...
+                    #[allow(unsafe_code)]
+                    let local_addr = match self.bind_mode {
+                        BindMode::UnspecifiedAddress{ local_port } => {
+                            let Some(local_addr) =
+                                // SAFETY: The call to `recvmsg` above updated
+                                // the control buffer length field.
+                                unsafe { control.iter(control_len) }.find_map(|cmsg| {
+                                    match cmsg {
+                                        cmsg::Message::IpPktinfo(pi) => {
+                                            let ipv4 = std::net::Ipv4Addr::from(pi.ipi_spec_dst.s_addr.to_ne_bytes());
```
kp-mariappan-ramasamy: We already use `std::net::SockAddr`
We might as well import Ipv4Addr ?

lightway-server/src/metrics.rs
```diff
@@ -27,6 +27,9 @@ const METRIC_UDP_SESSION_ROTATION_BEGIN: &str = "udp_session_rotation_begin";
 const METRIC_UDP_SESSION_ROTATION_FINALIZED: &str = "udp_session_rotation_finalized";
 const METRIC_UDP_SESSION_ROTATION_ATTEMPTED_VIA_REPLAY: &str =
     "udp_session_rotation_attempted_via_replay";
+const METRIC_UDP_RECV_TRUNCATED: &str = "udp_recv_truncated";
+const METRIC_UDP_RECV_INVALID_ADDR: &str = "udp_recv_invalid_addr";
+const METRIC_UDP_RECV_MISSING_PKTINFO: &str = "udp_recv_missing_pktinfo";
```
kp-mariappan-ramasamy, line 32: We need to update the docs here:
https://github.com/expressvpn/lightway/blob/main/docs/logs_and_metrics.md

lightway-server/src/io/outside/udp.rs
xv-ian-c: Yes, I spotted a couple of cases where I'd used the full path for SocketAddr too even though it is imported.

lightway-server/src/io/outside/udp.rs
kp-mariappan-ramasamy: I wonder whether we should use [bytemuck](https://docs.rs/bytemuck/latest/bytemuck/fn.cast.html) to convert to [u8;4] to be clearer instead of `pi.ipi_spec_dst.s_addr.to_ne_bytes()`
```rust
let ipv4_bytes: [u8; 4] = bytemuck::cast(pi.ipi_spec_dst.s_addr);
```

In any case, I think it deserves to be moved to a separate line, with a clear comment on why we did it. 
This is a common pitfall

xv-ian-c: I don't think we need bytemuck when there is a suitable method o t he primitive type.

I think the `to_ne_bytes` name is making us both think there is more going on here than there is. It's not actually converting anything to native order, it's just a type cast of the memory underlying the u32 which is already in native order. I guess it's not named `as_bytes` or `to_bytes` to avoid confusion with `to_(le|be)_bytes`.

xv-ian-c: I might be wrong here... I'm going to figure out what is actually happening and write a comment.

xv-ian-c: I was wrong, it is actually changing byte order too it seems.

I figured out it was clearer to use `u32::from_be()` and `IPv4Addr::from_bits()` along with a comment:
```rust
                                    // From https://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html
                                    // the `s_addr` is an `in_addr`
                                    // which is in network byte order
                                    // (big endian).
                                    let ipv4 = u32::from_be(pi.ipi_spec_dst.s_addr);
                                    let ipv4 = Ipv4Addr::from_bits(ipv4);
                                    let ip = IpAddr::V4(ipv4);
```

kp-mariappan-ramasamy: The description of to_ne_bytes is not very clear. `native byte order` trips everytime when i read it.
```
Return the memory representation of this integer as a byte array in native byte order.
```
~~But I believe the api does what we need.~~

Ignore this, I typed this without refreshing and seeing your last two messages

lightway-server/src/io/outside/udp.rs
```diff
@@ -167,13 +214,114 @@ impl UdpServer {
...
+                                unsafe { control.iter(control_len) }.find_map(|cmsg| {
+                                    match cmsg {
+                                        cmsg::Message::IpPktinfo(pi) => {
+                                            // From https://pubs.opengroup.org/onlinepubs/009695399/basedefs/netinet/in.h.html
+                                            // the `s_addr` is an `in_addr`
+                                            // which is in network byte order
+                                            // (big endian).
+                                            let ipv4 = u32::from_be(pi.ipi_spec_dst.s_addr);
+                                            let ipv4 = Ipv4Addr::from_bits(ipv4);
+                                            let ip = IpAddr::V4(ipv4);
```
kp-mariappan-ramasamy, line 300: This seems more saner. Thanks!

lightway-server/src/io/outside/udp.rs
xv-ian-c: I read "ne" as "network endian" every single time...

lightway-server/src/io/outside/udp.rs
```diff
@@ -178,7 +179,10 @@ impl Server for UdpServer {
         loop {
             let mut buf = BytesMut::with_capacity(MAX_OUTSIDE_MTU);
 
-            let (_len, addr) = self.sock.recv_buf_from(&mut buf).await?;
+            let (_len, addr) = self
+                .sock
+                .async_io(Interest::READABLE, || self.sock.try_recv_buf_from(&mut buf))
```
xv-ian-c: I missed that in [`tokio::net::UdpSocket::async_io`](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.async_io) the docs say:
> The closure should not perform the IO operation using any of the methods defined on the Tokio UdpSocket type, as this will mess with the readiness flag and can cause the socket to behave incorrectly.

Which this use of `try_recv_buf_from` violates (although it appeared to work in the limited testing this commit got in isolation).

Fortunately this was just a step along the way and the very next commit switches to `socket2::SockRef::recv_from` which puts us back in to compliance.

PR #67, opened 2024-08-15
Revert "[auto] Update Rust toolchain to 1.80.1"

xv-ian-c:
This reverts commit 9d643e98494f5d3ae7ea8e6f7f9a1ec2c5b15e10.

We have seen strange issue in CI since this update, e.g. https://github.com/expressvpn/lightway/actions/runs/10400777425/job/28801984468?pr=63 which failed because it looks like `lightway-server` build was using a different version of `lightway-core` (which didn't have the new generic type on `ServerAuth`).

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -6,7 +6,7 @@ ARG --global debian = bookworm
 IMPORT github.com/earthly/lib/rust:a49d2a0f4028cd15666d19904f8fc5fbd0b9ba87 AS lib-rust
 
 install-build-dependencies:
-    FROM rust:1.80.1-$debian
+    FROM rust:1.80.0-$debian
     WORKDIR /lightway
     RUN dpkg --add-architecture arm64
     RUN apt-get update -qq


Earthfile
```diff
@@ -6,7 +6,7 @@ ARG --global debian = bookworm
 IMPORT github.com/earthly/lib/rust:a49d2a0f4028cd15666d19904f8fc5fbd0b9ba87 AS lib-rust
 
 install-build-dependencies:
-    FROM rust:1.80.1-$debian
+    FROM rust:1.80.0-$debian
```
kp-mariappan-ramasamy, line 9: Our nightly bot will again create the PR to change it back. But we can keep that one open.

https://github.com/expressvpn/lightway/blob/main/.github/workflows/nightly-rust-update-check.yaml

WDYT about adding a step in job to ignore this versions ?

xv-ian-c: I think it's ok to have an open PR as a reminder that we are behind, plus it'll do a semi regular run with 1.80.1 which would help validate a bit if the issue gets fixed by some other change.

PR #69, opened 2024-08-19
Add per-incoming address ip pools

xv-ian-c:
## Description

Add support for mapping from the incoming IP address (i.e. that which the client connected to) and subranges of the "ip_pool", i.e. the backend network.

Most of the PR is about refactoring the existing ip manager arrangements to allow for this possibility, in particular factoring out the core per-subnet IP allocator (now call `IpPool`).

## Motivation and Context

This allows a server to be configured to listen on 0.0.0.0 on a system with multiple valid IP addresses and to use different backend ranges for each actual incoming IP. Together with an appropriate SNAT configuration on the system this can allow mapping to the desired outgoing IP.

## How Has This Been Tested?

Unit tests and manual tests observing correct backend IP range used.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/ip_manager.rs
+++ b/lightway-server/src/ip_manager.rs
@@ -3,7 +3,7 @@ mod ip_pool;
 use ipnet::Ipv4Net;
 use lightway_core::{InsideIpConfig, ServerIpPool};
 use std::collections::HashMap;
-use std::net::Ipv4Addr;
+use std::net::{IpAddr, Ipv4Addr};
 use std::sync::{Arc, RwLock};
 use tracing::info;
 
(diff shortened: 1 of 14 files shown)


lightway-server/src/ip_manager.rs
```diff
@@ -26,6 +26,9 @@ pub(crate) struct IpManager<T = Arc<Connection>> {
 struct IpManagerInner<T> {
     /// Client IP to lightway::Connection hashmap
     ip_to_conn_map: HashMap<Ipv4Addr, T>,
+    /// Mapping of pool's for specific IPs
+    #[allow(dead_code)]
```
xv-raihaan-m: Do we need this annotation? It looks like we do infact use this variable

xv-ian-c: It is indeed stale. Will fix.

PR #79, opened 2024-09-12
lightway-app-utils: iouring: Rework to push more processing into uring thread

xv-ian-c:
## Description

This removes the need for the send and receive tasks posting sqe's to the thread for it to post by doing that as part of the loop. It also pulls the cqe processing into the same loop.

This now uses a model similar to that in most of the uring examples (in both C and Rust).

This change improves an iperf upload test from 2.03 Gbits/sec to 2.19 Gbigs/sec Gbits/sec and download from 2.34 Gbits/sec to 2.46 Gbits/sec.

The channel used for the tx and rx queues is changed from the async_channel to tokio::mpsc. The performance difference appears to be in the noise, however we no longer need the multi-consumer property (in fact we'd be happy with a spsc channel if there was one) and we use tokio::mpsc everywhere else.

## Motivation and Context

Trying to improve performance, following on from #78 (cumulatively upload perf has gone from 1.85->2.19 Gbits/sec and  download from 2.21 -> 2.46 Gbits/sec).

## How Has This Been Tested?

Using iperf in the containerized test setup on an otherwise idle c7i.8xlarge EC2 node.

Numbers quoted above are with both client and server using a 1024 entry io uring.

## Types of changes

Performance improvement.

## Checklist:

### DIFF
--- a/lightway-app-utils/src/iouring.rs
+++ b/lightway-app-utils/src/iouring.rs
@@ -1,125 +1,48 @@
 use anyhow::Result;
-use async_channel::{bounded, Receiver, Sender};
-use bytes::{BufMut, BytesMut};
-use dashmap::DashMap;
+use bytes::{BufMut, Bytes, BytesMut};
 use thiserror::Error;
 
 use crate::metrics;
 use io_uring::{
-    cqueue::Entry as CEntry, opcode, squeue::Entry as SEntry, types::Fixed, CompletionQueue,
-    IoUring,
+    cqueue::Entry as CEntry, opcode, squeue::Entry as SEntry, types::Fixed, IoUring,
+    SubmissionQueue, Submitter,
 };
 use std::{
     os::fd::{AsRawFd, RawFd},
-    sync::{
-        atomic::{AtomicU64, Ordering},
-        Arc,
-    },
-    thread::{self, JoinHandle},
+    sync::Arc,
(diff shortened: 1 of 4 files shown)


lightway-app-utils/src/iouring.rs
```diff
@@ -183,136 +86,372 @@ impl<T: AsRawFd> IOUring<T> {
...
+
+                    sq.sync();
+
+                    continue 'io_loop;
+                }
+
+                Ok(a) = event_fd.read(&mut completed_number) => {
+                    metrics::tun_iouring_idle_thread_time(start_time.elapsed());
+                    metrics::tun_iouring_wake_eventfd();
+                    assert_eq!(a, 8);
```
xv-raihaan-m, line 349: Did you intend to keep this assert?

xv-ian-c: Yes, it was there in the original code and is checking a property of the [eventfd](https://man7.org/linux/man-pages/man2/eventfd.2.html) API:
> Each successful [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) returns an 8-byte integer.  A
>  [read(2)](https://man7.org/linux/man-pages/man2/read.2.html) fails with the error EINVAL if the size of the
> supplied buffer is less than 8 bytes.

PR #87, opened 2024-09-26
lightway-core: Accept inside and outside data by reference

xv-ian-c:
## Description

In both cases this could allow the application to reuse buffers in the I/O loop(s). That is achieved here for the outside I/O case but I was unable to find a workable solution for the uring case of inside I/O due to the separate processing thread (returning buffers over another channel ends up _much_ slower).
 
## Motivation and Context

The allocations and frees do show up in the flamegraphs, but the performance impact of this is marginal/in the noise.

This originally came out of the native I/O uring PoC in #88 where it did make an interesting difference.

## How Has This Been Tested?

I ran some iperf tests using the containerized environment and also ran some speedtests with two machines directly connected with a cross over test against a nearby speedtest server.

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -293,7 +293,7 @@ pub async fn inside_io_task<T: Send + Sync>(
             };
         }
 
-        match conn.inside_data_received(buf) {
+        match conn.inside_data_received(&mut buf) {
             Ok(()) => {}
             Err(ConnectionError::PluginDropWithReply(reply)) => {
                 // Send the reply packet to inside path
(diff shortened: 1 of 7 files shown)


lightway-client/src/lib.rs
```diff
@@ -232,12 +232,15 @@ pub async fn outside_io_task<T: Send + Sync>(
     outside_io: Arc<dyn io::outside::OutsideIO>,
     keepalive: Keepalive,
 ) -> Result<()> {
+    let mut buf = BytesMut::with_capacity(mtu);
```
xv-raihaan-m, line 235: if `MAX_OUTSIDE_MTU` is greater or equal than `mtu` is it not just better to reserve `MAX_OUTSIDE_MTU` here?

xv-ian-c: This effectively is just code motion, the size was `mtu` previously and I think that's the correct one since it is the value the user has told us to use.

My mistake was below where I should have done `buf.reserve(mtu)` not `MAX_OUTSIDE_MTU`.

PR #95, opened 2024-10-02
CVPN-1185: Ignore network unreachable socket error for linux-like OS when sending out packets from the udp socket of lightway-client

kp-max-li:
## Description
Ignore network unreachable socket error for linux-like OS when sending out packets from the udp socket of lightway-client. This change only applies to linux-like OS. Other OS like WIndows may need a different implementation.

## Motivation and Context
if this error is not ignored and passed to the WolfSSL layer, the WolfSSL layer will enter a fatal socket error state and stop the inside IO loop during network change. Ignoring this error to prevent the inside IO loop from exiting.

## How Has This Been Tested?
The change has been dev tested in two environments: an android emulator (Pixel Fold API 35) and a physical android phone (Samsung A33).
Without the change, the error can be reproduced in both environments by switching off WiFi.
With the change, the error can no longer be reproduced in both environments.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -82,6 +82,22 @@ impl OutsideIOSendCallback for Udp {
                 // This way we can continue if/when the server shows up.
                 IOCallbackResult::Ok(0)
             }
+            Err(err) if matches!(err.raw_os_error(), Some(libc::ENETUNREACH)) => {
+                // This case indicates network unreachable error.
+                // Possibly there is a network change at the moment.
+                //
+                // Swallow the socket error so the error is not passed to the
+                // WolfSSL layer. Then the WolfSSL layer would not enter a
+                // fatal error state
+                //
+                // Note: this case should be matched by
+                // matches!(err.kind(), std::io::ErrorKind::NetworkUnreachable)
+                // However, at the time of implementing this match case,
+                // the version of rust we use does not support NetworkUnreachable.
+                // This match case can be rewritten to use NetworkUnreachable
+                // once rust is updated to support it.
+                IOCallbackResult::Ok(0)
+            }
             Err(err) => IOCallbackResult::Err(err),
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1339,6 +1339,7 @@ dependencies = [
  "ctrlc",
  "educe",
  "futures",
+ "libc",
  "lightway-app-utils",
  "lightway-core",
  "more-asserts",
--- a/lightway-client/Cargo.toml
+++ b/lightway-client/Cargo.toml
@@ -26,6 +26,7 @@ clap.workspace = true
 ctrlc.workspace = true
 educe.workspace = true
 futures = "0.3.30"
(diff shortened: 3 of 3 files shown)


lightway-client/src/io/outside/udp.rs
```diff
@@ -82,6 +82,27 @@ impl OutsideIOSendCallback for Udp {
...
+                target_os = "android",
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "ios"
+            ))]
+            Err(err) if matches!(err.raw_os_error().unwrap_or(0), libc::ENETUNREACH) => {
+                // This case indicates network unreachable error.
+                // Possibly there is a network change at the moment.
+                //
+                // Swallow the error so the WolfSSL socket does not
```
xv-raihaan-m: Comment is inaccurate, wolfSSL itself does not own a socket. This change is to ensure that this error is not returned up to the wolfSSL layer which assumes any socket error is fatal.

kp-max-li: Makes sense. I will make a change.

kp-max-li: Done. I have also changed the PR's description and commit message.

lightway-client/src/io/outside/udp.rs
```diff
@@ -82,6 +82,28 @@ impl OutsideIOSendCallback for Udp {
                 // This way we can continue if/when the server shows up.
                 IOCallbackResult::Ok(0)
             }
+            #[cfg(any(
+                target_os = "android",
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "ios"
+            ))]
```
kp-mariappan-ramasamy: None of our code has been adapted for windows platform.
Since it covers all linux and unix variants (atleast the ones we care about) I am thinking whether we can remove this. 

We can handle windows case later, when we add support.

kp-mariappan-ramasamy: For now, we only target for Linux platform. We can handle theses cases later instead of it partially.
Supported platforms: https://github.com/expressvpn/lightway?tab=readme-ov-file#supported-platforms

kp-max-li: Agreed. I shall do an update to this.

lightway-client/src/io/outside/udp.rs
```diff
@@ -82,6 +82,22 @@ impl OutsideIOSendCallback for Udp {
                 // This way we can continue if/when the server shows up.
                 IOCallbackResult::Ok(0)
             }
+            Err(err) if matches!(err.raw_os_error().unwrap_or(0), libc::ENETUNREACH) => {
```
kp-mariappan-ramasamy: This can be rewritten like below instead of unwrapping with 0 as default

```suggestion
            Err(err) if matches!(err.raw_os_error(), Some(libc::ENETUNREACH)) => {
```

kp-max-li: Yeah that's better. 0 felt like a magic number here. I'll change it.

PR #98, opened 2024-10-04
lightway-server: Make it non fatal if a new clients cannot be accepted

xv-ian-c:
## Description

If we cannot accept a new incoming connection we may still be able to carry on servicing existing connections. At least there is no real harm in trying... 

## Motivation and Context

I was doing some testing of a TCP server with many clients and had forgotten to increase `RLIMIT_NOFILE`.

The result was the server existed with:
```
Error: Outside IO loop exited

Caused by:
    0: ConnectionBuilder Error: WolfSSL Error: Failed to allocate WolfSSL Session
    1: WolfSSL Error: Failed to allocate WolfSSL Session
    2: Failed to allocate WolfSSL Session
```

(ultimately caused by a failure to open `/dev/urandom` due to not having a spare fd).

## How Has This Been Tested?

By setting `RLIMIT_NOFILE` back down to the default (1024) and starting 1013 clients (there are other open fds than the socket ones).

## Types of changes

## Checklist:

### DIFF
--- a/docs/logs_and_metrics.md
+++ b/docs/logs_and_metrics.md
@@ -26,6 +26,7 @@ Lightway server also supports metrics to monitor. The following are the metrics 
 
 | Metric Name | Source | Metric Type | Definition |
 | ------------------------------------------------------------------- | ------ | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


lightway-server/src/connection_manager.rs
kp-mariappan-ramasamy, line 1: Minor update needed in comment:
```
Previously if we failed to create a connection failed the TCP server would exit with:
```

to
```
Previously if we failed to create a connection, the TCP server would exit with:
```

PR #100, opened 2024-10-07
server: Assign IPs to clients in LRU order

xv-ian-c:
## Description

We currently store available IPs in a `HashSet`, this means that the current effective (but unspecified) behaviour of the `HashSet` was to allocate the most recently freed IP. This can make certain attacks based around guessing a target's internal IP address somewhat easier to pull off.

## Motivation and Context

Makes certain classes of attack against VPN clients harder to pull off.

## How Has This Been Tested?

Unit test updates are included.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/ip_manager/ip_pool.rs
+++ b/lightway-server/src/ip_manager/ip_pool.rs
@@ -1,4 +1,5 @@
 use ipnet::Ipv4Net;
+use rand::prelude::*;
 use std::{
     collections::{HashSet, VecDeque},
     net::Ipv4Addr,
@@ -24,10 +25,16 @@ impl IpPool {
                 .chain(std::iter::once(ip_pool.broadcast())),
         );
 
-        let available_ips = ip_pool
+        let mut available_ips: VecDeque<_> = ip_pool
             .hosts()
             .filter(|ip| !reserved_ips.contains(ip))
             .collect();
+
+        // make it harder to guess IPs during the early life of a
+        // server.
+        available_ips
+            .make_contiguous()
+            .shuffle(&mut rand::thread_rng());
 
         Self {
             reserved_ips,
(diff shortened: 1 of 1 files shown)


lightway-server/src/ip_manager/ip_pool.rs
```diff
@@ -339,6 +346,36 @@ mod tests {
...
+        //
+        // The average delta between two consecutive allocations from
+        // a non-sequential list should be a non-small number.
+        //
+        // If the list were sorted then the average would be 1.
+        //
+        // Given the 2^16 entry IP pool here the average delta in
+        // practice is 21-22,000 (about 1/3 of the pool size). The
+        // chances of this coming out as less than 512 in practice are
+        // miniscule.
```
kp-mariappan-ramasamy, line 366: Sounds like dark ma(th)ic :exploding_head:

lightway-server/src/ip_manager/ip_pool.rs
```diff
@@ -30,27 +30,21 @@ impl IpPool {
...
-            ip_pool,
             reserved_ips,
+            allocated_ips: Default::default(),
             available_ips,
         }
     }
 
     pub fn allocate_ip(&mut self) -> Option<Ipv4Addr> {
-        self.available_ips.pop_front()
+        // self.available_ips.pop_front()
```
kp-mariappan-ramasamy: nit!: remove the commented line

lightway-server/src/ip_manager/ip_pool.rs
xv-ian-c: Yeah.

Intuitively it seems right to me that the average delta between two random points in a space would be 1/3 of the total space, but I couldn't find any actual maths to back that up...

PR #91, opened 2024-09-26
Support a simple user database

xv-ian-c:
TODO:
- [x] rebase with merge of https://github.com/expressvpn/lightway/pull/90

## Description

Adds support to `lightway-server` for a simple user database following the htpasswd syntax.

For `lightway-client` make a user and password mandatory.

## Motivation and Context

A single user/pass is rather restrictive.

## How Has This Been Tested?

Covered by E2E tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] The correct base branch is being used, if not `main`

PR #103, opened 2024-10-08
Relax permissions checks on configuration files

xv-ian-c:
## Description

In 23c30e61c7e6 ("lightway-app-utils: Use `fs_mistrust` to validate config file safety.") restrictions were added to ensure that configuration files and secrets had permissions which would prevent another user from tampering with them. This also required that the files were not readable by other users.

As of b8e2fc51e0ef ("lightway-server: Support multiple users") the configuration file no longer contains any sensitive secrets so we can be more tolerant of the file being readable.

We continue to enforce non-readableness for the server key and the user database. We also add the missing checks for the server certificate, which we allow to be readable.

## Motivation and Context

#90 was rather restrictive in hindsight and since #91 there is no solid reason to be so restrictive for all files.

## How Has This Been Tested?

E2E has been adjusted

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/utils.rs
+++ b/lightway-app-utils/src/utils.rs
@@ -3,15 +3,30 @@ use std::path::PathBuf;
 use anyhow::Result;
 use fs_mistrust::Mistrust;
 
+/// Specifies the limits which `validate_configuration_file_path`
+/// should enforce.
+pub enum Validate {
+    /// Ensure that only the file's owner can access (read or write)
+    OwnerOnly,
+    /// Ensure that only the file's owner can write but allow anyone
+    /// to read
+    AllowWorldRead,
+}
+
 /// Verifies whether input `path` points to a valid configuration file
 /// present in filesystem. Validates that the ownership permissions
 /// are acceptable.
-pub fn validate_configuration_file_path(path: &PathBuf) -> Result<()> {
-    Mistrust::builder()
+pub fn validate_configuration_file_path(path: &PathBuf, validate: Validate) -> Result<()> {
+    let mistrust = Mistrust::builder()
         .controlled_by_env_var("LW_DANGEROUSLY_DISABLE_PERMISSIONS_CHECKS")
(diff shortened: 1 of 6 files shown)


lightway-app-utils/src/utils.rs
```diff
@@ -6,12 +6,18 @@ use fs_mistrust::Mistrust;
 /// Verifies whether input `path` points to a valid configuration file
 /// present in filesystem. Validates that the ownership permissions
 /// are acceptable.
-pub fn validate_configuration_file_path(path: &PathBuf) -> Result<()> {
-    Mistrust::builder()
+pub fn validate_configuration_file_path(path: &PathBuf, allow_readable: bool) -> Result<()> {
```
kp-mariappan-ramasamy: Wonder whether enum instead of boolean is more readable in the callers.

xv-ian-c: It would be nice (although the benefit is really only materialized if there are multiple options which could be confused). I'm really struggling with the enum variant naming though?

`Validate::OwnerOnly` and `Validate::WorldReadable` are the best I've come up with, but I don't really like them. Ideas?

xv-ian-c: Here's what I went with:
```rust
/// Specifies the limits which `validate_configuration_file_path`
/// should enforce.
pub enum Validate {
    /// Ensure that only the file's owner can access (read or write)
    OwnerOnly,
    /// Ensure that only the file's owner can write but allow anyone
    /// to read
    AllowWorldRead,
}
```

kp-mariappan-ramasamy: I would have suggested, `AllowWorldRead`, `BlockWorld`
But your changes looks fine too.

I think we could merge both commits into one?

xv-ian-c: I kept it separate in case we decided it looked worse I've now merged.

PR #104, opened 2024-10-09
app-utils: add tracing at unexpected errors

kp-mariappan-ramasamy:
## Description
There are few places where we return error without logging. Since main_task run from a separate thread, the error returned by the main_task is not catched or logged.

Log it when there are unexpected errors and also when main_task exits

## Motivation and Context

Enable debugging in case of iouring errors

### DIFF
--- a/lightway-app-utils/src/iouring.rs
+++ b/lightway-app-utils/src/iouring.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{anyhow, Context, Result};
 use bytes::{BufMut, Bytes, BytesMut};
 use lightway_core::IOCallbackResult;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


lightway-app-utils/src/iouring.rs
```diff
@@ -251,6 +260,20 @@ fn push_rx_events_to(
...
+    ring_size: usize,
+    mtu: usize,
+    tx_queue: mpsc::Receiver<Bytes>,
+    rx_queue: mpsc::Sender<BytesMut>,
+) -> Result<()> {
+    _main_task(fd, ring_size, mtu, tx_queue, rx_queue)
+        .await
+        .inspect_err(|err| {
+            tracing::error!("i/o uring task stopped: {:?}", err);
+        })
```
xv-ian-c: I think you can do this as part of the `block_on` call which calls `main_task` and avoid the extra layer of indirection.

BTW, `_FOO` in Rust has a meaning (it disables the dead code lint) which I think you probably don't want, so if there is some reason why you can't move this then I'd suggest naming the inner main as `iouring_io_loop` or something.

PR #106, opened 2024-10-11
server: tcp: support haproxy protocol

xv-ian-c:
## Description

Using the PROXY protocol: https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

## Motivation and Context

This allows configuration which involve a load balancer or other proxy.

## How Has This Been Tested?

Instructions are included in the README.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/io/outside/tcp.rs
+++ b/lightway-server/src/io/outside/tcp.rs
@@ -8,6 +8,7 @@ use lightway_core::{
     MAX_OUTSIDE_MTU,
 };
 use socket2::SockRef;
+use tokio::io::AsyncReadExt as _;
 use tracing::{debug, info, instrument, warn};
 
 use crate::{connection_manager::ConnectionManager, metrics};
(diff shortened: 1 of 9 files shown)


lightway-server/src/io/outside/tcp.rs
```diff
@@ -72,19 +73,86 @@ async fn handle_connection(sock: Arc<tokio::net::TcpStream>, conn: Arc<Connectio
...
+    let addr = match header.addresses {
+        ppp::v2::Addresses::Unspecified => {
+            debug!("Unspecified PROXY connection");
+            return None;
+        }
+        ppp::v2::Addresses::IPv4(addr) => {
+            SocketAddr::new(addr.source_address.into(), addr.source_port)
+        }
+        ppp::v2::Addresses::IPv6(_) => {
+            debug!("IPv6 PROXY connection");
```
kp-mariappan-ramasamy: Repeated debug statements make me wonder whether we should return Result instead of Option and add trace there.

lightway-server/src/metrics.rs
```diff
@@ -109,6 +112,10 @@ pub(crate) fn connection_accept_failed() {
     METRIC_CONNECTION_ACCEPT.increment(1)
 }
 
+pub(crate) fn connection_accept_proxy_header_failed() {
```
kp-mariappan-ramasamy, line 115: Please add these metrics in docs.

lightway-server/src/io/outside/tcp.rs
```diff
@@ -110,6 +178,15 @@ impl Server for TcpServer {
             };
 
             sock.set_nodelay(true)?;
+
+            if self.proxy_protocol {
+                let Some(real_addr) = handle_proxy_protocol(&mut sock).await else {
```
xv-ian-c: Although we expect the proxy to be local and somewhat trusted if it dies without sending a header then we will block here (in the reads within this function) and this will stop us from accepting other connections until the TCP connection closes or is times out.

So I think I need to push this into the `tokio::spawn` task below.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: Yes, good idea. These were left from before I moved this code into a function.

lightway-server/src/io/outside/tcp.rs
```diff
@@ -35,8 +35,48 @@ impl OutsideIOSendCallback for TcpStream {
...
-#[instrument(level = "trace", skip_all, fields(session = ?conn.session_id()))]
-async fn handle_connection(sock: Arc<tokio::net::TcpStream>, conn: Arc<Connection>) {
+#[instrument(level = "trace", skip_all)]
+async fn handle_connection(
+    sock: tokio::net::TcpStream,
+    addr: SocketAddr,
+    conn_manager: Arc<ConnectionManager>,
+) {
+    if let Err(err) = sock.set_nodelay(true) {
+        debug!(?err, "Failed to set NODELAY");
```
kp-mariappan-ramasamy: I think as a side effect of this change, we lose some of these errors.

Previously any of these errors in this path, will shutdown the server. (not saying it's the right behavior, but it was the existing behavior)

But now, we had just debug log whcih we will not enable. 
Might be worth to add metrics in these error cases.

lightway-server/src/io/outside/tcp.rs
```diff
@@ -35,17 +36,82 @@ impl OutsideIOSendCallback for TcpStream {
...
+    };
+
+    let addr = match header.addresses {
+        ppp::v2::Addresses::Unspecified => {
+            return Err(anyhow!("Unspecified PROXY connection"));
+        }
+        ppp::v2::Addresses::IPv4(addr) => {
+            SocketAddr::new(addr.source_address.into(), addr.source_port)
+        }
+        ppp::v2::Addresses::IPv6(_) => {
```
kp-mariappan-ramasamy, line 82: Do we need to handle each of these cases?
Would match `other => return Err(anyhow!(other).context("Unexpected ...));` not suffice ?

lightway-server/src/io/outside/tcp.rs
xv-ian-c: Actually, this one could remain in the caller.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: I wanted a distinct log message (emitted in the caller now) for each, but yes I could combine into one, but the variant content is not an error it's a whole data structure in each case.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: So could the local addr stuff. I did so.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: I think I'll leave this as is.

kp-mariappan-ramasamy: I do not see the point in having this distinction between different cases. Its invalid in all cases.
And we do not enable logging in servers. Without individual metrics, this will be a moot

lightway-server/src/io/outside/tcp.rs
kp-mariappan-ramasamy: I think your existing change was perfect, before reverting.
We should not stop the server, since we could not set nodelay or get sock address for individual connection.

It was a good change and it only needed metrics

lightway-server/src/io/outside/tcp.rs
xv-ian-c: Some users might enable logging, I think it is useful for them.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: IMHO failure to set nodelay or get a sock address indicates that something is _very_ wrong with the system, i.e. invariants (that the socket is the sort of socket we created) have been violated.

They could reasonably even be made asserts but I think failing with an error is better in this case.

PR #107, opened 2024-10-14
CVPN-1449: Handle IOuring Sq push api failures

kp-mariappan-ramasamy:
## Description

There are cases where the sq.push() returns PushError instead of succeeding.
It is not a fatal error, and can be ignored safely.

When we encounter these errors in push_rx_events_to and push_tx_events_to, we add the
slot back to the available slots and restart to submit SQ.

## Motivation and Context

While running speedtest, the iouring task was exiting inconsistently.
From the logs, it is found that sq push returns error and iouring thread exited.

## How Has This Been Tested?

Pending...

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/iouring.rs
+++ b/lightway-app-utils/src/iouring.rs
@@ -298,32 +298,35 @@ async fn iouring_task(
         state.sender = Some(rx_queue.clone().reserve_owned().await?)
     }
 
+    let rx_sq_entries: Vec<_> = rx_slots
+        .drain(..)
+        .map(|slot| {
+            let state = &mut rx_state[slot.idx()];
+            opcode::Read::new(
+                Fixed(REGISTERED_FD_INDEX),
+                state.buf.as_mut_ptr(),
+                state.buf.capacity() as _,
+            )
+            .build()
+            .user_data(slot.user_data())
+        })
+        .collect();
+
+    // SAFETY: sqe points to a buffer on the heap, owned
+    // by a `BytesMut` in `rx_bufs[slot]`, we will not reuse
+    // `rx_bufs[slot]` until `slot` is returned to the slots vector.
+    #[allow(unsafe_code)]
+    unsafe {
+        let entries = rx_sq_entries;
+        // This call should not fail since the SubmissionQueue should be empty now
+        sq.push_multiple(&entries)?
+    };
+    sq.sync();
+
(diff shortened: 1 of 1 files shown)


lightway-app-utils/src/iouring.rs
```diff
@@ -313,6 +317,7 @@ async fn iouring_task(
         // `rx_bufs[slot]` until `slot` is returned to the slots vector.
         #[allow(unsafe_code)]
         unsafe {
+            // This call should not fail since the SubmissionQueue should be empty now
```
xv-ian-c: It's not quite true that the ring is empty "now".

I think what you mean to say is that the ring was empty when we started and we pushed `nr_tx_rx_slots` << `ring_size` entries so there is no way we can have filled the ring.

lightway-app-utils/src/iouring.rs
```diff
@@ -355,19 +355,7 @@ async fn iouring_task(
...
-                        &sbmt,
-                        &mut sq,
-                        &mut tx_queue,
-                        &mut tx_slots,
-                        &mut tx_bufs,
-                    )?;
-
-                    sq.sync();
-
-                    continue 'io_loop;
```
xv-ian-c, line 363: The logic behind this was that if we were woken because of `tx_queue.recv` and not `event_fd.read`  then we would expect the completion queue to be empty (else the event fd would have woken us, so there is no point in processing it.

I suppose there is a chance both might fire and we would happen to see the tx event one, so maybe it is worth consuming whatever is there on the off chance it contains useful stuff.

So I'm happy with the change but it's not quite the "no semantic change" which the commit message implies.

lightway-app-utils/src/iouring.rs
kp-mariappan-ramasamy: Yes correct. I will update

lightway-app-utils/src/iouring.rs
kp-mariappan-ramasamy: I was referring to the case, where at the start of the loop, if the cq is empty() and tx_queue is not empty()

https://github.com/expressvpn/lightway/blob/b6818c9c36ed1c2d1cd8aa00320d349d3f49b733/lightway-app-utils/src/iouring.rs#L334

In this case, we know that cq is empty and the if check will break on tx_queue empty.
Even now, we will process the CQ at the end after push all tx and rx events. We do not again check whether cq is empty before looping.

This change will make both the flow treat the same loop.

PR #113, opened 2024-10-25
CVPN-1483 Add SNI header setter in client connection builder

kp-thomas-yau:
## Description

To allow the client to set the SNI header during the lightway connection, we would need to add a SNI header setter when we are building client connection and configuring the session config of woflssl

## Motivation and Context

- To allow the android app to set SNI header while starting lightway connection
- C client already supports this feature but this isn't in Rust yet.

Issue: https://polymoon.atlassian.net/browse/CVPN-1483

## How Has This Been Tested?

Ran unit test

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection/builders.rs
+++ b/lightway-core/src/connection/builders.rs
@@ -180,6 +180,14 @@ impl<AppState: Send + 'static> ClientConnectionBuilder<AppState> {
         }
     }
 
+    /// Sets SNI header of the session
+    pub fn with_sni_header(self, server_hostname: &str) -> Self {
+        Self {
+            session_config: self.session_config.with_sni(server_hostname),
+            ..self
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


lightway-core/src/connection/builders.rs
```diff
@@ -180,6 +180,14 @@ impl<AppState: Send + 'static> ClientConnectionBuilder<AppState> {
         }
     }
 
+    /// Sets SNI header of the session
+    pub fn with_sni_header(self, server_hostname: String) -> Self {
```
xv-ian-c: Please use `server_hostname: &str`.

kp-thomas-yau: Fixed. Still getting used to ownership rules haha

xv-ian-c: Changes look good but please fold the commits.

PR #116, opened 2024-10-30
CVPN-1506 client first packet event

kp-max-li:
## Description
For each client connection, emits a first packet event when the client got the first outside packet.

## Motivation and Context
Indicate that the first packet is received from the server.

## How Has This Been Tested?
Put some `assert!` in the existing integration tests in `lightway-core` to make sure that the event is only emitted once.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -211,6 +211,9 @@ async fn handle_events<A: 'static + Send + EventCallback>(
                 }
             }
             Event::KeepaliveReply => keepalive.reply_received().await,
+            Event::FirstPacketReceived => {
+                info!("First outside packet received");
+            }
 
             // Server only events
             Event::SessionIdRotationAcknowledged { .. }
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -341,6 +341,9 @@ pub struct Connection<AppState: Send = ()> {
 
     /// Counter to use for `wire::DataFrag`
     fragment_counter: std::num::Wrapping<u16>,
+
+    // Is the first outside packet received
+    is_first_packet_received: bool,
 }
 
 /// Information about the new session being established with a new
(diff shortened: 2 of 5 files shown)


lightway-client/src/lib.rs
```diff
@@ -218,6 +218,10 @@ async fn handle_events<A: 'static + Send + EventCallback>(
             | Event::TlsKeysUpdateCompleted => {
                 unreachable!("server only event received");
             }
+
+            Event::FirstPacketReceived => {
+                // Do nothing
```
xv-thomas-leong: Can you add an `info!` message indicating the event is received?

lightway-core/src/connection.rs
```diff
@@ -678,6 +682,11 @@ impl<AppState: Send> Connection<AppState> {
     /// In case of UDP, it is almost always one frame per packet. With duplicated
     /// UDP packets, count can be 0.
     pub fn outside_data_received(&mut self, pkt: OutsidePacket) -> ConnectionResult<usize> {
+        if matches!(self.mode, ConnectionMode::Client { .. }) && !self.is_first_packet_received {
```
xv-thomas-leong: Please put the `is_first_packet_received` boolean check in the front

As Rust performs short circuit on boolean operators, by putting the easier check in the front should speed up execution. I'm assuming the boolean check is less expensive than the `matches!` marco.

Ref: https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators

kp-max-li: Good catch. `!self.is_first_packet_received` will be false the majority of the time for the client, so the change will give less overhead.
I will change it. Thanks.

lightway-client/src/lib.rs
kp-max-li: Sure

PR #117, opened 2024-10-30
CVPN-1514 Add Server UDP Buffer Size Config

kp-thomas-yau:
## Description

Reads UDP Buffer Size in config and set buffer size in socket level

## Motivation and Context

- Allow devs to tweak udp buffer size 
- Prevents hardcoding 

## How Has This Been Tested?

Ran cargo test locally + tried running lightway_server locally and print out the buffer size

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/args.rs
+++ b/lightway-server/src/args.rs
@@ -3,6 +3,7 @@ use std::{
     path::PathBuf,
 };
 
+use bytesize::ByteSize;
 use clap::Parser;
 use ipnet::Ipv4Net;
 use twelf::config;
@@ -99,6 +100,10 @@ pub struct Config {
     /// Enable PROXY protocol support (TCP only)
     #[clap(long)]
     pub proxy_protocol: bool,
+
+    /// Set UDP buffer size. Default value is 15 MiB.
+    #[clap(long, default_value_t = ByteSize::mib(15))]
+    pub udp_buffer_size: ByteSize,
 
     /// Enable WolfSSL debug logging
     #[cfg(feature = "debug")]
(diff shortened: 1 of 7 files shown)


lightway-server/src/args.rs
```diff
@@ -100,6 +100,10 @@ pub struct Config {
     #[clap(long)]
     pub proxy_protocol: bool,
 
+    /// Set UDP buffer size. Default value is 15 MB.
+    #[clap(long, default_value_t = 15 * 1024 * 1024)]
+    pub udp_buffer_size: usize,
```
xv-thomas-leong: Can you use the ByteSize Type like the client does? It will make it easier for users to type a size in natural language

https://github.com/expressvpn/lightway/blob/aa958c446a479b9dbce048fa1f3052e5b01e5ea8/lightway-client/src/args.rs#L82

tests/server/server_config.yaml
```diff
@@ -21,3 +21,4 @@ enable_tun_iouring: false
 iouring_entry_count: 1024
 key_update_interval: 15m
 user_db: "tests/server/lwpasswd"
+udp_buffer_size: 20971520
```
xv-thomas-leong: Values we put in this file should constitute the value we recommend for general users. Are you suggesting 20 MB should be used by default?

lightway-server/src/args.rs
kp-thomas-yau: oooo never seen this before~ lemme see see how it works

tests/server/server_config.yaml
kp-thomas-yau: i was just trying out different values and forgot to reset it back to 15 lol

PR #119, opened 2024-11-01
CVPN-1356 Check Session ID after connection is validated

kp-thomas-yau:
## Description
The new approach would first validate the outside packet (both content validation and replay attack verification) and would only read or update the session_id once the packet is validated.
1. Prevents client from accepting session_id from unknown sources
2. Allows server to process valid packets even with session_id mismatch, since it's only a backup lookup method after IP/Port search fails. A mismatch here won't affect anything and the packet has been validated already anyway.

(+ Minor refactoring)

## Motivation and Context
Previously, we did not validate packet content before processing session_id, causing:
- clients to accept session_id from unknown sources
- servers to reject packets on session_id mismatch

These are not ideal and would easily affect the VPN performance or security

## How Has This Been Tested?
- cargo test

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -1017,7 +1017,7 @@ impl<AppState: Send> Connection<AppState> {
                         self.event(Event::SessionIdRotationAcknowledged { old, new });
                     }
                     // Session id in server is only used to look up the session if it was not found by client IP/Port, so a mismatch here won't affect anything.
-                    _ => {}
+                    _ => metrics::session_id_mismatch(),
                 }
             }
         }
--- a/docs/logs_and_metrics.md
+++ b/docs/logs_and_metrics.md
@@ -31,6 +31,7 @@ Lightway server also supports metrics to monitor. The following are the metrics 
 | conn_create_failed | server | Counter | A new connection could not be created |
 | conn_alloc_frag_map | core | Counter | A connection has used a fragmented data packet.<br>Therefore the 2M FragmentMap has been allocated |
 | wolfssl_appdata | core | Counter | An AppData result occurred during a WolfSSL operation.<br><br>Given current configuration we do not expect this to be non-zero |
+| session_id_mismatch | core | Counter | Server has received a mismatched session_id in the header after the packet content has been validated <br><br>Should generally be expected to happen rarely|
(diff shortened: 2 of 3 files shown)


lightway-core/src/connection.rs
```diff
@@ -712,7 +712,12 @@ impl<AppState: Send> Connection<AppState> {
             return Err(ConnectionError::RejectedSessionID);
         };
 
-        self.process_new_outside_data(payload, session_id)
+        let result = self.process_new_outside_data(payload);
+        if result.is_ok() {
```
kp-mariappan-ramasamy: We should update the session id, only when the number of frames read is greater than 0 (similar to server: https://github.com/expressvpn/lightway/blob/1ad947b7ca93f922fba9ed8624f50752ec332c05/lightway-server/src/io/outside/udp.rs#L216)

In case of UDP, if there is a replay attack with different session id, the number of frames read will be 0, since wolfssl will ignore the previously read packets.
We need to accept session id only when wolfssl read one message successfully

lightway-core/src/connection.rs
```diff
@@ -1007,13 +1015,9 @@ impl<AppState: Send> Connection<AppState> {
...
                         self.event(Event::SessionIdRotationAcknowledged { old, new });
-
-                        Ok(())
                     }
-                    // Server session does not match current session (due to
-                    // check before the `match` above) nor the pending one
-                    // (due to the guard on the previous case in this match).
-                    Some(_) | None => Err(ConnectionError::UnknownSessionID),
+                    // Session id in server is only used to look up the session if it was not found by client IP/Port, so a mismatch here won't affect anything.
+                    _ => {}
```
kp-mariappan-ramasamy: Can you please add a metric here ?
Even though, we are going to accept this packet, we do not expect this to happen.
Adding a metric will help us later to identify how often we hit this.

kp-thomas-yau: I don't think we need to save both session_id from the client and server's end right? I think saving them won't help much?

kp-mariappan-ramasamy: In general, server is the one generating session id and sending to client.
Client usually accepts the session_id. 

If we try to store session_id from client in server, might create a circular deps. Better to leave it with a metric

PR #123, opened 2024-11-11
Switch lightway-server to a direct uring only i/o loop

xv-ian-c:
## Description

This does not consistently improve performance but reduces CPU overheads (by
around 50%-100% i.e. half to one core) under heavy traffic, which adding
perhaps a few hundred Mbps to a speedtest.net download test and making
negligible difference to the upload test. It also removes about 1ms from the
latency in the same tests. Finally the STDEV across multiple test runs appears
to be lower.

This appears to be due to a combination of avoiding async runtime overheads, as
well as removing various channels/queues in favour of a more direct model of
interaction between the ring and the connections.

As well as those benefits we are now able to reach the same level of
performance with far fewer slots used for the TUN rx path, here we use 64 slots
(by default) and reach the same performance as using 1024 previously. The way
uring handles blocking vs async for tun devices seems to be non-optimal. In
blocking mode things are very slow. In async mode more and more time is spent
on bookkeeping and polling, as the number of slots is increased, plus a high
level of EAGAIN results (due to a request timing out after multiple failed
polls[^0]) which waste time requeueing. This is related to
https://github.com/axboe/liburing/issues/886 and https://github.com/axboe/liburing/issues/239.

For UDP/TCP sockets io uring behaves well with the socket in blocking mode
which avoids processing lots of EAGAIN results.

Tuning the slots for each I/O path is a bit of an art (more is definitely not
always better) and the sweet spot varies depending on the I/O device, so
provide various tunables instead of just splitting the ring evenly. With this
there's no real reason to have a very large ring, it's the number of inflight
requests which matters.

This is specific to the server since it relies on kernel features and
correctness(/lack of bugs) which may not be upheld on an arbitrary client
system (while it is assumed that server operators have more control over what
they run). It is also not portable to non-Linux systems. It is known to work
with Linux 6.1 (as found in Debian 12 AKA bookworm).

Note that this kernel version contains a bug which causes the `iou-sqp-*`
kernel thread to get stuck (unkillable) if the tun is in blocking mode,
therefore an option is provided. Enabling that option on a kernel which
contains [the fix][] allows equivalent performance with fewer slots on the
ring.

[^0]: When data becomes available _all_ requests are woken but only one will
      find data, the rest will see EAGAIN and after a certain number of such
      events I/O uring will propagate this back to userspace.
[the fix]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=438b406055cd21105aad77db7938ee4720b09bee

## Motivation and Context

Attempting to improve performance and CPU usage (bang per buck)

## How Has This Been Tested?

I tested with a pair of EC2 c7i.8xlarge nodes, one acting as a lightway server and the other as a client. The nodes have a fast interconnect (10Gbps)

iperf3 tests were run from the client to the server and speedtest from the client to a speedtest node via the server.

| connection| iperf download | iperf upload | speedtest download | speedtest upload | CPU use download[^0] | CPU use upload |
|--- | --- | --- | --- | --- | --- | --- |
| UDP | +16% | -25% | +22% | -0.75% | ~200% -> ~140% | ~270% -> 140% |
| TCP | +33% | -13% | +32% | -1% | ~200% -> ~140% | ~240% -> 140% | 

The iperf upload results are worse, however this does not appear to be reflected in the speedtest upload results which should better reflect real world outcomes.

[^0]: These CPU figures includes the iouring kernel thread which consumes 100% in all cases, the remainder is the userspace effort to keep the ring fed. In the before case 3 threads are active (kernel thread + uring thread + a tokio thread) while in the after case only 2 threads are active (kernel thread + the userspace io thread).

## Types of changes

- [x] Performance enhancement

## Checklist:

---

This applies some of the techniques from #88.

I'm undecided if/how to apply this to the client, since the I/O loop is so different it's not really possible to have an implementation which flips from uring to using the Tokio abstractions other than by having two different i/o loops and an if statement.

### DIFF
--- a/lightway-app-utils/src/net.rs
+++ b/lightway-app-utils/src/net.rs
@@ -1,0 +1,179 @@
+use std::{io, net::SocketAddr};
+
+/// Convert from `libc::sockaddr_storage` to `std::net::SocketAddr`
+#[allow(unsafe_code)]
+pub fn socket_addr_from_sockaddr(
+    storage: &libc::sockaddr_storage,
+    len: libc::socklen_t,
+) -> io::Result<SocketAddr> {
+    match storage.ss_family as libc::c_int {
+        libc::AF_INET => {
+            if (len as usize) < std::mem::size_of::<libc::sockaddr_in>() {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "invalid argument (inet len)",
+                ));
+            }
+
+            // SAFETY: Casting from sockaddr_storage to sockaddr_in is safe since we have validated the len.
+            let addr =
+                unsafe { &*(storage as *const libc::sockaddr_storage as *const libc::sockaddr_in) };
+
+            let ip = u32::from_be(addr.sin_addr.s_addr);
+            let ip = std::net::Ipv4Addr::from_bits(ip);
+            let port = u16::from_be(addr.sin_port);
+
+            Ok((ip, port).into())
+        }
+        libc::AF_INET6 => {
+            if (len as usize) < std::mem::size_of::<libc::sockaddr_in6>() {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    "invalid argument (inet6 len)",
+                ));
+            }
+            // SAFETY: Casting from sockaddr_storage to sockaddr_in6 is safe since we have validated the len.
+            let addr = unsafe {
+                &*(storage as *const libc::sockaddr_storage as *const libc::sockaddr_in6)
+            };
+
+            let ip = u128::from_be_bytes(addr.sin6_addr.s6_addr);
+            let ip = std::net::Ipv6Addr::from_bits(ip);
+            let port = u16::from_be(addr.sin6_port);
+
+            Ok((ip, port).into())
+        }
+        _ => Err(io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            "invalid argument (ss_family)",
+        )),
+    }
+}
+
+/// Convert from `std::net::SocketAddr` to `libc::sockaddr_storage`+`libc::socklen_t`
+#[allow(unsafe_code)]
+pub fn sockaddr_from_socket_addr(addr: SocketAddr) -> (libc::sockaddr_storage, libc::socklen_t) {
+    // SAFETY: All zeroes is a valid sockaddr_storage
+    let mut storage: libc::sockaddr_storage = unsafe { std::mem::zeroed() };
+
+    let len = match addr {
+        SocketAddr::V4(v4) => {
+            let p = &mut storage as *mut libc::sockaddr_storage as *mut libc::sockaddr_in;
+            // SAFETY: sockaddr_storage is defined to be big enough for any sockaddr_*.
+            unsafe {
+                p.write(libc::sockaddr_in {
+                    sin_family: libc::AF_INET as _,
+                    sin_port: v4.port().to_be(),
+                    sin_addr: libc::in_addr {
+                        s_addr: v4.ip().to_bits().to_be(),
+                    },
+                    sin_zero: Default::default(),
+                })
+            };
+            std::mem::size_of::<libc::sockaddr_in>() as libc::socklen_t
+        }
+        SocketAddr::V6(v6) => {
+            let p = &mut storage as *mut libc::sockaddr_storage as *mut libc::sockaddr_in6;
+            // SAFETY: sockaddr_storage is defined to be big enough for any sockaddr_*.
+            unsafe {
+                p.write(libc::sockaddr_in6 {
+                    sin6_family: libc::AF_INET6 as _,
+                    sin6_port: v6.port().to_be(),
+                    sin6_flowinfo: v6.flowinfo().to_be(),
+                    sin6_addr: libc::in6_addr {
+                        s6_addr: v6.ip().to_bits().to_be_bytes(),
+                    },
+                    sin6_scope_id: v6.scope_id().to_be(),
+                })
+            };
+            std::mem::size_of::<libc::sockaddr_in6>() as libc::socklen_t
+        }
+    };
+
+    (storage, len)
+}
+
+#[cfg(test)]
+mod tests {
+    #![allow(unsafe_code, clippy::undocumented_unsafe_blocks)]
+
+    use std::{
+        net::{IpAddr, Ipv4Addr, Ipv6Addr},
+        str::FromStr as _,
+    };
+
+    use super::*;
+
+    use test_case::test_case;
+
+    #[test]
+    fn socket_addr_from_sockaddr_unknown_af() {
+        // Test assumes these don't match the zero initialized
+        // libc::sockaddr_storage::ss_family.
+        assert_ne!(libc::AF_INET, 0);
+        assert_ne!(libc::AF_INET6, 0);
+
+        let storage = unsafe { std::mem::zeroed() };
+        let err =
+            socket_addr_from_sockaddr(&storage, std::mem::size_of::<libc::sockaddr_storage>() as _)
+                .unwrap_err();
+
+        assert!(matches!(err.kind(), std::io::ErrorKind::InvalidInput));
+        assert!(err.to_string().contains("invalid argument (ss_family)"));
+    }
+
+    #[test]
+    fn socket_addr_from_sockaddr_unknown_af_inet_short() {
+        let mut storage: libc::sockaddr_storage = unsafe { std::mem::zeroed() };
+        storage.ss_family = libc::AF_INET as libc::sa_family_t;
+
+        let err = socket_addr_from_sockaddr(
+            &storage,
+            (std::mem::size_of::<libc::sockaddr_in>() - 1) as _,
+        )
+        .unwrap_err();
+
+        assert!(matches!(err.kind(), std::io::ErrorKind::InvalidInput));
+        assert!(err.to_string().contains("invalid argument (inet len)"));
+    }
+
+    #[test]
+    fn socket_addr_from_sockaddr_unknown_af_inet6_short() {
+        let mut storage: libc::sockaddr_storage = unsafe { std::mem::zeroed() };
+        storage.ss_family = libc::AF_INET6 as libc::sa_family_t;
+
+        let err = socket_addr_from_sockaddr(
+            &storage,
+            (std::mem::size_of::<libc::sockaddr_in6>() - 1) as _,
+        )
+        .unwrap_err();
+
+        assert!(matches!(err.kind(), std::io::ErrorKind::InvalidInput));
+        assert!(err.to_string().contains("invalid argument (inet6 len)"));
+    }
+
+    #[test]
+    fn sockaddr_from_socket_addr_inet() {
+        let socket_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8080);
+        let (storage, len) = sockaddr_from_socket_addr(socket_addr);
+        assert_eq!(storage.ss_family, libc::AF_INET as libc::sa_family_t);
+        assert_eq!(len as usize, std::mem::size_of::<libc::sockaddr_in>());
+    }
+
+    #[test]
+    fn sockaddr_from_socket_addr_inet6() {
+        let socket_addr = SocketAddr::new(IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1)), 8080);
+        let (storage, len) = sockaddr_from_socket_addr(socket_addr);
+        assert_eq!(storage.ss_family, libc::AF_INET6 as libc::sa_family_t);
+        assert_eq!(len as usize, std::mem::size_of::<libc::sockaddr_in6>());
+    }
+
+    #[test_case("127.0.0.1:443")]
+    #[test_case("[::1]:8888")]
+    fn round_trip(addr: &str) {
+        let orig = SocketAddr::from_str(addr).unwrap();
+        let (storage, len) = sockaddr_from_socket_addr(orig);
+        let round_tripped = socket_addr_from_sockaddr(&storage, len).unwrap();
+        assert_eq!(orig, round_tripped)
+    }
+}
--- a/lightway-server/Cargo.toml
+++ b/lightway-server/Cargo.toml
@@ -9,9 +9,8 @@ license = "AGPL-3.0-only"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["io-uring"]
+default = []
 debug = ["lightway-core/debug"]
-io-uring = ["lightway-app-utils/io-uring"]
 
 [lints]
 workspace = true
@@ -26,6 +25,7 @@ clap.workspace = true
 ctrlc.workspace = true
 delegate.workspace = true
 educe.workspace = true
+io-uring.workspace = true
 ipnet.workspace = true
 jsonwebtoken = "9.3.0"
 libc.workspace = true
@@ -48,6 +48,7 @@ tokio-stream = { workspace = true, features = ["time"] }
 tracing.workspace = true
 tracing-log = "0.2.0"
 tracing-subscriber = { workspace = true, features = ["json"] }
+tun.workspace = true
 twelf.workspace = true
 
 [dev-dependencies]
--- a/lightway-server/src/args.rs
+++ b/lightway-server/src/args.rs
@@ -71,13 +71,25 @@ pub struct Config {
     #[clap(long, default_value_t)]
     pub enable_pqc: bool,
 
-    /// Enable IO-uring interface for Tunnel
-    #[clap(long, default_value_t)]
-    pub enable_tun_iouring: bool,
-
-    /// IO-uring submission queue count. Only applicable when
-    /// `enable_tun_iouring` is `true`
-    // Any value more than 1024 negatively impact the throughput
+    /// Total IO-uring submission queue count.
+    ///
+    /// Must be larger than the total of:
+    ///
+    /// UDP:
+    ///
+    ///   iouring_tun_rx_count + iouring_udp_rx_count +
+    ///   iouring_tx_count + 1 (cancellation request)
+    ///
+    /// TCP:
+    ///
+    ///   iouring_tun_rx_count + iouring_tx_count + 1 (cancellation
+    ///   request) + 2 * maximum number of connections.
+    ///
+    ///   Each connection actually uses up to 3 slots, a persistent
+    ///   recv request and on demand slots for TX and cancellation
+    ///   (teardown).
+    ///
+    /// There is no downside to setting this much larger.
     #[clap(long, default_value_t = 1024)]
     pub iouring_entry_count: usize,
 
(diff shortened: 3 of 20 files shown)


lightway-server/Cargo.toml
```diff
@@ -26,6 +25,7 @@ clap.workspace = true
 ctrlc.workspace = true
 delegate.workspace = true
 educe.workspace = true
+io-uring = "0.7.0"
```
kp-mariappan-ramasamy: io-uring is also a dep of app-utils:
https://github.com/expressvpn/lightway/pull/123/commits/43fd992f20499b285b2bc70b6cd7394365ec7569#diff-727d8cfa56ae937a58e105ece9ff656b41ef39ad3f172d4c5845ecc2f1ffd886R26

Better to move it to a workspace dep

lightway-server/Cargo.toml
```diff
@@ -9,9 +9,8 @@ license = "GPL-2.0-only"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["io-uring"]
+default = []
```
kp-mariappan-ramasamy, line 12: With this change, we forbid lightway-server to run on any other platforms, without any major refactor.

I think it's worth mentioning explicitly in README about this:
https://github.com/expressvpn/lightway?tab=readme-ov-file#supported-platforms

lightway-server/Cargo.toml
xv-ian-c: Agreed, I swear I had done but it must have been in some experimental branch which I disgarded.

lightway-server/Cargo.toml
xv-ian-c: Ack

lightway-app-utils/src/net.rs
```diff
@@ -0,0 +1,94 @@
...
+        _ => Err(io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            "invalid argument (ss_family)",
+        )),
+    }
+}
+
+/// Convert from `std::net::SocketAddr` to `libc::sockaddr_storage`+`libc::socklen_t`
+#[allow(unsafe_code)]
+pub fn sockaddr_from_socket_addr(addr: SocketAddr) -> (libc::sockaddr_storage, libc::socklen_t) {
```
kp-mariappan-ramasamy, line 55: We can consider adding tests for these util functions

lightway-server/src/args.rs
```diff
@@ -71,13 +71,25 @@ pub struct Config {
...
+    ///   iouring_tun_rx_count + iouring_tx_count + 1 (cancellation
+    ///   request) + 2 * maximum number of connections.
+    ///
+    ///   Each connection actually uses up to 3 slots, a persistent
+    ///   recv request and on demand slots for TX and cancellation
+    ///   (teardown).
+    ///
+    /// There is no downside to setting this much larger.
     #[clap(long, default_value_t = 1024)]
     pub iouring_entry_count: usize,
```
kp-mariappan-ramasamy, line 94: I am not sure whether I missed the change.
But i think we can add this check inside code and throw error if not, instead of warning in command line

lightway-server/src/lib.rs
```diff
@@ -201,67 +211,42 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
...
-                        break Err(anyhow!(err).context("Inside data handling error"));
-                    }
-                }
-            } else {
-                metrics::tun_rejected_packet_no_connection();
-            }
-        }
+    // On exit dropping _io_handle will cause EPIPE to be delivered to
+    // io_cancel. This causes the corresponding read request on the
+    // ring to complete and signal the loop should exit.
```
kp-mariappan-ramasamy, line 268: Is this guaranteed ? AFAIK this current flow will be running in main_thread. And as soon as it exits, the process will exit.

Or does tokio runtime join the worker threads and make it work since we do not use thread directly and use task_blocking ?

lightway-server/src/io.rs
```diff
@@ -1,2 +1,280 @@
...
+        unsafe {
+            sq.push(&sqe)?
+        };
+
+        self.outside.push_initial_ops(&mut sq)?;
+        self.inside.push_initial_ops(&mut sq)?;
+        sq.sync();
+
+        loop {
+            let _ = submitter.submit_and_wait(1)?;
```
kp-mariappan-ramasamy, line 221: What is the significance of this waiting ? What happens if we dont and continuously loop ?
Are we wasting CPU cycles

lightway-server/src/io.rs
```diff
@@ -1,2 +1,280 @@
...
+    fn outside_rx_user_data(idx: u32) -> u64 {
+        Self::OUTSIDE_RX_USER_DATA_BASE + (idx as u64)
+    }
+
+    /// Return user data for a particular inside rx index.
+    fn inside_rx_user_data(idx: u32) -> u64 {
+        Self::INSIDE_RX_USER_DATA_BASE + (idx as u64)
+    }
+
+    /// Return user data for a particular outside tx index.
```
kp-mariappan-ramasamy:
```suggestion
    /// Return user data for a particular inside tx index.
```

lightway-server/src/io.rs
```diff
@@ -1,2 +1,280 @@
...
+    fn inside_tx_user_data(idx: u32) -> u64 {
+        Self::INSIDE_TX_USER_DATA_BASE + (idx as u64)
+    }
+
+    /// Return user data for a particular inside tx index.
```
kp-mariappan-ramasamy:
```suggestion
    /// Return user data for a particular outside tx index.
```

lightway-server/src/io.rs
```diff
@@ -1,2 +1,280 @@
...
+    /// Use for inside IO requests, `self.inside.as_raw_fd` will be registered in this slot.
+    const FIXED_INSIDE_FD: Fixed = Fixed(1);
+
+    /// Masks the bits used by `*_USER_DATA_BASE`
+    const USER_DATA_TYPE_MASK: u64 = 0xe000_0000_0000_0000;
+
+    /// Indexes in this range will result in a call to `self.outside.complete_rx`
+    const OUTSIDE_RX_USER_DATA_BASE: u64 = 0xc000_0000_0000_0000;
+    /// Indexes in this range will result in a call to `self.outside.complete_tx`
+    const OUTSIDE_TX_USER_DATA_BASE: u64 = 0x8000_0000_0000_0000;
```
kp-mariappan-ramasamy, line 130: These overlapping bits looks dangerous.

With this overlap, the pattern matching will fail, if we compare OUTSIDE_TX_USER_DATA_BASE first and  OUTSIDE_RX_USER_DATA_BASE.

Ex: 0x8000_0000_0000_0001 value will match both OUTSIDE_RX_USER_DATA_BASE and OUTSIDE_TX_USER_DATA_BASE

lightway-server/src/io/inside/tun.rs
```diff
@@ -1,68 +1,251 @@
...
         };
 
         ipv4_update_source(buf.as_mut(), client_ip);
-        metrics::tun_from_client(buf.len());
-        self.0.try_send(buf)
+        metrics::tun_from_client(len);
+
+        let buf = buf.freeze();
+
+        let mut ring = self.0.lock().unwrap();
```
kp-mariappan-ramasamy: `ring` sounded like iouring struct. txqueue might have a better name.

lightway-server/src/io/outside/tcp.rs
```diff
@@ -1,230 +1,570 @@
...
+}
 
-        let pkt = OutsidePacket::Wire(&mut buf, ConnectionType::Stream);
-        if let Err(err) = conn.outside_data_received(pkt) {
-            warn!("Failed to process outside data: {err}");
-            if conn.handle_outside_data_error(&err).is_break() {
-                break anyhow!(err).context("Outside data fatal error");
-            }
+pub(in super::super) struct TcpSocketBuffer {
+    ring: Arc<Mutex<TxQueue>>,
```
kp-mariappan-ramasamy: Naming is super confusing in this place and difficult to follow.

We have a variable `ring` which is a `TxQueue` and another variable `tx_queue` which is a bytesMut.
Suggest to rename `ring` to tx_queue and `tx_queue` to `tx_buffer`

lightway-server/src/io/tx.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    #[allow(unsafe_code)]
+    /// Push an arbitrary entry to the tx queue. Does not consume a slot.
+    ///
+    /// Callers are responsible for completion.
+    ///
+    /// # Safety:
+    ///
+    /// - sqe must meet the safety requirements <https://docs.rs/io-uring/latest/io_uring/squeue/struct.SubmissionQueue.html#method.push>
+    pub(super) unsafe fn push(&mut self, sqe: SEntry) {
```
kp-mariappan-ramasamy, line 127: This function is some kind of bypass function of above two.
Also its not clear when do we use these functions. Can you please update comment on when do we use this ?

lightway-server/src/io/outside/tcp.rs
```diff
@@ -1,230 +1,570 @@
...
+    }
+
+    pub fn complete_tx(
+        &mut self,
+        _sq: &mut io_uring::SubmissionQueue,
+        cqe: io_uring::cqueue::Entry,
+    ) -> Result<()> {
+        let res = io_uring_res(cqe.result()).with_context(|| "outside send completion")? as usize;
+
+        // We use MSG_WAITALL so this should not happen
```
kp-mariappan-ramasamy, line 309: Sorry, my google fu failed.

I could not find the reference to MSG_WAITALL flag for send operation. 
I can see it is available for recv and friends:
https://man7.org/linux/man-pages/man2/recv.2.html

Can you please post a link where it is referenced for send ?

lightway-server/src/args.rs
xv-ian-c: I think you are referring to the `/// Must be larger than the total of:` bit which ended up outside of what GH shows on the comment. In which case I agree, I'll add something.

lightway-server/src/lib.rs
xv-ian-c: Before I added this I observed that even after I killed the main thread the process kept going because the uring userspace thread was still going (actually IIRC it was blocked in the submit syscall).

I don't know if it is Tokio cleaning up (as part of the `#[tokio::main]` infra) or if this is a Linux/POSIX thing which won't actually destroy a process while a thread still exists (being blocked in the kernel might be a relevant factor too).

The FD is guaranteed to be closed at the point `lightway_server::server` returns, which happens (shortly) before whatever happens on return from `main` which is enough to kick the thread out of the syscall due to the arrival of EPIPE on the other end of the pipe which lets the process exit.

lightway-server/src/args.rs
kp-mariappan-ramasamy: Yes, I am referring to the requirements for this value. Thanks

lightway-server/src/io.rs
xv-ian-c: This corresponds to a call to [io_uring_submit_and_wait(3)](https://man7.org/linux/man-pages/man3/io_uring_submit_and_wait.3.html).

We need to do at least [`io_uring_submit(3)`](https://man7.org/linux/man-pages/man3/io_uring_submit.3.html) in order to post any new SQEs, but if we just did that we would indeed just spin unless we did our own wait (e.g. with eventfd stuff), but that would just mean more overhead -- you can think of the `_and_wait` as combining that into the submit.

I believe the `_and_wait` will DTRT if there are already available CQEs and not actually block waiting, it's "wait for >= N CQEs to be available" not "wait for N new CQEs to become available".

lightway-server/src/io.rs
xv-ian-c: I don't think it will. The test is `user_data & Self::USER_DATA_TYPE_MASK` vs the specific values here, it's not checking bits.

So `0x8000_0000_0000_0001 & MASK` => `0x8000_0000_0000_0000` which `== OUTSIDE_TX_USER_DATA_BASE` and `!= 0xc000_0000_0000_0000`.

Or did I misunderstand the worry?

lightway-server/src/io/outside/tcp.rs
```diff
@@ -1,230 +1,570 @@
...
+        unsafe {
+            ring.push(sqe)
+        };
+    }
+
+    fn send(&mut self, buf: CowBytes) -> IOCallbackResult<usize> {
+        let bytes = buf.as_bytes();
+
+        if !self.tx_in_flight.is_empty() {
+            self.tx_queue.extend_from_slice(bytes);
```
kp-mariappan-ramasamy: I think we need to bound this `tx_queue` buffer and apply back pressure to caller.
Otherwise in case of slow connection, the application does not know about the network congestion and continuously send data.

lightway-server/src/io/tx.rs
xv-ian-c: The distinction is:
```
    /// Callers are responsible for calling `::complete` when the
    /// request completes to free the slot.
```

vs 

```
    /// Callers are responsible for completion.
```

The user is TCP send which does not need to manage buffers/slots in the TxQueue (it only uses the queuing bit) it uses the TCP socket fd as the index not a slot.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: It's mentioned in https://man7.org/linux/man-pages/man3/io_uring_prep_send.3.html. It appears to be a uring specific extension: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/io_uring/net.c#n644

lightway-server/src/io/tx.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Push all entries (added by `push_*_slot` or `push`) to the uring.
+    pub(super) fn drain(&mut self, submitter: &Submitter, sq: &mut SubmissionQueue) -> Result<()> {
+        for sqe in self.sqe_ring.drain(..) {
+            if sq.is_full() {
+                match submitter.submit() {
+                    Ok(_) => (),
+                    Err(ref err) if err.raw_os_error() == Some(libc::EBUSY) => break,
+                    Err(err) => return Err(err.into()),
```
kp-mariappan-ramasamy, line 138: From doc, 
https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.drain
> Removes the specified range from the deque in bulk, returning all removed elements as an iterator. If the iterator is dropped before being fully consumed, it drops the remaining removed elements.

In these error cases, the remaining sqe entries will be dropped. I think it is not desired here

xv-pete-m: Not meaning to hijack this thread, but in the C version our approach was send as fast as possible, if the underlying layer can't handle it, drop the packet. The idea was that if it was a UDP packet, then no big deal as it's unreliable, and if it's TCP, it'll be resent. We didn't try and manage the flow in the app.

I don't know if this is a pro or a con, but I wonder if this might slow things down?

xv-ian-c: I think that is also relevant to the previous https://github.com/expressvpn/lightway/pull/123#discussion_r1851600419. I think that is basically what @kp-mariappan-ramasamy is proposing there too.

lightway-server/src/io.rs
kp-mariappan-ramasamy: Oh, you are right. I overlooked the pattern matching to be bitwise AND as the value we use to match.
Thanks for clarifying

lightway-server/src/io.rs
kp-mariappan-ramasamy: Yes, I meant to use [io_uring_submit](https://man7.org/linux/man-pages/man3/io_uring_submit.3.html) instead of _wait function, which is in our other iouring implementation.

I can think of a problem with waiting for the first time.
We will only have RX ops and a cancel req pending in the uring. We cannot TX anything unless one RX operation is completed. 

But this may not be a problem for Server, since only clients initiate connection and there will be an RX in server before any lightway connection.

lightway-server/src/args.rs
```diff
@@ -87,6 +99,35 @@ pub struct Config {
...
+    /// IO-uring. Setting this too large may negatively impact
+    /// performance.
+    #[clap(long, default_value_t = 64)]
+    pub iouring_tun_rx_count: u32,
+
+    /// Configure TUN device in blocking mode. This can allow
+    /// equivalent performance with fewer `ìouring-tun-rx-count`
+    /// entries but can significantly harm performance on some kernels
+    /// where the kernel does not indicate that the tun device handles
+    /// `FMODE_NOWAIT`.
```
kp-mariappan-ramasamy, line 112: Is there a kernel version from which the blocking mode works ? If so, might be easier to mention it here for easier reference

lightway-server/src/io.rs
xv-ian-c: If there was an initial tx required then it should be pushed by the relevant `UringIoSource::push_initial_ops` operation and then would be submitted by the first call to `submit_and_wait` and would complete.

But as you say server connections are always initiated by an rx so it's not an issue in practice.

I suppose having a tx queue drain after the calls to `push_initial_ops` and the main loop might be convenient for some implementations of `UringIoSource` (so they don't have to special case initial tx pushes from the rest). We don't have any of those but no harm in adding it.

lightway-server/src/io/inside/tun.rs
```diff
@@ -1,68 +1,251 @@
-use crate::{io::inside::InsideIO, metrics};
+//! Tun UringIoSource
+//!
+//! Uses uring indexes:
+//!
+//! Loop::outside_rx_user_data:
```
kp-mariappan-ramasamy:
```suggestion
//! Loop::inside_rx_user_data:
```

Same for tx in next few lines

lightway-server/src/args.rs
xv-ian-c: The referenced commit was part of v6.4-rc1, and does not appear to have been backported to any stable branches. I'll mention in the comment.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: I added a `tcp_buffer_size` option alongside `udp_buffer_size` and plumbed it through.

I didn't repurpose the existing option because TCP buffer size is per connection and UDP buffer size is for the whole server, so I think the appropriate default is different.

I didn't implement it as a hard limit, we'll overshoot a bit but then stop queuing new stuff.

xv-ian-c: TODO: Retest the perf impact.

lightway-server/src/io/tx.rs
xv-ian-c: I switched this to `while let Some(sqe) = self.sqe_ring.pop_front() {` which was the original intention of this code.

I couldn't decide if all possible errors here would warrant dropping the whole rest of the queue.

lightway-server/src/io/outside/tcp.rs
xv-ian-c: 1M is ample, no performance hit. In fact upload improves by up to 25%! (download is fractions of a percent, basically in the noise)

I tested in steps downwards. 128K is as good as 1M. At 64K and lower perf falls off a cliff.

I went with 256K to give some headroom for performance on different systems.

lightway-server/src/io.rs
xv-ian-c: I decided not to make a change here. The use case of pushing to the tx queue instead of pushing directly is hypothetical for now and it'll be easy enough to change if it ever becomes useful.

PR #125, opened 2024-11-13
lightway-app-utils: iouring: Make sqpoll idle time configurable

xv-ian-c:
## Description

Make it configurable and plumb it all the way through to the config file. The default remains 100ms.

## Motivation and Context

This is a useful tunable to play with when trading CPU use vs performance.

## How Has This Been Tested?

Observing the behaviour using htop and in particular observing that with it set to 0 no kernel thread is created.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/args.rs
+++ b/lightway-server/src/args.rs
@@ -81,6 +81,12 @@ pub struct Config {
     #[clap(long, default_value_t = 1024)]
     pub iouring_entry_count: usize,
 
+    /// IO-uring sqpoll idle time. If non-zero use a kernel thread to
+    /// perform submission queue polling. After the given idle time
+    /// the thread will go to sleep.
+    #[clap(long, default_value = "100ms")]
+    pub iouring_sqpoll_idle_time: Duration,
+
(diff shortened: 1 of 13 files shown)


lightway-server/src/args.rs
```diff
@@ -81,6 +81,12 @@ pub struct Config {
     #[clap(long, default_value_t = 1024)]
     pub iouring_entry_count: usize,
 
+    /// IO-uring sqpoll idle time. If non-zero use a kernel thread to
+    /// perform submission queue polling. After the given idle time
+    /// the thread will go to sleep.
+    #[clap(long, default_value = "100ms")]
```
kp-mariappan-ramasamy, line 87: Think it is better if we should do a sanity check for this value for upper bound like 1 or 2 seconds.
Setting this as large value will make CPU run 100% for longer time.

What do you think ?

xv-ian-c: I think sysadmins need to understand the consequences of fiddling with these kinds of low level tunables.

PR #127, opened 2024-11-18
CVPN-1445 CVPN-1555 Bring ML-KEM to lightway

kp-thomas-yau:
## Description

As WolfSSL finally supports serving ML-KEM and Kyber (by applying git patch on top of the 5.7.4 release), this PR aims to add support for both KEMs on the server-side so that old and new lightway clients would be supported simultaneously.

The client connection builder is also now using P521 ML-KEM 1024 instead of P521 Kyber Level 5

## Motivation and Context

Allow the server to use the latest and officially approved standard of post-quantum cryptography
Client use the latest NIST standard

The reason we also upgrade the client to use ML-KEM in this PR is that we can work on e2e tests to test out both Kyber and ML-KEM client can indeed connect to a server, and to also demonstrate ML-KEM works without any issue.

## How Has This Been Tested?

- cargo test
- By running lightway server and client locally with the following combos:
1. Server (serves both ML-KEM & Kyber) ↔︎ Client (uses ML-KEM w/o liboqs)
2. Server (serves both ML-KEM & Kyber) ↔︎ Client (uses Kyber w/o liboqs)
3. Server (serves both ML-KEM & Kyber) ↔︎ Client (uses Kyber WITH liboqs)

- Added some e2e tests that tests the connection with a server that supports both KEM and client that supports EITHER ML-KEM and Kyber.

## Types of changes

## Checklist:

### DIFF
--- a/tests/client/Earthfile
+++ b/tests/client/Earthfile
@@ -3,8 +3,13 @@ ARG --global debian
 
 build-container:
     ARG TOKIO_WORKER_THREADS
+    ARG KYBER_CLIENT
     FROM ../base+container --debian=$debian
-    COPY (../..+build/lightway-client --debian=$debian) .
+    IF [ "$KYBER_CLIENT" = "true" ]
+        COPY (../..+build-kyber-client/lightway-client --debian=$debian) .
+    ELSE
+        COPY (../..+build/lightway-client --debian=$debian) .
+    END
     COPY --dir ../certs+client/* tests/certs/
     COPY --dir docker-entrypoint.sh client_config.yaml run-test-inside.sh .
     HEALTHCHECK --interval=1s --timeout=1s --start-period=0s --retries=30 CMD ping -c1 nginx
--- a/Earthfile
+++ b/Earthfile
@@ -61,6 +61,17 @@ build:
 build-arm64:
     BUILD +build --ARCH="arm64"
 
+build-kyber-client:
+    FROM +source
+    ARG ARCH=$(dpkg --print-architecture)
+    LET target = "x86_64-unknown-linux-gnu"
+
+    IF [ "$ARCH" = "arm64" ]
+        SET target = "aarch64-unknown-linux-gnu" 
+    END
+    DO lib-rust+CARGO --args="build --release --features kyber_only --target=$target --bin lightway-client --target-dir ./lightway-client-kyber" --output="lightway-client-kyber/$target/release/lightway-client"
+    SAVE ARTIFACT ./lightway-client-kyber/$target/release/lightway-client AS LOCAL ./target/$target/release/lightway-client-kyber
+
 # test executes all unit and integration tests via Cargo, in the host's native platform
 test:
     FROM +source
@@ -72,6 +83,7 @@ test:
     END
 
     DO lib-rust+CARGO --args="test --target=$target"
+    DO lib-rust+CARGO --args="test --features kyber_only --target=$target"
 
 # test-miri runs tests for modules which make use of `unsafe` under Miri.
 test-miri:
--- a/tests/Earthfile
+++ b/tests/Earthfile
@@ -21,10 +21,16 @@ TEST:
     ARG CLIENT_EXTRA_ARGS=""
     ARG TEST_SCRIPT=./run-simple-test.sh
 
+    ARG KYBER_CLIENT
+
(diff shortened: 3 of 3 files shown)


lightway-core/tests/connection.rs
```diff
@@ -470,7 +470,7 @@ impl PQCrypto {
 
     fn expected_curve(&self) -> &str {
         match self {
-            PQCrypto::Enabled => "P521_KYBER_LEVEL5",
+            PQCrypto::Enabled => "P521_ML_KEM_1024",
```
xv-ian-c, line 474: If I run `cargo test -F kyber` this would fail, I think you need the same `#[cfg` stuff as in `builder.rs`.

lightway-core/src/connection/builders.rs
```diff
@@ -172,6 +172,9 @@ impl<AppState: Send + 'static> ClientConnectionBuilder<AppState> {
     /// Enable Post Quantum Crypto
     #[cfg(feature = "postquantum")]
     pub fn with_pq_crypto(self) -> Self {
+        #[cfg(not(feature = "kyber"))]
+        let curve = wolfssl::CurveGroup::P521MLKEM1024;
+        #[cfg(feature = "kyber")]
         let curve = wolfssl::CurveGroup::P521KyberLevel5;
```
xv-ian-c, line 178: It might be overkill for this simple case but FYI https://crates.io/crates/cfg-if can make these constructs read more like "regular" code.

A third option would be:
```rust
let curve = {
    #[cfg(not(feature = "kyber"))]
    wolfssl::CurveGroup::P521MLKEM1024
    #[cfg(feature = "kyber")]
    wolfssl::CurveGroup::P521KyberLevel5
};
```
which avoids duplicating the `let curve = ...`. (Assuming it is valid syntax, I'm only 80% sure and I haven't tried it...)

tests/client/Earthfile
```diff
@@ -3,8 +3,13 @@ ARG --global debian
 
 build-container:
     ARG TOKIO_WORKER_THREADS
+    ARG KYBER_CLIENT
     FROM ../base+container --debian=$debian
-    COPY (../..+build/lightway-client --debian=$debian) .
+    IF [ "$KYBER_CLIENT" = "true" ]
+        COPY (../..+build-kyber-client/lightway-client --debian=$debian) .
```
xv-ian-c, line 9: This is fine, but WDYT of adding an `ARG FEATURES`  to `+build/lightway-client` and conditionally setting the `--features` flag to cargo based on that? It appears that `cargo --features default` is valid, so you can just default it to `default` and not too much logic is needed

Then you could just forward a similar `ARG` from here.

xv-ian-c: Oh, maybe the desire to do
```
    SAVE ARTIFACT ./lightway-client-kyber/$target/release/lightway-client AS LOCAL ./target/$target/release/lightway-client-kyber
```
to a different path makes that too fiddly.

lightway-core/src/connection/builders.rs
kp-thomas-yau: I think both approach doesn't work as cfg-if does not allow expression, and there's a syntax error for the second option. 

Or
We can use cfg! macro but it will be evaluated during runtime instead of being compiled conditionally

xv-ian-c: In which case what you have is perfectly fine.

tests/client/Earthfile
kp-thomas-yau: I think it would be too fiddly and in this specific case, we just need to build a kyber client only but not kyber server. So the clearer name distinction here makes more sense for the user to know it builds a kyber client only

lightway-core/Cargo.toml
```diff
@@ -15,6 +15,7 @@ debug = ["wolfssl/debug"]
 # Enable additional APIs to support wire protocol fuzzing
 fuzzing_api = []
 postquantum = ["wolfssl/postquantum"]
+kyber = ["postquantum"]
```
kp-mariappan-ramasamy: The name `kyber` is misleading in the context of core crate.
Both kyber and ml-kem is enabled for server if we enable post-quantum. This will cause confusion.

Might be worth to rename it to something like `force_kyber`

kp-thomas-yau: Yeah I will update it. It's just a draft PR rn so I will update the naming soon when it's ready. Thanks

kp-mariappan-ramasamy: Pardon me. I was deceived by previous reviews. 
Thought it is ready to review

PR #134, opened 2024-12-02
Relicense under AGPLv3

xv-ian-c:


lightway-core/Cargo.toml
```diff
@@ -34,7 +34,7 @@ rand.workspace = true
 rand_core = "0.6.4"
 thiserror.workspace = true
 tracing.workspace = true
-wolfssl = { git = "https://github.com/expressvpn/wolfssl-rs", default-features = false}
+wolfssl = { git = "https://github.com/expressvpn/wolfssl-rs", default-features = false, version = "3.0.0" }
```
kp-mariappan-ramasamy: Do we need this version field ?
Is it just to make sure we use `GPL-2.0-or-later` license

xv-ian-c: It's not needed (AFAIK). 

All it does is cause cargo to confirm that the thing at the `git` URL has the corresponding semver (I'm unclear if it wants exact match or merely "semver compatible"), i.e. it would complain if the latest wolfssl git was actually 4.0.0 or 2.0.0. I'm not sure how valuable that is. Maybe I should just drop this bit?

xv-ian-c: I rebased so this commit went away anyway due to #136 bumping wolfssl already.

PR #148, opened 2025-01-09
WIP: quick performance boost changes

kp-omer-shamash:
## Description

feat: introduce cpu-arch, dashmap, parking_lot

* Change mutex usage from std to parking_lot (faster impl)
* Calculate checksums/ip-addr with cpu-native ASM
* Move from locked hashmap to dashmap (lockfree)

## Motivation and Context
9g5ud6


lightway-core/src/utils.rs
```diff
@@ -9,17 +9,35 @@ use std::net::Ipv4Addr;
 use std::ops;
 use tracing::warn;
 
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
```
kp-mariappan-ramasamy: I guess the changes in this file are ought to be in a separate commit ?

lightway-core/src/utils.rs
```diff
@@ -33,156 +51,290 @@ impl ops::Sub<u16> for Checksum {
...
+    fn from_ipv4_address(old: Ipv4Addr, new: Ipv4Addr) -> Self {
+        // The generic version is simpler and safe
+        Self::from_ipv4_address_generic(old, new)
+    }
 
-    let checksum = Checksum(packet.get_checksum());
-    let checksum = checksum.update(&updates);
-    packet.set_checksum(*checksum);
+    #[inline]
+    fn from_ipv4_address_generic(old: Ipv4Addr, new: Ipv4Addr) -> Self {
```
kp-mariappan-ramasamy: Sorry i failed to understand what is the use of this `from_ipv4_address_generic` fn split.

lightway-core/src/utils.rs
```diff
@@ -33,156 +51,290 @@ impl ops::Sub<u16> for Checksum {
...
-    let checksum = Checksum(packet.get_checksum());
-    let checksum = checksum.update(&updates);
-    packet.set_checksum(*checksum);
-
-    // In case of fragmented packets, TCP/UDP header will be present only in the first fragment.
-    // So skip updating the checksum, if it is not the first fragment (i.e frag_offset != 0)
-    if 0 != packet.get_fragment_offset() {
-        return;
+fn update_tcp_checksum(packet: &mut MutableIpv4Packet, updates: ChecksumUpdate) {
+    if let Some(mut tcp_packet) = MutableTcpPacket::new(packet.payload_mut()) {
```
kp-mariappan-ramasamy: Is this `if check` flipped from the original version to make branch prediction happy ?

lightway-core/src/utils.rs
```diff
@@ -33,156 +51,290 @@ impl ops::Sub<u16> for Checksum {
...
 
-/// Utility function to update destination ip address in ipv4 packet buffer
-/// Nop if buf is not a valid IPv4 packet
-pub fn ipv4_update_destination(buf: &mut [u8], ip: Ipv4Addr) {
-    let packet = MutableIpv4Packet::new(buf);
-    let Some(mut packet) = packet else {
-        warn!("Invalid packet size (less than Ipv4 header)!");
+/// Update destination IP address in an IPv4 packet
+pub fn ipv4_update_destination(buf: &mut [u8], new_ip: Ipv4Addr) {
+    if buf.len() < 20 {
```
kp-mariappan-ramasamy: This check is not needed.
The following `MutableIpv4Packet::new(buf)` will fail, if the size is less than 20 bytes

lightway-core/src/utils.rs
```diff
@@ -33,156 +51,290 @@ impl ops::Sub<u16> for Checksum {
...
+        IpNextHeaderProtocols::Tcp
+    ) {
         return None;
     }
 
+    // Get TCP packet
     let mut tcp_packet = MutableTcpPacket::new(ipv4_packet.payload_mut())?;
 
-    // Skip if the packet is not TCP SYN packet
+    // Check if it's a SYN packet
```
kp-mariappan-ramasamy: It is better to do this cosmetic changes in a separate commit.
Will make reviewer life easier :D

lightway-core/src/utils.rs
```diff
@@ -33,156 +51,290 @@ impl ops::Sub<u16> for Checksum {
...
+    /// Based on RFC-1624 [Eqn. 4]
     fn update_word(self, old_word: u16, new_word: u16) -> Self {
         self - !old_word - new_word
     }
 
+    /// Apply multiple checksum updates
+    #[allow(unsafe_code)]
     fn update(self, updates: &ChecksumUpdate) -> Self {
-        updates.0.iter().fold(self, |c, x| c.update_word(x.0, x.1))
+        if has_avx2() {
```
kp-mariappan-ramasamy: Most of our usecases are updating Ip address and MSS which are 2 words and 1 word.

I thought AVX2 impl will be useful if we are calculating the checksum of full packet. 
Is AVX2 implementation still faster than arithmetic op, even for this low word count?

Is it possible to add a quick benchmark to validate this ?

lightway-core/src/utils.rs
kp-omer-shamash: Sure, I can split them, for now it's basically just me pushing changes that don't change the architecture of the code

lightway-core/src/utils.rs
kp-omer-shamash: since AVX2 is parallel computing 8 words at a time, it is faster as long as we change around 3-4 words,
so specifically if our use-case (which I haven't fully been over) is only MSS and srcIP - then we can just use the "generic" version - I'll add comment the ability to call the AVX2 version and keep it for future reference (like needing IPv6) - good catch

lightway-core/src/utils.rs
kp-omer-shamash: i tried playing around with vector operations and see if i can sort it with checksum at once
since it never produced any good results - i removed it so we are left with the "generic"
I'll consolidate them again to just be `from_ipv4_address`

lightway-core/src/utils.rs
kp-omer-shamash: it is potentially better for branch prediction, but since we don't use any [intrinsics](https://doc.rust-lang.org/std/intrinsics/index.html) (which is not in the language-stable yet)
it won't really matter - I'll introduce proper `un/likely` API to it to ensure this helps

lightway-core/src/utils.rs
kp-omer-shamash: true, good catch

lightway-core/src/utils.rs
kp-omer-shamash: True - did it mostly for my own understanding, I'll revert it

lightway-core/src/utils.rs
```diff
@@ -33,122 +72,220 @@ impl ops::Sub<u16> for Checksum {
...
+        ];
+        let new_words = [
+            u16::from_be_bytes([new_bytes[0], new_bytes[1]]),
+            u16::from_be_bytes([new_bytes[2], new_bytes[3]]),
+        ];
+
+        Self(vec![
+            (old_words[0], new_words[0]),
+            (old_words[1], new_words[1]),
+        ])
```
kp-omer-shamash, line 187: usually loop-unrolling optimization would've fixed this for us, but since we use `vec.push` - it might miss it, so I unrolled manually

lightway-core/src/utils.rs
```diff
@@ -33,122 +72,220 @@ impl ops::Sub<u16> for Checksum {
...
-        let checksum = checksum.update(&updates);
-        packet.set_checksum(checksum.0);
+    match packet.get_next_level_protocol() {
+        IpNextHeaderProtocols::Tcp => update_tcp_checksum(packet, updates),
+        IpNextHeaderProtocols::Udp => update_udp_checksum(packet, updates),
+        IpNextHeaderProtocols::Icmp => {} // ICMP doesn't need checksum update for IP changes
+        protocol => {
+            if unlikely(true) {
+                warn!(protocol = ?protocol, "Unknown protocol, skipping checksum update");
+            }
```
kp-omer-shamash, line 205: if you are wondering on all the `unlikely(true)` - it's basically a way to ensure that the WARN gets pushed out of the HOT path, so it's intended

lightway-core/src/utils.rs
```diff
@@ -33,122 +72,220 @@ impl ops::Sub<u16> for Checksum {
...
-    // In case of fragmented packets, TCP/UDP header will be present only in the first fragment.
-    // So skip updating the checksum, if it is not the first fragment (i.e frag_offset != 0)
-    if 0 != packet.get_fragment_offset() {
-        return;
+fn update_tcp_checksum(packet: &mut MutableIpv4Packet, updates: ChecksumUpdate) {
+    if likely(MutableTcpPacket::new(packet.payload_mut()).is_some()) {
+        let mut tcp_packet = MutableTcpPacket::new(packet.payload_mut()).unwrap();
+        let checksum = tcp_packet.get_checksum();
+        // Only update if checksum is present (not 0)
+        if checksum != 0 {
```
kp-omer-shamash, line 215: added `checksum != 0` check on all layers to allow OFFLOADING to hardware when possible

.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
 runner = ["qemu-aarch64-static"] # use qemu user emulation for cargo run and test
+
+[build]
+rustflags = ["-C", "target-cpu=native"]
```
xv-pete-m: We need to be careful with `target-cpu=native`. That will compile based on the CPU of whatever build machine is used, which varies a lot, even between the CI build machines. We had trouble with this creating binaries that couldn't run on customer machines...

PR #161, opened 2025-02-25
CVPN-1818 Inside Packet Codec

kp-max-li:
## Description
Added inside packet codec interface for Lightway UDP:
- Lightway
  - Interface to include an optional inside packet codec
  - inside packets can be accumulated, and sent at once as encoded packets (vice versa).
  - Support for toggling the codec at runtime without disrupting the connection
    - Simultaneously processing encoded and un-encoded packets by introducing new inside data frame types `EncodedData` and `EncodedDataFrag`
    - New inside wire frame type `EncodingRequest` and `EncodingResponse` that allows the client to enable encoding on both sides
- Lightway Client's CLI
  - `-e` flag in CLI args to automatically send a `EncodingRequest` packet to the server once lightway-core changes state to `Connected`
  - mpsc channel for toggling encoding at runtime

## Motivation and Context
FEC code can be integrated to inside packets. Throughput and latency can be reduced under high packet loss network conditions.

## How Has This Been Tested?
1. E2E test with network namespace bash scripts. 
    - `iperf3` and `ping 8.8.8.8`

### Integration with C servers
With the network namespace scripts, running the latest version of lightway C UDP server: Verified that sending wire frames with the new ids does not break the connection with the server.

## Types of changes

## Checklist:


kp-max-li:
TODO:
- Lightway-server
- More elegant way to enable/disable the pkt flush task (Checking whether we should flush every 100 ns is very wasteful if it is not needed + locking is expensive)
- pkt accumulator config: where should we put it?
- (nice-to-have) avoid some cloning to improve performance

lightway-core/src/packet_accumulator.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+struct NoOpPacketAccumulator {
+    pkts: Vec<BytesMut>,
+}
+
+/// Indicates whether the accumulator is ready to be flushed or not
+pub enum AccumulatorState {
+    /// Ready to flush
+    ReadyToFlush,
```
kp-mariappan-ramasamy: We will need one more state, for skipping packets which are not encoded

kp-max-li: I changed a bit of the logic: Normal packets will not go through packet accumulator at all

I have added a `is_encoding_enabled` state in `Connection`. During ingress, if it is disabled, the packet will not go through the packet accumulator: https://github.com/expressvpn/lightway/pull/161/commits/7f6dd92b694fd8986d7e90436897cd2965df9619#diff-0b81b1a3bac2f54ffebe931b68e3a587d014d8cdfc36b4153f426a018d738b03R786-R788

For egress, if the wire frame type is not `EncodedData` or `EncodedDataFrag`, it will also not go through the packet accumulator:
https://github.com/expressvpn/lightway/pull/161/files#diff-0b81b1a3bac2f54ffebe931b68e3a587d014d8cdfc36b4153f426a018d738b03R1229-R1234
https://github.com/expressvpn/lightway/pull/161/files#diff-0b81b1a3bac2f54ffebe931b68e3a587d014d8cdfc36b4153f426a018d738b03R1382-R1384

lightway-server/src/main.rs
```diff
@@ -117,6 +118,16 @@ async fn main() -> Result<()> {
     let mut tun_config = TunConfig::default();
     tun_config.tun_name(config.tun_name);
 
+    let ingress_pkt_accumulator = Box::new(lightway_app_utils::RaptorEncoderFactory::new(
```
kp-mariappan-ramasamy: This should be done only for UDP connections ??
Same for client

kp-max-li: By default the packet accumulator is disabled, and lightway-client has full control of when to enable it.  The toggling is done by an async task here in lightway-client: https://github.com/expressvpn/lightway/pull/161/files#diff-76cc785c1d261df977f03e710141e50595aa4dbe21f36ea29edccc143ff96c3cR454-R466

For now (as a POC), I created signal handlers for SIGUSR1 and SIGUSR2, in which SIGUSR1 enables the pkt accumulator, and SIGUSR2 disables it. I decided to use SIGUSR1 and SIGUSR2 because it is convenient for manual tests. For later we can use other ways to enable or disable the pkt accumulator (in which we can set to never enable it if the connection is TCP)

lightway-core/src/connection.rs
```diff
@@ -344,6 +350,15 @@ pub struct Connection<AppState: Send = ()> {
 
     // Is the first outside packet received
     is_first_packet_received: bool,
+
+    // Inside ingress packet accumulator
+    pkt_accumulator_ingress: PacketAccumulatorType,
```
kp-mariappan-ramasamy: Let's make this as optional and not force apps to provide one.

lightway-core/src/connection.rs
```diff
@@ -344,6 +350,15 @@ pub struct Connection<AppState: Send = ()> {
...
+
+    // Inside egress packet accumulator
+    pkt_accumulator_egress: PacketAccumulatorType,
+
+    // Is encoding enabled
+    is_encoding_enabled: bool,
```
kp-mariappan-ramasamy: This may not be needed if packet accumulator is Some(..)

lightway-core/src/packet_accumulator.rs
```diff
@@ -0,0 +1,138 @@
+use anyhow::Result;
```
kp-mariappan-ramasamy: Anyhow shoud not be used for core library errors.
Create a custom error like `ConnectionError` which covers all possible errors and return it.

lightway-core/src/connection.rs
```diff
@@ -1333,29 +1444,42 @@ impl<AppState: Send> Connection<AppState> {
...
+                self.handle_outside_data_bytes(data, is_encoded)?;
                 Ok(())
             }
             FragmentMapResult::Incomplete => Ok(()),
             FragmentMapResult::Err(err) => Err(err.into()),
         }
     }
+
+    /// Triggers cleaning up stale egress packet accumulator states
+    pub fn clean_up_stale_egress_pkt_accumulator_states(&mut self) {
```
kp-mariappan-ramasamy: This api looks so odd. Let me go thro futrher commits to understand it better

lightway-core/src/packet_accumulator.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Build a new instance of `PacketAccumulatorType`
+    fn build(&self) -> PacketAccumulatorType;
+
+    /// Returns the accumulator name for debugging purpose
+    fn get_accumulator_name(&self) -> String;
+}
+
+/// Factory to build `Box<NoOpPacketAccumulator>`
+#[derive(Default)]
+pub struct NoopPacketAccumulatorFactory {}
```
kp-mariappan-ramasamy: This should not be inside core. If you need it, it should either be in app-utils or server/client

lightway-core/src/connection.rs
```diff
@@ -763,6 +783,49 @@ impl<AppState: Send> Connection<AppState> {
...
+                    // Go on
+                }
+                Err(ConnectionError::InvalidState) => {
+                    // Ignore the packet till the connection is online
+                }
+                Err(ConnectionError::InvalidInsidePacket(_)) => {
+                    // Ignore invalid inside packet
+                }
+                Err(err) => {
+                    // Propagate fatal error up
```
kp-mariappan-ramasamy, line 853: This will drop all the remaining packets in the current vector.
We will need a metric to monitor this usage

lightway-core/src/packet_accumulator.rs
kp-mariappan-ramasamy: This `is_encoding_enabled` should not be global parameter. It should be per packet based.

i.e Packet Accumulator takes the packet, but it can decide whether it needs to skip encoding that packet, instead of globally
In case of skipping, we need a state to notify that.

It means even with Packet accumulation enabled, data packets can go in either wire::Data or wire::EncodedData

lightway-server/src/connection_manager.rs
```diff
@@ -448,6 +448,25 @@ impl ConnectionManager {
...
+            conn.clean_up_stale_egress_pkt_accumulator_states();
+        }
+    }
+
+    pub(crate) fn flush_ingress_pkt_accumulator(
+        &self,
+    ) -> Vec<(Arc<Connection>, ConnectionResult<()>)> {
+        let connections = self.connections.lock().unwrap().get_all_connections();
+
+        connections
```
kp-mariappan-ramasamy: Worried about this loop since this mutex will content with packet receive flow

kp-mariappan-ramasamy: Can you please  test once with and without raptor q in a test machine and see the performance impact ?

lightway-core/src/packet_accumulator.rs
kp-max-li: Could we set the `PacketAccumulation`'s trait functions to return `Box<dyn std::error::Error + Sync + Send>` instead (just like the Plugins)

We have an error variant of `PacketAccumulatorError(Box<dyn std::error::Error + Sync + Send>)`.
With this, custom structs (i.e. Raptor) that implement the trait can return a custom error.

lightway-core/src/connection.rs
kp-max-li: This is perhaps needed to make the packet accumulator togglable at runtime. For now I will move this into the `PacketAccumulation` trait instead.

lightway-core/src/packet_accumulator.rs
kp-max-li: Deleted this when making accumulator optional https://github.com/expressvpn/lightway/pull/161/commits/1c9282d370ba7a0d12c6461ef58587315d2ec7e0

lightway-core/src/connection.rs
kp-max-li: Sure. I will move this to the decoder itself (after we finalize how we optimize the locking)

lightway-core/src/connection.rs
kp-max-li: Sure. Done in this commit https://github.com/expressvpn/lightway/pull/161/commits/1c9282d370ba7a0d12c6461ef58587315d2ec7e0

lightway-server/src/main.rs
kp-max-li: I have created an `-e` arg option in lightway client. When lightway's state changes to connect, a toggle encoding packet will be sent.

Also added some checks to prevent encoding from being enabled when TCP is used.

All these changes can be found in this commit https://github.com/expressvpn/lightway/pull/161/commits/e499d14ffd17c37f3b794139ee7e9db2e0f39161

Note: I have not removed the SIGUSR1 and SIGURS2 handlers for now (for testing purpose). When we finialize the PR we can easily remove it by reverting the changes in this commit: https://github.com/expressvpn/lightway/pull/161/commits/ed74cae67e2bf9cc0ce3e5c27121ac95759a9bff

lightway-core/src/packet_accumulator.rs
kp-max-li: I have moved the state to the packet accumulator. It is needed as we need the encoder to be toggled at runtime.

I have also modified the packet accumulator to return a variant called `Skip`, code in `Connection` is also updated to act accordingly.

All these changes can be found in this commit: https://github.com/expressvpn/lightway/pull/161/commits/1ac4302961ed56f7a6f23d7546154d60a8a95f69

lightway-server/src/connection_manager.rs
kp-max-li: FYI some testings suggests that:
- Main branch: 100% bandwidth
- This branch, with no raptor enabled: 90% bandwidth
- With raptor enabled: 40 - 45% bandwidth

But I will do some more testing in the future after we finalize how we optimize the implementation.

lightway-core/src/connection.rs
kp-max-li: Done in this commit https://github.com/expressvpn/lightway/pull/161/commits/1ac4302961ed56f7a6f23d7546154d60a8a95f69

lightway-core/src/packet_accumulator.rs
kp-mariappan-ramasamy: Yes sounds good

lightway-client/src/lib.rs
```diff
@@ -154,6 +157,24 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
...
+
+    /// How often the pkt encoder is flushed
+    pub pkt_encoder_flush_interval: Duration,
+
+    /// How often the pkt decoder's states are cleaned up
+    /// If None, the task will not be spawn
+    pub pkt_decoder_clean_up_interval: Duration,
+
+    /// Enables inside packet encoding when connection is established.
+    pub enable_inside_pkt_encoding_at_connect: bool,
```
kp-mariappan-ramasamy: Do we need this ? `inside_pkt_codec: Option<PacketCodecFactoryType>,` is an Option.
Is it not enough ?

kp-max-li: I would like to cater a scenario where the client does not want to turn on the codec right after connection (e.g. client app only turns it on when it detects packet loss)

This flag is only for triggering a ToggleEncoding packet when we receive a state change to Online

So like the flag's purpose is only for conveniently enabling the encoding from the CLI right from the start of the connection

whereas even if this flag is disabled and there is Some packet codec, the user can still enable it afterwards manually

lightway-core/src/packet_codec.rs
```diff
@@ -0,0 +1,86 @@
...
+/// PacketEncoder and PacketDecoder's trait function's return type
+pub type PacketCodecResult<T> = std::result::Result<T, Box<dyn std::error::Error + Sync + Send>>;
+
+/// PacketEncode trait. Accumulates inside packets and turn them to encoded packets
+pub trait PacketEncoder {
+    /// Store one inside packet into the PacketEncoder
+    ///
+    /// Returns a [`CodecStatus`].
+    /// If the status is [`CodecStatus::ReadyToFlush`], encoded packets are ready to be retrieved.
+    /// If the status is [`CodecStatus::SkipPacket`], the packet should not be added to the encoder.
```
kp-mariappan-ramasamy: Does not this mean `the packet is skipped by the encoder and lightway can send it as normal Wire::Data packet.

lightway-core/src/packet_codec.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Retrieve the decoded inside packets
+    fn get_decoded_pkts(&mut self) -> PacketCodecResult<Vec<BytesMut>>;
+
+    /// Clean up the inner stale states.
+    /// Should be called periodically to avoid stale states
+    /// from unnecessarily holding memory.
+    fn cleanup_stale_states(&mut self);
+}
+
+/// Indicates whether the [`PacketEncoder`] or [`PacketDecoder`] is ready to be flushed or not
```
kp-mariappan-ramasamy: Indicates the status of [`PacketEncoder`] or [`PacketDecoder`] after storing the current packet

lightway-server/src/main.rs
```diff
@@ -137,6 +138,9 @@ async fn main() -> Result<()> {
         key_update_interval: config.key_update_interval.into(),
         inside_plugins: Default::default(),
         outside_plugins: Default::default(),
+        inside_pkt_codec: None,
+        pkt_encoder_flush_interval: Duration::from_secs_f64(0.0001),
+        pkt_decoder_clean_up_interval: Duration::from_secs_f64(0.5),
```
kp-mariappan-ramasamy: Can we make this values as cli args ?

lightway-server/src/lib.rs
```diff
@@ -265,30 +377,25 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
...
-                }
+                let result = conn.inside_data_received(&mut buf);
+                handle_inside_io_error(conn, result);
             } else {
                 metrics::tun_rejected_packet_no_connection();
             }
         }
     });
 
+    tokio::spawn(pkt_encoder_flush(
```
kp-mariappan-ramasamy: This two tasks should be created only if `config.inside_pkt_codec` is not None, should not this ?

lightway-client/src/main.rs
```diff
@@ -97,6 +99,9 @@ async fn main() -> Result<()> {
         server: config.server,
         inside_plugins: Default::default(),
         outside_plugins: Default::default(),
+        inside_pkt_codec: None,
+        pkt_encoder_flush_interval: Duration::from_secs_f64(0.0001),
+        pkt_decoder_clean_up_interval: Duration::from_secs_f64(0.5),
```
kp-mariappan-ramasamy: Same as server, better this be an cli config

lightway-client/src/lib.rs
```diff
@@ -351,6 +371,96 @@ async fn handle_network_change(
...
+
+    let maybe_encoder_weak = conn.lock().unwrap().get_inside_packet_encoder();
+
+    let encoder_weak = match maybe_encoder_weak {
+        Some(encoder_weak) => encoder_weak,
+        None => {
+            // encoder is not set. Awaits forever...
+            // Not returning immediately to avoid triggering an exit of the
+            // entire client program.
+            tracing::debug!("Pkt encoder flush task is freezed");
```
kp-mariappan-ramasamy: FYI, tokio::select supports enabling the branch only on valid condition (valid packet codec)
You might use it to disable this task check instead of  this.

lightway-client/src/lib.rs
```diff
@@ -164,6 +164,13 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback> {
     /// How often the packet decoder's states are cleaned up
     pub pkt_decoder_clean_up_interval: Duration,
 
+    /// Enables inside packet encoding when connection is established.
+    pub enable_inside_pkt_encoding_at_connect: bool,
+
+    /// Signal for toggling inside packet encoding
+    #[educe(Debug(ignore))]
+    pub toggle_encoding_signal: tokio::sync::mpsc::Receiver<bool>,
```
kp-mariappan-ramasamy: We should not force apps to provide this value. Better to make this as option similar to `inside_pkt_codec`

kp-mariappan-ramasamy: Or should we create a new PacketEncoderConfig and add all these variables inside it and make it as an option.
It reduces the burden for apps to skip this, if codec is not needed

lightway-core/src/packet_codec.rs
kp-max-li: Yes thats a better description. Changed.

lightway-server/src/main.rs
kp-max-li: Done

lightway-core/src/connection.rs
kp-max-li: Done

lightway-server/src/lib.rs
kp-max-li: Yeah sure. Done.

lightway-client/src/main.rs
kp-max-li: Done.

lightway-client/src/lib.rs
kp-max-li: Sure. I have created it.

lightway-client/src/lib.rs
kp-max-li: Done

kp-mariappan-ramasamy (approved):

PR #169, opened 2025-03-13
app-utils: conditionally build libc constants for Apple instead

kp-thomas-yau:
## Description

Since iOS also doesn't support SOL_IP, we are now updating the some of the libc constant to be built conditionally based on `target_vendor = "apple"` instead of just MacOS.

We have also updated `socket_enable_pktinfo` to be built/included only when we are not using an apple platform.

## Motivation and Context
Originally, `socket_enable_pktinfo` can't be built on Mac/iOS platform because mac/ios platform doesn't have `SOL_IP` in libc. So this PR fixes this, and allows us to build libraries on apple platforms.

## How Has This Been Tested?

Tested our internal clients can be built on iOS platform (`aarch64-apple-ios`, `aarch64-apple-ios-sim`) when it's referencing this new branch and includes `lightway-core`, `lightway-client`, `lightway-app-utils` in their dependency

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/sockopt.rs
+++ b/lightway-app-utils/src/sockopt.rs
@@ -5,4 +5,5 @@ mod ip_mtu_discover;
 mod ip_pktinfo;
 
 pub use ip_mtu_discover::*;
+#[cfg(not(target_vendor = "apple"))]
 pub use ip_pktinfo::*;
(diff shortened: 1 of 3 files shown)


kp-mariappan-ramasamy (approved):
LGTM

PR #176, opened 2025-03-25
Move mutex outside

kp-mariappan-ramasamy:
## Description

Cleanup packet codec traits

## Motivation and Context

Move the responsibility of maintaining the codec to the external application from lightway-core

## How Has This Been Tested?
Verified manually

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -419,7 +419,7 @@ async fn pkt_encoder_flush(
 
     let maybe_encoder_weak = conn.lock().unwrap().get_inside_packet_encoder();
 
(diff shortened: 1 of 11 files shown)


lightway-client/src/lib.rs
```diff
@@ -419,7 +419,7 @@ async fn pkt_encoder_flush(
 
     let maybe_encoder_weak = conn.lock().unwrap().get_inside_packet_encoder();
```
kp-max-li: Shall we remove `_weak` from the variable name since `get_inside_packet_encoder()` returns an `Arc` now.

kp-max-li (dismissed):

PR #188, opened 2025-04-09
KM-8884: Mock send success with non fatal UDP send error

kp-max-li:
## Description
With non-fatal UDP send errors, return the number of bytes requested to be sent to mock that the send is successful.
Switched to use `NetworkUnreachable` variant in `std::io::ErrorKind` as it is no longer a part of the nightly features.

## Motivation and Context
Currently, WolfSSL perceives that no data is sent and try to send the same data again, creating a live-lock until the connection is available.
This is not good as the thread/async task that calls this will be stuck looping whereas holding the `core::Connection` lock at the same time.

## How Has This Been Tested?
Tested with CI.

Android:
- Verified that the app does not hang when it loses connection from the WAN (emulator, physical device)
- UDP floating still works (Android emulator)

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -82,23 +82,16 @@ impl OutsideIOSendCallback for Udp {
                 // Swallow the error so the WolfSSL socket does not
                 // enter the error state.
                 //
-                // This way we can continue if/when the server shows up.
+                // This way we can continue if/when the server shows up. 
                 IOCallbackResult::Ok(0)
             }
-            Err(err) if matches!(err.raw_os_error(), Some(libc::ENETUNREACH)) => {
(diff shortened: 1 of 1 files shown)


lightway-client/src/io/outside/udp.rs
```diff
@@ -82,23 +82,16 @@ impl OutsideIOSendCallback for Udp {
                 // Swallow the error so the WolfSSL socket does not
                 // enter the error state.
                 //
-                // This way we can continue if/when the server shows up.
+                // This way we can continue if/when the server shows up.
```
kp-mariappan-ramasamy: Unintented change ?

kp-mariappan-ramasamy (approved):
LGTM with minor nit

lightway-client/src/io/outside/udp.rs
kp-max-li: Yeah.. fixed this in the second commit

PR #194, opened 2025-04-28
CVPN-1958 Allow inside codec to actively flush the encoded/decoded pkts

kp-max-li:
## Description
- Added an encoded/decoded packet mpsc channel interface to PacketCodec.
    - Whenever a packet is ready to be emitted by the encoder/decoder, it is sent to the mpsc channel, and the client/server is responsible for sending the packets to the core.
    - packets are no longer sent as a batch, and is sent one by one when received from the mpsc channel.
- Removed the periodic flush logic of the encoder

## Motivation and Context
For now:
- Flushing of encoded packets can only be triggered by the periodic flush task and the receive of inside data 
- Flushing of decoded packets can only be triggered by the receive of outside data

which is not ideal because the codec cannot actively trigger a send to the core when the packets are ready, and can only passively wait for a flush.
If the packets sit in the codec for too long, the delay could be huge.

Hence, logic is needed to allow the codec to actively trigger a send to core.

## How Has This Been Tested?
Integrated Raptor Codec with the new changes
Do manual end-to-end test with the network namespace scripts.
Added debugging log lines and ran `ping` and `iperf3` and expected results are shown.

## Types of changes

## Checklist:

### DIFF
--- a/docs/logs_and_metrics.md
+++ b/docs/logs_and_metrics.md
@@ -32,7 +32,6 @@ Lightway server also supports metrics to monitor. The following are the metrics 
 | conn_alloc_frag_map | core | Counter | A connection has used a fragmented data packet.<br>Therefore the 2M FragmentMap has been allocated |
 | wolfssl_appdata | core | Counter | An AppData result occurred during a WolfSSL operation.<br><br>Given current configuration we do not expect this to be non-zero |
 | session_id_mismatch | core | Counter | Server has received a mismatched session_id in the header after the packet content has been validated <br><br>Should generally be expected to happen rarely|
-| inside_pkt_dropped_due_to_fatal_err | core | Counter | The number of inside packets dropped due to a fatal error when trying to send packets to outside
 | received_encoding_req_non_online | core | Counter | Server received an encoding request when the Connection state is not Online |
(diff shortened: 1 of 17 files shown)


lightway-core/Cargo.toml
```diff
@@ -33,6 +33,7 @@ parking_lot.workspace = true
 pnet.workspace = true
 rand.workspace = true
 rand_core = "0.9.3"
+tokio.workspace = true
```
kp-mariappan-ramasamy: We designed Lightway core to be sync and do not want to add tokio to lightway-core.

kp-mariappan-ramasamy:
Core idea behind this cleanup is good.

Only thing to avoid is using tokio inside lightway core. This will vendor lock our library and make core not usable in systems which does not support tokio like embedded platforms.
This is the conscious decision we took, when we design lightway core and apps.

lightway-core/Cargo.toml
kp-max-li: Removed it and moved the factory to app-utils

kp-mariappan-ramasamy (approved):
LGTM now. Thanks for updating

PR #197, opened 2025-05-12
Add `set_logging_callback` when `debug` feature is on

kp-thomas-yau:
## Description

It enables debug logs from WolfSSL and also sets the callback for the logs for easier debugging. We also expose the function in `lightway-client` as well so client app does not need to depend on lightway-core

## Motivation and Context

Easier to read and get debug logs from WolfSSL without jumping a bunch of hops 

## How Has This Been Tested?

Tested with our internal Android library, which does in fact export logs from WolfSSL to logcat

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:


kp-mariappan-ramasamy (approved):
LGTM

PR #210, opened 2025-06-04
Client: Bind to v4 or v6 based on server Ip protocol

kp-mariappan-ramasamy:
## Description

Until now, we were hardcoding UDP client socket to bind to Ipv4 address, before connecting to server.
This prevents the client from connecting to Ipv6 server.

With this change, client UDP sockets will be bound to v4/v6 protocol, based on server ip address.

Note that this is not needed for TCP since we connect to server address directly which will take care of creating the socket with correct protocol. For UDP, there is no such api which can connect directly. Only way is to bind to local socket and then use it.

## Motivation and Context

Found Ipv6 UDP traffic is not working with Ipv6 servers.

## How Has This Been Tested?

Manually verified lightway client and server can communicate using Ipv6 protocol using Ipv4 in tunnel.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -1,6 +1,6 @@
 use anyhow::{Result, anyhow};
 use async_trait::async_trait;
-use std::{net::SocketAddr, sync::Arc};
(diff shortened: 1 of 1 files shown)


kp-max-li (approved):
lgtm

PR #214, opened 2025-06-18
CVPN-1935: Add `set_outside_io` to update outside IO callback

kp-thomas-yau:
## Description

Introduce `set_new_outside_io` to update outside IO callback in WolfSSL session.

## Motivation and Context

This can be used when the client apps need to destroy/recreate the outside socket to make sure the same WolfSSL session can still be used for this connection.

## How Has This Been Tested?

Tested this on a test iPhone, verified that the WolfSSL session is using the new callback arguments properly.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -20,7 +20,6 @@ use thiserror::Error;
 use tracing::{debug, error, info, trace, warn};
 use wolfssl::{ErrorKind, IOCallbackResult, ProtocolVersion};
 
-use crate::max_dtls_mtu;
 use crate::{
     ConnectionType, IPV4_HEADER_SIZE, InsideIOSendCallbackArg, PluginResult, SessionId,
     TCP_HEADER_SIZE, Version,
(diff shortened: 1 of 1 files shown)


lightway-core/src/connection.rs
```diff
@@ -738,6 +738,11 @@ impl<AppState: Send> Connection<AppState> {
         Ok(auth_handle.expired())
     }
 
+    /// Update WolfSSL Session to use the new outside IO Callback
+    pub fn set_new_outside_io(&mut self, new_io: OutsideIOSendCallbackArg) {
```
kp-mariappan-ramasamy:
```suggestion
    pub fn set_outside_io(&mut self, new_io: OutsideIOSendCallbackArg) {
```
Let's makes this consistent with other apis.

kp-mariappan-ramasamy: Sorry am i mistaken. I still see `set_new_outside_io`, but commit message says `set_outside_io`

kp-thomas-yau: Damn I must be blind lol, just pushed it

PR #221, opened 2025-07-02
Add nix flake for lightway server and client

kp-mariappan-ramasamy:
## Description

Add flake.nix to make it easier for building and testing client and server from nix enabled machines.

The flake works in x64, arm64 linux and mac machines.
This PR has also changes so that lightway can be built and run successfully in macbook.

## Motivation and Context

Make lightway available in more platforms/distros.

## How Has This Been Tested?

Verified both client and server can be bulit in mac book and linux machines.
Also added CI to run build using nix flake

## Types of changes

## Checklist:


README.md
```diff
@@ -223,6 +223,23 @@ docker compose -f tests/e2e/docker-compose.yml exec client bash
 
 To run things within the containers
 
+## Nix flake
+
+Lightway can also be build and test using nix commands:
```
kp-weiwen-goh: Should be `built and tested`.

kp-weiwen-goh (approved):
LGTM. Managed to build on my machine.

flake.nix
```diff
@@ -0,0 +1,87 @@
...
+        _module.args.pkgs = import inputs.nixpkgs {
+          inherit system;
+          overlays = [(import inputs.rust-overlay)];
+        };
+
+        packages.default = self'.packages.lightway-client;
+        devShells.default = self'.devShells.stable;
+
+        packages.lightway-client = rustPackage "lightway-client" "";
+        packages.lightway-server = rustPackage "lightway-server" "";
```
kp-weiwen-goh: Do we need to add the `io-uring` feature on Linux?

kp-mariappan-ramasamy: Good idea. Added for linux

flake.nix
```diff
@@ -0,0 +1,89 @@
+{
+  description = "Lightway flake";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
```
kp-samuel-tam, line 5: Since we not building anything on NixOS (yet), I guess we don't need to wait for the NixOS-tested branch and can use fresher set of packages closer to `master`.

```suggestion
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
```

flake.nix
```diff
@@ -0,0 +1,89 @@
+{
```
kp-samuel-tam, line 1: Nix has an "official" standardised formatter, `nixpkgs#nixfmt-rfc-style`. We can use that if we want.

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+        cargoMemberToml = package: builtins.fromTOML (builtins.readFile ./${package}/Cargo.toml);
+        msrv = cargoToml.workspace.package.rust-version;
+
+        rustPackage = package: features:
+          (pkgs.makeRustPlatform {
+            cargo = pkgs.rust-bin.stable.latest.minimal;
+            rustc = pkgs.rust-bin.stable.latest.minimal;
+          }).buildRustPackage {
+            inherit ((cargoMemberToml package).package) name version;
+            src = ./.;
```
kp-samuel-tam, line 49:
```suggestion
            src = inputs.self;
```

kp-weiwen-goh (approved):

flake.nix
```diff
@@ -0,0 +1,89 @@
+{
+  description = "Lightway flake";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    flake-parts.url = "github:hercules-ci/flake-parts";
```
kp-samuel-tam, line 6: I advise against using unnecessary non-standard libraries like `flake-parts` and `flake-utils`, IMHO they do templating wrong; (1) we will now have one more third-party dependency, (2) we hide complexity that is needed to be visible for newcomers to understand the nix syntax, and (3) we also lose flexibility when it comes to cross-compilation (this happened in the shim repo with me and Kenneth).

Plus wrapping around the systems should only take one builtin function.
```nix
  outputs = { self, nixpkgs }: let
    forAllSystems = with nixpkgs.lib; genAttrs platforms.unix;
  in {
    packages = forAllSystems (system: rec {
      default = ...
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+        msrv = cargoToml.workspace.package.rust-version;
+
+        rustPackage = package: features:
+          (pkgs.makeRustPlatform {
+            cargo = pkgs.rust-bin.stable.latest.minimal;
+            rustc = pkgs.rust-bin.stable.latest.minimal;
+          }).buildRustPackage {
+            inherit ((cargoMemberToml package).package) name version;
+            src = ./.;
+            cargoLock.lockFile = ./Cargo.lock;
```
kp-samuel-tam, line 50: I suggest to use a fixed hash here since we have a git dependency `wolfssl`, which must be hash-locked. We can hash-lock the entire vendor and remove `cargoLock.outputHashes`.

If we need a fixed hash regardless, we can at least write less code.

```suggestion
            cargoHash = "sha256-FYHXa4a1g6GJP5D+6o1EI/J2lNb8hv3SxPMYVQrOUqA=";
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+            buildFeatures = features;
+            buildInputs = runtimeDeps;
+            nativeBuildInputs = buildDeps;
+            cargoBuildFlags = "-p ${package}";
+            checkType = "debug";
+            cargoLock.outputHashes = {
+              "wolfssl-3.0.0" = "sha256-ToCpJ8moKMT2rMiBmMQLnPNAme3F2VVzHQ0jVXx3kTs=";
+            };
```
kp-samuel-tam, line 58:
```suggestion
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+            rustc = pkgs.rust-bin.stable.latest.minimal;
+          }).buildRustPackage {
+            inherit ((cargoMemberToml package).package) name version;
+            src = ./.;
+            cargoLock.lockFile = ./Cargo.lock;
+            buildFeatures = features;
+            buildInputs = runtimeDeps;
+            nativeBuildInputs = buildDeps;
+            cargoBuildFlags = "-p ${package}";
+            checkType = "debug";
```
kp-samuel-tam, line 55: I think we should add a comment about why we cannot test in release mode. I spent some time looking at the how assert macros work in release/debug mode. 😅

```suggestion
            # Some tests rely on debug_assert! and fail in release.
            checkType = "debug";
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+            buildInputs = runtimeDeps;
+            nativeBuildInputs = buildDeps ++ devDeps ++ [rustc];
+          };
+        clientFeatures = [] ++ lib.optional pkgs.stdenv.isLinux [ "io-uring" ];
+        serverFeatures = [] ++ lib.optional pkgs.stdenv.isLinux [ "io-uring" ];
+      in {
+        _module.args.pkgs = import inputs.nixpkgs {
+          inherit system;
+          overlays = [(import inputs.rust-overlay)];
+        };
```
kp-samuel-tam, line 75: We can probably avoid writing these `import` calls, they are discouraged in flakes.
https://nixcademy.com/posts/mastering-nixpkgs-overlays-techniques-and-best-practice/#with-flakes

kp-samuel-tam (commented):
Merged already but leaving this for future reference. :D

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+          rustPlatform.bindgenHook
+          openssl
+          gnumake
+          autoconf
+          automake
+          libtool
+          libevent
+          cargo-nextest
+          cargo-deny
+          cargo-outdated
```
kp-samuel-tam, line 33: Do we need the extra dependencies?

```suggestion
          autoconf
          automake
          libtool
          rustPlatform.bindgenHook
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+      systems = ["x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin"];
+      perSystem = {
+        config,
+        self',
+        pkgs,
+        lib,
+        system,
+        ...
+      }: let
+        runtimeDeps = with pkgs; [];
```
kp-samuel-tam, line 21: We can remove this and all the `buildInputs`. We can always `overrideAttrs` it if we need.
```suggestion
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+        rustPackage = package: features:
+          (pkgs.makeRustPlatform {
+            cargo = pkgs.rust-bin.stable.latest.minimal;
+            rustc = pkgs.rust-bin.stable.latest.minimal;
+          }).buildRustPackage {
+            inherit ((cargoMemberToml package).package) name version;
+            src = ./.;
+            cargoLock.lockFile = ./Cargo.lock;
+            buildFeatures = features;
+            buildInputs = runtimeDeps;
```
kp-samuel-tam, line 52:
```suggestion
```

flake.nix
```diff
@@ -0,0 +1,89 @@
...
+              "wolfssl-3.0.0" = "sha256-ToCpJ8moKMT2rMiBmMQLnPNAme3F2VVzHQ0jVXx3kTs=";
+            };
+          };
+
+        mkDevShell = rustc:
+          pkgs.mkShell {
+            shellHook = ''
+              export RUST_SRC_PATH=${pkgs.rustPlatform.rustLibSrc}
+            '';
+            buildInputs = runtimeDeps;
```
kp-samuel-tam, line 66:
```suggestion
```

flake.nix
kp-mariappan-ramasamy: Oops, some of them are pulled from my local dev which I used for other purposed.
Also some of this need to go to devDeps for local shell. Will sort it.

flake.nix
kp-mariappan-ramasamy: This will help me to add any runtime deps which will be added later.
I just left the template there for easier addition later.
It will not do any harm in leaving this template

flake.nix
kp-mariappan-ramasamy: What is the use of this ?

flake.nix
kp-mariappan-ramasamy: I was planning to make this wolfssl as a crate dep instead of git, since we do not update it often.
And moreover, this gives clear idea of what all git deps are present in the crate.
So I will leave this as it is

flake.nix
kp-mariappan-ramasamy: Yes I am aware of this.  IMO we are a end user trying to use nix for building and testing in reproducible way. I don't expect this repo to a place to teach nix complexity.
Using flake-parts abstracts `system` and make us write readable code, which I think will be easier for newcomers.
I have not used it for cross-compilation yet and may be you are right. We can change it, if it is a problem.

kp-mariappan-ramasamy:
@kp-samuel-tam Thanks for the suggestions. I am creating a new PR for addressing some of these suggestions

flake.nix
kp-samuel-tam: Just nitpick 😄 I usually choose `self` over `./.`, avoid paths where I can. It's consistent with upstream's style: https://github.com/NixOS/nix/blob/c79b8018c07381a25d113ca145115ad055942d09/src/nix/flake.md?plain=1#L412

PR #227, opened 2025-07-15
feat: let ServerAuth determine if inside packet encoding can be used

kp-weiwen-goh:
By default, in `lightway-server`, inside packet encoding is enabled. The application can choose to override this behaviour in its `ServerAuth` implementation.

## Description
A new `features()` function is added to `ServerAuthHandle`. This allows the Server to specify which features are available to the connection during authorization. An application can choose to override this field in its `ServerAuth` implementation.

## Motivation and Context
This will allow applications to control which connections can use inside packet encoding.

## How Has This Been Tested?
Unit tests are added to check if `can_use_inside_pkt_encoding` defaults to true. This maintains existing behaviour, so is not breaking.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/context/server_auth.rs
+++ b/lightway-core/src/context/server_auth.rs
@@ -1,15 +1,17 @@
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
 use bytes::Bytes;
 use tracing::info;
 
-use crate::{Version, wire};
+use crate::{LightwayFeature, Version, wire};
 
 /// A handle onto a successful auth result.
 pub trait ServerAuthHandle: std::fmt::Debug {
     /// Validate if this authentication is still valid, returns
     /// true if it has expired.
     fn expired(&self) -> bool;
+    /// All features available to this connection.
+    fn features(&self) -> HashSet<LightwayFeature>;
 }
 
 /// Result of [`ServerAuth`] `authorize_*` methods.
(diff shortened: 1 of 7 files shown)


lightway-core/src/context/server_auth.rs
```diff
@@ -23,6 +25,10 @@ pub enum ServerAuthResult {
         /// if [`Some`] then the connection will
         /// switch to that tunnel protocol version.
         tunnel_protocol_version: Option<Version>,
+        /// Whether the server will accept EncodingRequests. This defaults to
+        /// true, but the application can override this behaviour in its
+        /// ServerAuth implementation
+        can_use_inside_pkt_encoding: bool,
```
kp-mariappan-ramasamy: We do not need this, if we have features() api return whether it supports InsidePktCodec, do we ?

lightway-core/src/metrics.rs
```diff
@@ -42,6 +44,11 @@ pub(crate) fn session_id_mismatch() {
     METRIC_SESSION_ID_MISMATCH.increment(1);
 }
 
+/// Server received an encoding request when the Connection does not have authorization to use inside packet encoding
```
kp-mariappan-ramasamy:
```suggestion
/// Server received an encoding request when the client does not have authorization to use inside packet encoding
```
Client might have been more suitable ?

kp-mariappan-ramasamy (approved):

PR #228, opened 2025-07-16
Increase tests timeout for riscv CI

kp-ben-hung:
## Description
Increase test timeout from 2000ms to 5000ms.

## Motivation and Context

This change is needed to solve flaky tests, especially the RISC-V tests, where emulation incurs a significant performance overhead.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/tests/connection.rs
+++ b/lightway-core/tests/connection.rs
@@ -616,6 +616,12 @@ impl PQCrypto {
     }
 }
 
+fn get_test_timeout() -> u64 {
+    #[cfg(target_arch = "riscv64")]
+    return 5000;
+    2000
+}
+
 async fn run_test_tcp<S: TestSock>(
     cipher: Option<Cipher>,
     pqc: PQCrypto,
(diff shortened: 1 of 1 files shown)


xv-thomas-leong (dismissed):

lightway-core/tests/connection.rs
```diff
@@ -640,7 +640,7 @@ async fn run_test<S: TestSock>(
         )
     };
 
-    tokio::time::timeout(std::time::Duration::from_millis(2000), test)
+    tokio::time::timeout(std::time::Duration::from_millis(5000), test)
```
kp-mariappan-ramasamy: We could consider making this timeout as a conditional value based on target.
This will avoid missing cases later, if we break the default connection time in mainstream platforms.

kp-mariappan-ramasamy (approved):
LGTM

lightway-core/tests/connection.rs
```diff
@@ -616,6 +616,12 @@ impl PQCrypto {
     }
 }
 
+fn get_test_timeout() -> u64 {
+    #[cfg(target_arch = "riscv64")]
+    return 5000;
```
kp-mariappan-ramasamy, line 621: I thought we needed !cfg for default or clippy will scream.
But we dont run clippy for riscv, so it passed

PR #231, opened 2025-07-16
External inside io

xv-carlos-c:
## Description

This PR adds support for providing external inside io instead of the inbuilt tunnel implementation.
This was done previously for server and missing in client.

## Motivation and Context

Give external applications more flexibility to use tunnel interface.

## How Has This Been Tested?

Verified the existing tunnel works as expected. 
Also verified with a external tunnel interface.

## Types of changes

## Checklist:


kp-brandon-chew (approved):
LGTM

PR #234, opened 2025-07-22
client: Specified target_os for route_manager

kp-brandon-chew:
This prevents non linux OS from compiling the routemanager code.
RouteManager does not support Ios android and tvos
## Types of changes

### DIFF
--- a/lightway-client/src/args.rs
+++ b/lightway-client/src/args.rs
@@ -87,6 +87,13 @@ pub struct Config {
     ///     noexec : Does not setup any routes
     ///     lan    : Sets up default + additional lan routes
     #[clap(long, value_enum, default_value_t = RouteMode::Default)]
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "openbsd",
+        target_os = "windows"
+    ))]
     pub route_mode: RouteMode,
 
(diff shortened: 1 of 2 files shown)


lightway-client/src/args.rs
```diff
@@ -87,6 +87,13 @@ pub struct Config {
     ///     noexec : Does not setup any routes
     ///     lan    : Sets up default + additional lan routes
     #[clap(long, value_enum, default_value_t = RouteMode::Default)]
+    #[cfg(any(
+        target_os = "freebsd",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "openbsd",
+        target_os = "windows"
+    ))]
```
kp-thomas-yau, line 96: Not sure if it's better, but should we just compile route manager when we are not building on tvos/ios/android?

kp-brandon-chew: These are all the platforms currently supported by route manager. 
Using (Not tvos/ios/android) will include platforms that route manager does not support.
I guess the trade-off is that since its exactly defined, no one ever needs to request for an exclusion to use lightway-client for any possible new platforms.

kp-thomas-yau: Ah cool cool

kp-thomas-yau (approved):

PR #239, opened 2025-07-25
client: Parallel connect

kp-weiwen-goh:
Adds support for multiple server connections in lightway-client. The servers are connected in parallel and after the "best connection" is chosen, all other connections are disconnected.

## Description
The configuration file now supports specifying multiple servers via the `servers` field, and falls back to the single-server configuration if `servers` is not specified. This maintains full backward compatibility with existing server configurations.

The client attempts to connect to all servers at once. After the first connection is established, if `defer_connect_timeout` is set, we wait that duration before choosing the best connection. The best connection is the first connection in the order of the original list of servers that manages to connect within the specified `defer_connect_timeout`.

## Motivation and Context

Improves connection speed compared to connecting to each server sequentially.

## How Has This Been Tested?

Configuration parsing tests for single and multi-server configurations
e2e test for connecting to multiple servers at once

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/args/connection_type.rs
+++ b/lightway-app-utils/src/args/connection_type.rs
@@ -3,13 +3,14 @@ use serde::{Deserialize, Serialize};
 
 use lightway_core::ConnectionType as LWConnectionType;
 
-#[derive(Copy, Clone, ValueEnum, Debug, Serialize, Deserialize)]
+#[derive(Copy, Clone, ValueEnum, Debug, Serialize, Deserialize, Default)]
 #[serde(rename_all = "lowercase")]
 /// [`lightway_core::ConnectionType`] wrapper compatible with clap and twelf
 pub enum ConnectionType {
     /// UDP (Datagram)
     Udp,
     /// TCP (Stream)
+    #[default]
     Tcp,
 }
 
--- a/lightway-client/src/args.rs
+++ b/lightway-client/src/args.rs
@@ -93,6 +93,11 @@ pub struct Config {
     #[clap(long, default_value = "0s")]
     pub keepalive_timeout: Duration,
 
+    /// How long to wait before selecting the best connection. If the preferred
+    /// connection connects before the timeout, it will be used immediately.
+    #[clap(long, default_value = "0s")]
+    pub preferred_connection_wait_interval: Duration,
+
     /// Socket send buffer size
     #[clap(long)]
     pub sndbuf: Option<ByteSize>,
--- a/lightway-client/src/args.rs
+++ b/lightway-client/src/args.rs
@@ -4,6 +4,7 @@ use bytesize::ByteSize;
 use clap::Parser;
 use lightway_app_utils::args::{Cipher, ConnectionType, Duration, LogLevel};
 use lightway_core::{AuthMethod, MAX_OUTSIDE_MTU};
+use serde::Serialize;
 use std::{net::Ipv4Addr, path::PathBuf};
 use twelf::config;
 
@@ -15,9 +16,31 @@ pub struct Config {
     #[clap(short, long)]
     pub config_file: PathBuf,
 
+    /// Servers to attempt to connect to. Configuration is only supported in
+    /// config file, not command line or environment variable
+    #[clap(skip)]
+    #[serde(default)]
+    pub servers: Vec<ConnectionConfig>,
+
+    /// Server to connect to in <hostname>:<port> format
+    /// Only used if `servers` is empty
+    #[clap(short, long, default_value_t)]
+    pub server: String,
+
     /// Connection mode
+    /// Only used if `servers` is empty
     #[clap(short, long, value_enum, default_value_t = ConnectionType::Tcp)]
     pub mode: ConnectionType,
+
+    /// Server domain name
+    /// Only used if `servers` is empty
+    #[clap(long, default_value = None)]
+    pub server_dn: Option<String>,
+
+    /// Cipher to use for encryption
+    /// Only used if `servers` is empty
+    #[clap(long, value_enum, default_value_t = Cipher::Aes256)]
+    pub cipher: Cipher,
 
     /// Auth token
     /// If both token and user/pass are provided, token auth will
@@ -56,10 +79,6 @@ pub struct Config {
     /// DNS IP to use in Tun device
     #[clap(long, default_value = "100.64.0.1")]
     pub tun_dns_ip: Ipv4Addr,
-
-    /// Cipher to use for encryption
-    #[clap(long, value_enum, default_value_t = Cipher::Aes256)]
-    pub cipher: Cipher,
 
     /// Enable Post Quantum Crypto
     #[cfg(feature = "postquantum")]
@@ -118,14 +137,6 @@ pub struct Config {
     #[clap(long, default_value = "100ms")]
     pub iouring_sqpoll_idle_time: Duration,
 
-    /// Server domain name
-    #[clap(long, default_value = None)]
-    pub server_dn: Option<String>,
-
-    /// Server to connect to
-    #[clap(short, long, default_value_t)]
-    pub server: String,
-
     /// Enable inside packet encoding once lightway connects
     /// Only used if a codec is set
     #[clap(short, long, default_value_t)]
@@ -151,5 +162,58 @@ impl Config {
                 "Either a token or username and password is required"
             )),
         }
+    }
+}
+
+#[config]
+#[derive(Parser, Debug, Serialize)]
+pub struct ConnectionConfig {
+    /// Server to connect to in <hostname>:<port> format
+    pub server: String,
+
+    /// Connection mode
+    #[serde(default)]
+    pub mode: ConnectionType,
+
+    /// Server domain name
+    #[serde(default)]
+    pub server_dn: Option<String>,
+
+    /// Cipher to use for encryption
+    #[serde(default)]
+    pub cipher: Cipher,
+}
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use super::*;
+    use clap::CommandFactory;
+    use test_case::test_case;
+    use twelf::Layer;
+
+    #[test_case("../tests/client/client_config.yaml", true, 0)]
+    #[test_case(
+        "../tests/client/parallel_connect/client_config.tcp_then_udp.yaml",
+        false,
+        2
+    )]
+    #[test_case("../tests/client/parallel_connect/client_config.tcp.yaml", false, 10)]
+    #[test_case(
+        "../tests/client/parallel_connect/client_config.udp_then_tcp.yaml",
+        false,
+        2
+    )]
+    #[test_case("../tests/client/parallel_connect/client_config.udp.yaml", false, 10)]
+    fn test_parse_config(config_file: &str, has_top_level_server: bool, servers_len: usize) {
+        let matches =
+            Config::command().get_matches_from(["lightway-client", "--config-file", config_file]);
+        let config_file = PathBuf::from_str(config_file).unwrap();
+        let config =
+            Config::with_layers(&[Layer::Yaml(config_file), Layer::Clap(matches)]).unwrap();
+
+        assert_eq!(config.server.is_empty(), !has_top_level_server);
+        assert_eq!(config.servers.len(), servers_len);
     }
 }
--- a/lightway-client/src/io/inside.rs
+++ b/lightway-client/src/io/inside.rs
@@ -15,12 +15,14 @@ use crate::ConnectionState;
 #[async_trait]
 /// Trait for InsideIORecv
 /// This will be used client app to fetch inside packets
-pub trait InsideIORecv<T: Send + Sync>: Send + Sync {
+pub trait InsideIORecv<ExtAppState: Send + Sync>: Send + Sync {
     async fn recv_buf(&self) -> IOCallbackResult<BytesMut>;
 
     fn try_send(&self, pkt: BytesMut, ip_config: Option<InsideIpConfig>) -> Result<usize>;
 
-    fn into_io_send_callback(self: Arc<Self>) -> InsideIOSendCallbackArg<ConnectionState<T>>;
+    fn into_io_send_callback(
+        self: Arc<Self>,
+    ) -> InsideIOSendCallbackArg<ConnectionState<ExtAppState>>;
 }
 
 /// Trait for InsideIO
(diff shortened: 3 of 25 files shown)


lightway-client/src/lib.rs
```diff
@@ -541,14 +545,24 @@ pub async fn encoding_request_task<T: Send + Sync>(
...
-/// 1. Disconnect all other connections
-/// 2. Forward the stop signal to the remaining connection
+/// 1. Wait for the defer connect timeout
+/// 2. Find the best connection
+/// 3. Disconnect all connections
+/// 4. Forward the stop signal to the best connection
 async fn handle_first_connection_online(
     mut connected_rx: mpsc::Receiver<usize>,
     mut stop_signal: Option<oneshot::Receiver<()>>,
     mut stop_txs: Vec<oneshot::Sender<()>>,
```
kp-thomas-yau: I think the naming could be improved, maybe something like `handle_online_connection` (can't think of one right now)

lightway-client/src/lib.rs
```diff
@@ -528,54 +543,122 @@ pub async fn encoding_request_task<T: Send + Sync>(
...
+    defer_connect_timeout: Duration,
+) -> Option<usize> {
+    let mut best_connection_index = connected_rx.recv().await?;
+
+    if !defer_connect_timeout.is_zero() {
+        tokio::time::sleep(defer_connect_timeout).await;
+        while let Ok(index) = connected_rx.try_recv() {
+            best_connection_index = best_connection_index.min(index);
+        }
+    }
```
kp-thomas-yau: Hmm I think this new logic would make total connection time exceed the configured defer timeout as there maybe cases where it take a while for the first connection to be online hmmm

kp-weiwen-goh: That's true, we need to tick the other connections in the meantime first. Let me think about how to address this.

lightway-client/src/args.rs
```diff
@@ -159,3 +171,47 @@ impl Config {
...
+    pub server_dn: Option<String>,
+
+    /// Cipher to use for encryption
+    #[serde(default)]
+    pub cipher: Cipher,
+
+    /// Enable inside packet encoding once lightway connects
+    /// Only used if a codec is set
+    #[serde(default)]
+    pub enable_inside_pkt_encoding_at_connect: bool,
```
kp-mariappan-ramasamy: Is not this a client config and not server dependent ?

lightway-client/src/lib.rs
```diff
@@ -175,6 +169,34 @@ pub struct ClientConfig<'cert, A: 'static + Send + EventCallback, T: Send + Sync
...
+    pub tls_debug: bool,
+
+    /// File path to save wireshark keylog
+    #[cfg(feature = "debug")]
+    pub keylog: Option<PathBuf>,
+}
+
+#[derive(educe::Educe)]
+#[educe(Debug)]
+pub struct ClientConnectionConfig<EventHandler: 'static + Send + EventCallback> {
```
kp-mariappan-ramasamy, line 180: Is there a reason to not move inside ClientConfig ?

lightway-client/src/args.rs
```diff
@@ -22,7 +22,7 @@ pub struct Config {
     #[serde(default)]
     pub servers: Vec<ConnectionConfig>,
 
-    /// Server to connect to in <hostname>:<port> format
+    /// Server to connect to
```
kp-mariappan-ramasamy: Is this intentional change ?

lightway-client/src/args.rs
```diff
@@ -206,7 +206,7 @@ mod tests {
     use twelf::Layer;
 
     #[test_case("../tests/client/client_config.yaml")]
-    #[test_case("../tests/client/client_config.parallel_connect.yaml")]
+    #[test_case("../tests/client/client_config.parallel_connect.tcp.yaml")]
```
kp-mariappan-ramasamy: We also have a udp.yaml which could be added here

lightway-client/src/routing_table.rs
```diff
@@ -78,6 +78,25 @@ pub enum RoutingTableError {
     ServerRouteAlreadyExists,
 }
 
+#[derive(Debug, Clone)]
+pub struct RouteConfig {
```
kp-mariappan-ramasamy: Is this unintentional change ? I could not find wher it is used.

lightway-client/src/lib.rs
```diff
@@ -832,11 +836,26 @@ pub async fn connect<EventHandler: 'static + Send + EventCallback, T: Send + Syn
 /// When the first connection is established, we need to:
```
kp-mariappan-ramasamy: Comment needs update after the defer logic

lightway-app-utils/src/args/connection_type.rs
```diff
@@ -3,13 +3,14 @@ use serde::{Deserialize, Serialize};
 
 use lightway_core::ConnectionType as LWConnectionType;
 
-#[derive(Copy, Clone, ValueEnum, Debug, Serialize, Deserialize)]
+#[derive(Copy, Clone, ValueEnum, Debug, Serialize, Deserialize, Default)]
```
kp-mariappan-ramasamy, line 6: This commit is quite big and does many atomic changes. 
Could we please split this into smaller piece like:

- Adding default to cipher and connection_type
- Moving connection related config to a struct 
- Then make it a vector of connection config

It is easier miss details in review, if the commit is big.

lightway-client/src/lib.rs
```diff
@@ -171,13 +171,13 @@ pub struct ClientConfig<'cert, T: Send + Sync> {
 
     /// Specifies if the program responds to INT/TERM signals
     #[educe(Debug(ignore))]
-    pub stop_signal: oneshot::Receiver<()>,
+    pub stop_signal: Option<oneshot::Receiver<()>>,
```
kp-mariappan-ramasamy: Could not understand why do we need Option for this signal. 
Would be better if we split this in separate commit if valid

tests/Earthfile
```diff
@@ -129,3 +138,5 @@ run-all-tests:
     BUILD +run-tcp-keepalive-test
     BUILD +run-udp-single-threaded-test
     BUILD +run-tcp-single-threaded-test
+    BUILD +run-tcp-parallel-connect-test
+    BUILD +run-udp-parallel-connect-test
```
kp-mariappan-ramasamy, line 158: We will also need a mix of both tcp and udp connections

lightway-client/src/lib.rs
```diff
@@ -739,64 +746,271 @@ pub async fn client<A: 'static + Send + EventCallback, T: Send + Sync>(
...
+}
+
+/// Returns the index of the best connection
+/// When the first connection is established, we need to:
+/// 1. Disconnect all other connections
+/// 2. Forward the stop signal to the remaining connection
+async fn find_best_connection(
+    mut connected_signals: Vec<oneshot::Receiver<()>>,
+    defer_connect_timeout: Duration,
+) -> Option<usize> {
```
kp-mariappan-ramasamy: This function could return usize. I could not find where we send None

lightway-client/src/lib.rs
kp-weiwen-goh: We want each connection to share a borrowed ConnectionConfig, but own their ClientConnectionConfig so they can take ownership of the socket and event handler.

lightway-client/src/lib.rs
```diff
@@ -739,64 +740,278 @@ pub async fn client<A: 'static + Send + EventCallback, T: Send + Sync>(
...
+    wait_timer_task.await;
+    for (i, signal) in connected_signals.iter_mut().enumerate() {
+        if let Ok(()) = signal.try_recv() {
+            preferred_connection_index = preferred_connection_index.min(i);
+            break;
+        }
+    }
+
+    Some(preferred_connection_index)
+}
```
kp-thomas-yau: Noice, is it possible to add like couple of unit test for this? seems like this can be unit tested super easily with different scenarios

lightway-client/src/main.rs
```diff
@@ -80,24 +106,25 @@ async fn main() -> Result<()> {
...
-        inside_pkt_codec: None,
-        inside_pkt_codec_config: None,
-        event_handler: Some(EventHandler),
+    let servers = if config.servers.is_empty() {
+        vec![ConnectionConfig {
+            server: config.server,
+            mode: config.mode,
+            server_dn: config.server_dn,
+            cipher: config.cipher,
+            enable_inside_pkt_encoding_at_connect: config.enable_inside_pkt_encoding_at_connect,
```
kp-mariappan-ramasamy: Could not we get rid of `ConnectionConfig::enable_inside_pkt_encoding_at_connect` ?
It is not connection specific config

lightway-client/src/main.rs
```diff
@@ -26,6 +28,35 @@ impl EventCallback for EventHandler {
     }
 }
 
+/// Creates a ClientConnectionConfig from a ConnectionConfig. Resolving the
+/// hostname blocks, so it should be done in a separate thread if needed
```
kp-mariappan-ramasamy: Good information.
We could also add a info trace mentioning Server address is resolving here.

And tokio has a async resolver:
https://docs.rs/tokio/latest/tokio/net/trait.ToSocketAddrs.html

We could use it instead of thread.

lightway-client/src/lib.rs
kp-weiwen-goh: Good idea

kp-thomas-yau: hmmm, why are we breaking when we check connections in order in the for-loop? I think this will stop at the first connected connections

kp-weiwen-goh: We loop in order of priority, so if we find any connected connections we know that nothing else can be the preferred connection.

lightway-client/src/lib.rs
```diff
@@ -823,14 +826,34 @@ pub async fn connect<EventHandler: 'static + Send + EventCallback, T: Send + Syn
...
+    let (_, mut preferred_connection_index, _) = select_all(&mut connected_signals).await;
     tracing::trace!("First connection is online: {preferred_connection_index}");
 
-    Some(preferred_connection_index)
+    // If the highest priority connection is already online, we don't need to wait
+    if preferred_connection_index == 0 {
+        return preferred_connection_index;
+    }
+
+    wait_timer_task.await;
```
kp-mariappan-ramasamy: The timer task has to be started even before the connection starts. 
We did not create task at line 837 and just created the structure. So we are starting the task only after the first connection is online.

lightway-client/src/args.rs
```diff
@@ -99,6 +99,10 @@ pub struct Config {
     #[clap(long, default_value = "0s")]
     pub keepalive_timeout: Duration,
 
+    /// How long to wait before selecting the preferred connection
```
kp-mariappan-ramasamy: As discussed, we could use the `best` connection instead of `preferred` connection

lightway-client/src/lib.rs
kp-thomas-yau: oh yeah i missed that thanks

lightway-client/src/lib.rs
```diff
@@ -528,54 +540,83 @@ pub async fn encoding_request_task<T: Send + Sync>(
...
+        server_config,
+        inside_io,
+    )
+)]
+pub async fn connect<EventHandler: 'static + Send + EventCallback, T: Send + Sync>(
+    config: &ClientConfig<'_, T>,
+    mut server_config: ClientConnectionConfig<EventHandler>,
+    inside_io: Arc<dyn io::inside::InsideIO<T>>,
+) -> Result<ClientConnection<T>> {
+    tracing::info!("Connection starting with config:\n{:#?}", &server_config);
```
kp-mariappan-ramasamy: We also print the servers config here:
https://github.com/expressvpn/lightway/pull/239/files#diff-76cc785c1d261df977f03e710141e50595aa4dbe21f36ea29edccc143ff96c3cR917

I think we can either remove this line or print only the server addrss instead of whole config.

lightway-client/src/main.rs
kp-weiwen-goh: Updated to use `tokio::net::lookup_host` and resolve in parallel.

lightway-client/src/main.rs
```diff
@@ -150,7 +150,8 @@ async fn main() -> Result<()> {
         iouring_entry_count: config.iouring_entry_count,
         #[cfg(feature = "io-uring")]
         iouring_sqpoll_idle_time: config.iouring_sqpoll_idle_time.into(),
-        inside_pkt_codec_config: None,
+        enable_inside_pkt_encoding_at_connect: config.enable_inside_pkt_encoding_at_connect,
+        encoding_request_signal: None,
```
kp-mariappan-ramasamy: Previously we intentionally merge this config to a struct, to avoid applications specifying lot of parameters if they are not using inside pkt codec.

What will be problem with it being inside a struct ?

The commit message says :
```
client: split inside_pkt_codec config
```
Please try to  add reason in commit message, on why it is done too

lightway-client/src/args.rs
```diff
@@ -93,6 +93,11 @@ pub struct Config {
     #[clap(long, default_value = "0s")]
     pub keepalive_timeout: Duration,
 
+    /// How long to wait before selecting the best connection. If the preferred
+    /// connection connects before the timeout, it will be used immediately.
+    #[clap(long, default_value = "0s")]
+    pub preferred_connection_wait_interval: Duration,
```
kp-mariappan-ramasamy, line 99: This change is not needed in this commit `client: Attempt to connect to multiple servers at once`

lightway-client/src/main.rs
kp-weiwen-goh: I initially thought `encoding_request_signal` is dependent on each connection, but looking back they are all independent of the connection settings. I will revert this change.

lightway-client/src/io/inside.rs
```diff
@@ -15,12 +15,12 @@ use crate::ConnectionState;
 #[async_trait]
 /// Trait for InsideIORecv
 /// This will be used client app to fetch inside packets
-pub trait InsideIORecv: Sync + Send {
+pub trait InsideIORecv<T: Send + Sync>: Send + Sync {
```
kp-mariappan-ramasamy: Might be better if we use `ExtAppState` as generic instead of T. Easier later.
What do you think ?

lightway-client/src/lib.rs
```diff
@@ -872,6 +919,9 @@ fn apply_platform_specific_config(
 }
 
 /// Launches connections concurrently and waits for the first one to complete.
+/// If `config.defer_connect_timeout`` is set, it will wait that duration after
```
kp-mariappan-ramasamy: This text `config.defer_connect_timeout` is different from actual variable `config.preferred_connection_wait_interval`

lightway-client/src/io/inside.rs
kp-weiwen-goh: Yes, I couldn't think of a good name for it. `ExtAppState` is good

kp-mariappan-ramasamy (approved):
LGTM

PR #240, opened 2025-07-28
CVPN-2123 Static Routing to Socket IP instead of Server IP

xv-thomas-leong:
Static routing to the server IP works only when the socket connects directly. In cases where connections are relayed, this approach fails. The correct IP to exclude from the VPN route should instead be derived from the OutsideIO struct, which reflects the actual remote endpoint.

## How Has This Been Tested?

CI tested.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -626,7 +626,7 @@ pub async fn client<A: 'static + Send + EventCallback, T: Send + Sync>(
     ))]
     route_table
         .initialize_routing_table(
-            &config.server.ip(),
+            &outside_io.peer_addr().ip(),
             tun_index,
             &config.tun_peer_ip.into(),
             &config.tun_dns_ip.into(),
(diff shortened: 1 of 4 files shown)


lightway-client/src/lib.rs
```diff
@@ -626,7 +626,7 @@ pub async fn client<A: 'static + Send + EventCallback, T: Send + Sync>(
     ))]
     route_table
         .initialize_routing_table(
```
kp-brandon-chew, line 628: Perhaps we might want to consider refactoring the input name of the `.initialize_routing_table()` method implementation, to better reflect this change, it's currently left as `server_ip: &IpAddr`

xv-thomas-leong: Good idea, let me update that 👍

xv-thomas-leong: On second thought, I think it’s better to leave this unchanged - let me know if you agree.

From the RoutingTable's perspective, nothing has fundamentally changed. It still needs to create a route to the server (or more precisely, the concept of the server) that is excluded from VPN routing. The only difference is that the parameter being passed to the server has changed; it now reflects the source of truth for the server’s location.

kp-brandon-chew: Yup that makes sense

PR #245, opened 2025-08-05
Fix not ignoring route exists error

kp-brandon-chew:
## Types of changes

## Checklist:


lightway-client/src/routing_table.rs
```diff
@@ -135,14 +135,36 @@ impl RoutingTable {
...
+    fn is_route_exists_error(&self, error: &std::io::Error) -> bool {
+        match error.raw_os_error() {
+            #[cfg(any(
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "freebsd",
+                target_os = "openbsd",
+                target_os = "netbsd"
+            ))]
+            Some(17) => true, // EEXIST
```
kp-mariappan-ramasamy: Please don't use magic number and try to use libc constants:
https://github.com/expressvpn/lightway/blob/2b33521ae275c858be2b018dfe67d308f6de4eae/lightway-app-utils/src/iouring.rs#L185

kp-brandon-chew: Fixed

lightway-client/src/routing_table.rs
```diff
@@ -135,14 +135,36 @@ impl RoutingTable {
...
+    }
+
+    fn is_route_exists_error(&self, error: &std::io::Error) -> bool {
+        match error.raw_os_error() {
+            #[cfg(any(
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "freebsd",
+                target_os = "openbsd",
+                target_os = "netbsd"
```
kp-mariappan-ramasamy, line 161: In general, we do not want to have cfg for unsupported os'es, since it will give impression that we support those Os's too. But it looks like there are few chunk of code already using these cfgs. 

Let's merge this as is and clean this later

kp-mariappan-ramasamy (approved):

PR #254, opened 2025-08-12
client: Added flag for dns_config options

kp-brandon-chew:
This PR aims to provide users the option to setup the DNS configuration.

## Description

We provide a `dns_config_mode` flag which users can set to `Default` or `NoExec` 

## Motivation and Context

Grants users more power to customize runtime behaviour. On some platforms, not all runs will require DNS configuration.

## How Has This Been Tested?

unittests
Running the VPN in both modes

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/dns_manager.rs
+++ b/lightway-client/src/dns_manager.rs
@@ -1,5 +1,15 @@
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::warn;
+
+/// DNS configuration mode
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum DnsConfigMode {
+    #[default]
+    Default,
+    NoExec,
+}
 #[derive(Error, Debug)]
 pub enum DnsManagerError {
     #[error("Unable to find primary service ID")]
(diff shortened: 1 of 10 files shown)


lightway-client/src/dns_manager.rs
```diff
@@ -1,5 +1,14 @@
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::warn;
+
+/// DNS configuration mode
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum, Serialize, Deserialize)]
```
kp-weiwen-goh: You can derive Default here so you don't have to specify the default mode in clap.

lightway-client/src/lib.rs
```diff
@@ -1020,8 +1023,14 @@ pub async fn client<
         )
         .await?;
 
-    let mut dns_manager = DnsManager::default();
-    dns_manager.set_dns(&config.tun_dns_ip.to_string())?;
+    #[cfg(any(target_os = "linux", target_os = "macos",))]
+    let _dns_manager = if config.dns_config_mode == DnsConfigMode::Default {
```
kp-weiwen-goh: Might be good to explicitly store it in the ClientConnection like we do for the route table so we don't accidentally drop this in the future.

kp-weiwen-goh (dismissed):
LGTM, some nits

lightway-client/src/lib.rs
kp-brandon-chew: Agreed, fixed

lightway-client/src/dns_manager.rs
kp-brandon-chew: Implemented

PR #257, opened 2025-08-13
CVPN-2114: Migration tun to tun-rs

kp-brandon-chew:
This PR moves away from using rust-tun and replaces it with tun-rs

## Description

`config.enable_routing(false)` was previously used as otherwise the tun crate would add this route on macos
```
Destination        Gateway            Flags               Netif Expire
100.64/24          100.64.0.5         UGSc             utun1995       
```
the tun-rs crate does not exhibit this behaviour and so such an option is not implemented by `TunConfig`. However,
tun-rs crate does seem to add this route on macos

```
Destination        Gateway            Flags               Netif Expire
100.64.0.6         link#23            UHS              utun1995       
```
which is not a problem as long as we set `device.associate_route(false)` on macos

## Motivation and Context

Among other reasons, tun-rs exposes methods with `&self` rather than `&mut self` e.g. in `set_mtu()` which allows allows us to use these methods through `Arc` in `InsideIO` without having to incur the cost of a `Mutex`

## How Has This Been Tested?

ran client to connect to production servers

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,5 +59,5 @@ tokio-stream = "0.1.14"
 tokio-util = "0.7.10"
 tracing = "0.1.37"
 tracing-subscriber = "0.3.17"
-tun = { version = "0.8", features = ["async"] }
+tun-rs = { version = "2.5.1", features = ["async"] }
 twelf = { version = "0.15.0", default-features = false, features = ["env", "clap", "yaml"]}
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -5,19 +5,119 @@ use lightway_core::IOCallbackResult;
 #[cfg(feature = "io-uring")]
 use std::time::Duration;
 use std::{
-    ops::Deref,
-    os::fd::{AsRawFd, RawFd},
+    net::{IpAddr, Ipv4Addr},
+    os::fd::{AsRawFd, IntoRawFd, RawFd},
 };
-
-use tun::{AbstractDevice, AsyncDevice as TokioTun};
-
-pub use tun::Configuration as TunConfig;
 
 #[cfg(feature = "io-uring")]
 use std::sync::Arc;
+use tun_rs::AsyncDevice;
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+use tun_rs::DeviceBuilder;
 
 #[cfg(feature = "io-uring")]
 use crate::IOUring;
+
+/// Configuration options for creating a interface
+///
+/// This struct provides a builder-like interface for configuring TUN interfaces
+/// with various network settings including address assignment, routing, and MTU.
+#[derive(Clone, Default, Debug)]
+pub struct TunConfig {
+    /// Optional name for the TUN interface (e.g., "utun3" on macOS)
+    pub tun_name: Option<String>,
+    /// IP address to assign to the TUN interface (IPv4 or IPv6)
+    pub address: Option<IpAddr>,
+    /// Destination/gateway address for the TUN interface
+    pub destination: Option<Ipv4Addr>,
+    /// Network mask for the assigned address (defaults to host route if not specified)
+    pub prefix: Option<u8>,
+    /// Maximum transmission unit size in bytes
+    pub mtu: Option<u16>,
+    /// Whether the interface should be brought up after creation
+    pub enabled: bool,
+}
+
+impl TunConfig {
+    /// Set the tun name.
+    /// # Note:
+    /// On macOS, the tun name must be the form `utunx` where `x` is a number, such as `utun3`
+    pub fn tun_name<T: Into<String>>(&mut self, tun_name: T) -> &mut Self {
+        self.tun_name = Some(tun_name.into());
+        self
+    }
+
+    /// Set the gateway address.
+    pub fn address(&mut self, value: IpAddr) -> &mut Self {
+        self.address = Some(value);
+        self
+    }
+
+    /// Set the destination address.
+    pub fn destination(&mut self, value: Ipv4Addr) -> &mut Self {
+        self.destination = Some(value);
+        self
+    }
+
+    /// Set the netmask for address
+    pub fn prefix(&mut self, prefix: u8) -> &mut Self {
+        self.prefix = Some(prefix);
+        self
+    }
+
+    /// Set the MTU.
+    pub fn mtu(&mut self, value: u16) -> &mut Self {
+        self.mtu = Some(value);
+        self
+    }
+
+    /// Set the interface to be enabled once created.
+    pub fn up(&mut self) -> &mut Self {
+        self.enabled = true;
+        self
+    }
+    /// Creates an async device based on TunConfig
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    pub fn create_as_async(&self) -> std::io::Result<AsyncDevice> {
+        let mut builder = DeviceBuilder::new();
+        if let Some(name) = self.tun_name.as_ref() {
+            builder = builder.name(name);
+        }
+        #[cfg(target_os = "macos")]
+        {
+            builder = builder.associate_route(false);
+        }
+        let device = builder.build_async()?;
+
+        if let Some(mtu) = self.mtu {
+            device.set_mtu(mtu)?;
+        }
+
+        device.enabled(self.enabled)?;
+
+        if let Some(address) = self.address {
+            match address {
+                IpAddr::V4(ipv4_addr) => {
+                    let netmask = self
+                        .prefix
+                        .map(|x| x.min(Ipv4Addr::BITS as u8))
+                        .unwrap_or(Ipv4Addr::BITS as u8);
+                    device.set_network_address(ipv4_addr, netmask, self.destination)?;
+                }
+                IpAddr::V6(ipv6_addr) => {
+                    use std::net::Ipv6Addr;
+
+                    let netmask = self
+                        .prefix
+                        .map(|x| x.min(Ipv6Addr::BITS as u8))
+                        .unwrap_or(Ipv6Addr::BITS as u8);
+                    device.add_address_v6(ipv6_addr, netmask)?;
+                }
+            }
+        }
+        Ok(device)
+    }
+}
 
 /// Tun enum interface to read/write packets
 pub enum Tun {
@@ -95,7 +195,7 @@ impl AsRawFd for Tun {
 
 /// Tun struct
 pub struct TunDirect {
-    tun: TokioTun,
+    tun: AsyncDevice,
     mtu: u16,
     fd: RawFd,
 }
@@ -103,10 +203,13 @@ pub struct TunDirect {
 impl TunDirect {
     /// Create a new `Tun` struct
     pub fn new(config: &TunConfig) -> Result<Self> {
-        let tun = tun::create_as_async(config)?;
-        let fd = tun.as_raw_fd();
-        let mtu = tun.mtu()?;
-
+        let tun_device = config.create_as_async()?;
+        let fd = tun_device.as_raw_fd();
+        #[cfg(any(target_os = "linux", target_os = "macos"))]
+        let mtu = tun_device.mtu()?;
+        // This currently is not supported for Android and IOS
+        #[cfg(not(any(target_os = "linux", target_os = "macos")))]
+        let mtu = 1350;
         Ok(TunDirect { tun, mtu, fd })
     }
 
@@ -130,8 +233,7 @@ impl TunDirect {
 
     /// Try write from Tun
     pub fn try_send(&self, buf: BytesMut) -> IOCallbackResult<usize> {
-        let try_send_res = self.tun.deref().send(&buf[..]);
-        match try_send_res {
+        match self.tun.try_send(&buf[..]) {
             Ok(nr) => IOCallbackResult::Ok(nr),
             Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
                 IOCallbackResult::WouldBlock
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use bytes::BytesMut;
+use educe::Educe;
 use lightway_core::IOCallbackResult;
 
 #[cfg(feature = "io-uring")]
@@ -9,6 +10,8 @@ use std::{
     os::fd::{AsRawFd, IntoRawFd, RawFd},
 };
 
+#[cfg(not(any(target_os = "linux", target_os = "macos")))]
+use std::os::fd::FromRawFd;
 #[cfg(feature = "io-uring")]
 use std::sync::Arc;
 use tun_rs::AsyncDevice;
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -60,4 +60,5 @@ tokio-util = "0.7.10"
 tracing = "0.1.37"
 tracing-subscriber = "0.3.17"
 tun = { version = "0.8", features = ["async"] }
```
kp-mariappan-ramasamy: We should remove this tun crate when its usage is removed

lightway-app-utils/src/tun.rs
```diff
@@ -95,18 +196,20 @@ impl AsRawFd for Tun {
...
 
 impl TunDirect {
     /// Create a new `Tun` struct
-    pub fn new(config: &TunConfig) -> Result<Self> {
-        let tun = tun::create_as_async(config)?;
-        let fd = tun.as_raw_fd();
-        let mtu = tun.mtu()?;
-
+    pub fn new(config: TunConfig) -> Result<Self> {
+        let tun_device = config.create_as_async()?;
```
kp-weiwen-goh, line 248: I can't find where we need TunConfig to be owned. `create_as_async` takes in a borrowed config.

lightway-app-utils/src/tun.rs
kp-mariappan-ramasamy, line 1: We generally use the crate name as prefix for commit message.
Ex: client, lightway-client, server, lightway-core etc

For this commit `Modified the if_index() trait api type from i32 to u32` the change originates from changing the type of core::io::InsideIOSendCallback, and the other changes are dependents of that.
So we might use `core` as prefix for commit message.

And please try to be clear and detailed in commit message

lightway-app-utils/src/tun.rs
```diff
@@ -5,20 +5,121 @@ use lightway_core::IOCallbackResult;
...
+    pub destination: Option<Ipv4Addr>,
+    /// Network mask for the assigned address (defaults to host route if not specified)
+    pub prefix: Option<u8>,
+    /// Maximum transmission unit size in bytes
+    pub mtu: Option<u16>,
+    /// Whether the interface should be brought up after creation
+    pub enabled: bool,
+}
+const MAX_PREFIX_LEN_IPV4: u8 = 32;
+const MAX_PREFIX_LEN_IPV6: u8 = 128;
```
kp-mariappan-ramasamy: We could use the values from rust std library instead of hardcoding.
https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.BITS

Use hard coding only when it is not a standard thing and not available in libraries

lightway-app-utils/src/tun.rs
```diff
@@ -5,20 +5,121 @@ use lightway_core::IOCallbackResult;
...
+        self
+    }
+
+    /// Set the interface to be enabled once created.
+    pub fn up(&mut self) -> &mut Self {
+        self.enabled = true;
+        self
+    }
+    /// Creates an async device based on TunConfig
+    #[allow(unsafe_code)]
```
kp-mariappan-ramasamy: I do not see any unsafe code here.
Its not clear, why do we need to allow unsafe code here ?

lightway-app-utils/src/tun.rs
```diff
@@ -170,7 +272,7 @@ pub struct TunIoUring {
 impl TunIoUring {
     /// Create `TunIoUring` struct
     pub async fn new(
-        config: &TunConfig,
+        config: TunConfig,
```
kp-mariappan-ramasamy: Is it not possible to use Reference here ?
Why did we change it ?

Please try to be clear in the commit message and the reason for any specific change made in a commit

lightway-app-utils/src/tun.rs
```diff
@@ -36,6 +39,10 @@ pub struct TunConfig {
     pub mtu: Option<u16>,
     /// Whether the interface should be brought up after creation
     pub enabled: bool,
+    /// File Descriptor of the Tunnel
+    pub fd: Option<RawFd>,
+    /// Whether to keep and prevent closing the file descriptor when the TUN device is dropped
+    pub keep_fd_on_drop: bool,
```
kp-weiwen-goh: Maybe something like `close_fd_on_drop` is a better name

lightway-app-utils/src/tun.rs
```diff
@@ -118,6 +139,19 @@ impl TunConfig {
         }
         Ok(device)
     }
+    /// Creates an async device based on TunConfig
+    #[allow(unsafe_code)]
+    #[cfg(not(any(target_os = "linux", target_os = "macos")))]
+    pub fn create_as_async(&self) -> std::io::Result<AsyncDevice> {
+        let device = match self.fd {
+            Some(fd) => unsafe { tun_rs::AsyncDevice::from_raw_fd(fd) },
```
kp-weiwen-goh: Add safety comment

lightway-app-utils/src/tun.rs
kp-brandon-chew: it is, but because default for `bool` is `false` I inverted it to use `keep_fd_on_drop`. Didnt seem like there was a macro-attribute that could modify it to true. or is something like `do_not_close_fd_on_drop` better?

kp-mariappan-ramasamy: I think We could use educe to have a specific default value.
https://docs.rs/educe/latest/educe/#the-default-values-for-specific-fields

We already use educe for Debug in client and server code

kp-brandon-chew: nice

Cargo.toml
kp-brandon-chew: removed

lightway-app-utils/src/tun.rs
kp-brandon-chew: added

lightway-app-utils/src/tun.rs
kp-brandon-chew: reverted

lightway-app-utils/src/tun.rs
kp-brandon-chew: deleted

lightway-app-utils/src/tun.rs
kp-brandon-chew: fixed

lightway-app-utils/src/tun.rs
kp-brandon-chew: reverted

lightway-app-utils/src/tun.rs
kp-brandon-chew: added

lightway-app-utils/src/tun.rs
kp-brandon-chew: added

lightway-client/src/main.rs
```diff
@@ -87,7 +87,7 @@ async fn main() -> Result<()> {
     // TODO: Fix in future PR
     tun_config
         .mtu(1350)
-        .address(config.tun_local_ip)
+        .address(std::net::IpAddr::V4(config.tun_local_ip))
```
kp-mariappan-ramasamy: nit: Might have used `.address(config.tun_local_ip.into())`

lightway-app-utils/src/tun.rs
```diff
@@ -95,18 +195,20 @@ impl AsRawFd for Tun {
...
     pub fn new(config: &TunConfig) -> Result<Self> {
-        let tun = tun::create_as_async(config)?;
-        let fd = tun.as_raw_fd();
-        let mtu = tun.mtu()?;
-
+        let tun_device = config.create_as_async()?;
+        let fd = tun_device.as_raw_fd();
+        #[cfg(any(target_os = "linux", target_os = "macos"))]
+        let mtu = tun_device.mtu()?;
+        #[cfg(not(any(target_os = "linux", target_os = "macos")))]
```
kp-mariappan-ramasamy, line 253: In which platforms, is this not working. please mention in comments for our future reference

lightway-server/src/main.rs
```diff
@@ -144,6 +144,7 @@ async fn main() -> Result<()> {
     if let Some(tun_name) = config.tun_name {
         tun_config.tun_name(tun_name);
     }
+    tun_config.up();
```
kp-mariappan-ramasamy, line 147: Please update the commit message to be clear:

```
Set tun to up in server as default behaviour
overrights tun routes if they have been set manually
before running server binary (fails e2e test)
```

Is it a single paragraph or different bullet points. 
I could not understand what do we mean by commit messsage last line

lightway-app-utils/src/tun.rs
```diff
@@ -261,6 +311,34 @@ impl AsRawFd for TunDirect {
...
+            // ManuallyDrop::drop is safe here because we never access self.tun again after this call.
+            unsafe {
+                ManuallyDrop::drop(&mut self.tun);
+            }
+        } else {
+            // Take ownership and call into_raw_fd (prevents fd closure)
+            // SAFETY: This is the final drop of TunDirect, so we have exclusive access to self.tun.
+            // ManuallyDrop::take is safe here because we immediately consume the value and never access self.tun again.
+            let tun = unsafe { ManuallyDrop::take(&mut self.tun) };
+            let _ = tun.into_raw_fd();
```
kp-mariappan-ramasamy, line 331: The main essense of this drop implementaion is this line it seems.

And this could be achieved by using Option<Tun> without any unsafe code. Is not it ? 
Or is it intentional to use ManuallyDrop. Please mention clearly in the commit message

<details open>
<summary>Changes</summary>

```diff
diff --git a/lightway-app-utils/src/tun.rs b/lightway-app-utils/src/tun.rs
index e6f48ee..337ff8f 100644
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -6,7 +6,6 @@ use lightway_core::IOCallbackResult;
 #[cfg(feature = "io-uring")]
 use std::time::Duration;
 use std::{
-    mem::ManuallyDrop,
     net::{IpAddr, Ipv4Addr},
     os::fd::{AsRawFd, IntoRawFd, RawFd},
 };
@@ -237,7 +236,7 @@ impl AsRawFd for Tun {

 /// Tun struct
 pub struct TunDirect {
-    tun: ManuallyDrop<AsyncDevice>,
+    tun: Option<AsyncDevice>,
     mtu: u16,
     fd: RawFd,
     close_fd_on_drop: bool,
@@ -252,7 +251,7 @@ impl TunDirect {
         let mtu = tun_device.mtu()?;
         #[cfg(not(any(target_os = "linux", target_os = "macos")))]
         let mtu = 1350;
-        let tun = ManuallyDrop::new(tun_device);
+        let tun = Some(tun_device);

         Ok(TunDirect {
             tun,
@@ -264,8 +263,9 @@ impl TunDirect {

     /// Recv from Tun
     pub async fn recv_buf(&self) -> IOCallbackResult<bytes::BytesMut> {
+        let tun = self.tun.as_ref().unwrap();
         let mut buf = BytesMut::zeroed(self.mtu as usize);
-        match self.tun.recv(buf.as_mut()).await {
+        match tun.recv(buf.as_mut()).await {
             // TODO: Check whether we can use poll
             // Getting spurious reads
             Ok(0) => IOCallbackResult::WouldBlock,
@@ -282,7 +282,8 @@ impl TunDirect {

     /// Try write from Tun
     pub fn try_send(&self, buf: BytesMut) -> IOCallbackResult<usize> {
-        match self.tun.try_send(&buf[..]) {
+        let tun = self.tun.as_ref().unwrap();
+        match tun.try_send(&buf[..]) {
             Ok(nr) => IOCallbackResult::Ok(nr),
             Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
                 IOCallbackResult::WouldBlock
@@ -298,8 +299,9 @@ impl TunDirect {

     /// Interface index of Tun
     pub fn if_index(&self) -> std::io::Result<u32> {
+        let tun = self.tun.as_ref().unwrap();
         #[cfg(any(target_os = "linux", target_os = "macos"))]
-        return self.tun.if_index();
+        return tun.if_index();
         #[cfg(not(any(target_os = "linux", target_os = "macos")))]
         return Err(std::io::Error::from(std::io::ErrorKind::Unsupported));
     }
@@ -322,18 +324,8 @@ impl IntoRawFd for TunDirect {
 impl Drop for TunDirect {
     #[allow(unsafe_code)]
     fn drop(&mut self) {
-        if self.close_fd_on_drop {
-            // Manually drop the AsyncDevice (closes the fd)
-            // SAFETY: This is the final drop of TunDirect, so we have exclusive access to self.tun.
-            // ManuallyDrop::drop is safe here because we never access self.tun again after this call.
-            unsafe {
-                ManuallyDrop::drop(&mut self.tun);
-            }
-        } else {
-            // Take ownership and call into_raw_fd (prevents fd closure)
-            // SAFETY: This is the final drop of TunDirect, so we have exclusive access to self.tun.
-            // ManuallyDrop::take is safe here because we immediately consume the value and never access self.tun again.
-            let tun = unsafe { ManuallyDrop::take(&mut self.tun) };
+        if !self.close_fd_on_drop {
+            let tun = self.tun.take().unwrap();
             let _ = tun.into_raw_fd();
         }
     }
```
</details>

kp-brandon-chew: We can use an option, which would remove the unsafe on dropping.

However, `Option` does not implement `Deref` and is an enum. So we would be introducing a branch whenever we try to do `self.tun.recv()` as the new implementation would be `self.tun.unwrap().recv()` (same goes for `try_send()`), as `unwrap()` is implemented via pattern matching (https://doc.rust-lang.org/src/core/option.rs.html#1002).  We could avoid this with `unwrap_unchecked()` (https://doc.rust-lang.org/src/core/option.rs.html#1110) but it also involves `unsafe` blocks.

`ManuallyDrop` creates no additional branches as it has `Deref` implemented. 

Realistically though, I think the branch predictor will get this right most of the time, as the `Option` will be `Some` until the end of its lifetime.

Let me know which is preferable.

kp-brandon-chew: ok wait i realise its not just `unwrap()` but with `as_ref()`
https://doc.rust-lang.org/src/core/option.rs.html#733

but still theres a branch...

kp-mariappan-ramasamy: I would suggest to use Safe code whenever possible and add optimizations only when benchmarked or found it is a bottleneck

kp-brandon-chew: changed

lightway-server/src/main.rs
kp-brandon-chew: modified

lightway-app-utils/src/tun.rs
kp-brandon-chew: added comment

lightway-client/src/main.rs
kp-brandon-chew: changed

lightway-app-utils/src/tun.rs
```diff
@@ -262,6 +314,24 @@ impl AsRawFd for TunDirect {
...
+impl IntoRawFd for TunDirect {
+    fn into_raw_fd(mut self) -> RawFd {
+        // Alters state to prevent drop from closing fd
+        self.close_fd_on_drop = false;
+        self.fd
+    }
+}
+
+impl Drop for TunDirect {
+    #[allow(unsafe_code)]
```
kp-mariappan-ramasamy: We do not need unsafe code, do we ?

kp-brandon-chew: fixed

PR #263, opened 2025-08-20
client: graceful disconnect

kp-weiwen-goh:
## Description

During an intended disconnect, ensure we wait for the Disconnect state before we abort tasks. This allows the graceful disconnect to take place.

## Motivation and Context
The existing implementation does not wait for the Lightway connection to fully disconnect before aborting tasks. This causes the connection to not go through the usual disconnect state changes.

This is fixed by waiting for the `Disconnect` state if the client requests a disconnect before the client can exit.

## How Has This Been Tested?

Manually test and check that all state transitions are correctly logged.

## Types of changes

## Checklist:


kp-mariappan-ramasamy (approved):
LGTM

PR #267, opened 2025-08-27
CVPN-2221: Add Windows support

kp-mariappan-ramasamy:
## Description

Add windows support for lightway client
The major hurdle was the socket options which was unix specific. 

I tried to unify IP_MTU_DISCOVER by adding some abstraction.
But it become complex that the resulting code loses readabliity. 

Also I found a bug in MacOs where we sends Pmtud::Probe => 3 instead of 1 since it does not support IP_MTU_DISCOVER and only support IP_DONTFRAG
To avoid these issues, I reverted to having multiple implementations based on platform.

Also, IP_PKTINFO socket option is needed for server and since we are not focusing on server, compiled it out on windows

## Motivation and Context

Add support for windows platform

## How Has This Been Tested?
Verified windows client can connect to prod servers and transfer data traffic

Pending:
- DNS manager is currently stubbed. Have to fix this in future PRs
- Also tun-rs crate creates a default route for Tun apart from 0/1 and 128/1 routes. Since we are moving to tun-rs, we will revisit after the migration

## Types of changes

## Checklist:

- [ ] Wait until https://github.com/expressvpn/wolfssl-rs/pull/286 is merged and rebase to wolfssl main


kp-weiwen-goh (approved):
LGTM

PR #268, opened 2025-08-28
CVPN-2218: Trigger keepalive (tracer packet) for connection liveness check

kp-thomas-yau:
## Description

Add ability to detect when server has gone away silently by tracking time since last outside data received. When threshold exceeded during inside packet processing, trigger keepalive (tracer packet) to verify server connectivity. The tracer packet would be disabled when we're using a continuous keepalive already.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -29,6 +29,7 @@ use tokio::sync::mpsc::UnboundedReceiver;
 use crate::debug::WiresharkKeyLogger;
 #[cfg(any(target_os = "linux", target_os = "macos",))]
 use crate::dns_manager::{DnsConfigMode, DnsManager, DnsManagerError, DnsSetup};
+use crate::keepalive::Config as KeepaliveConfig;
 #[cfg(any(target_os = "linux", target_os = "macos",))]
 use crate::routing_table::{RouteMode, RoutingTable};
 #[cfg(feature = "debug")]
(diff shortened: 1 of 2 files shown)


lightway-client/src/lib.rs
```diff
@@ -380,38 +393,60 @@ pub async fn inside_io_task<ExtAppState: Send + Sync>(
...
+
+            match conn.inside_data_received(&mut buf) {
+                Ok(()) => {
+                    let now = Instant::now();
+                    now.duration_since(conn.activity().last_outside_data_received)
+                }
+                Err(ConnectionError::PluginDropWithReply(reply)) => {
+                    // Send the reply packet to inside path
+                    let _ = inside_io.try_send(reply, ip_config);
+                    return Ok(());
```
kp-mariappan-ramasamy: Why are we returning here ? 
It will return the inside_io and close the client connection.

lightway-client/src/lib.rs
```diff
@@ -380,38 +393,60 @@ pub async fn inside_io_task<ExtAppState: Send + Sync>(
...
+                }
+                Err(ConnectionError::InvalidState) => {
+                    // Ignore the packet till the connection is online
+                    return Ok(());
```
kp-mariappan-ramasamy: Same as above

lightway-client/src/lib.rs
```diff
@@ -380,38 +393,60 @@ pub async fn inside_io_task<ExtAppState: Send + Sync>(
...
+                }
+                Err(ConnectionError::InvalidInsidePacket(_)) => {
+                    // Ignore invalid inside packet
+                    return Ok(());
```
kp-mariappan-ramasamy: Same as above

kp-mariappan-ramasamy (approved):
LGTM

Please ignore my earlier comment. I overlooked the code and confused.

PR #269, opened 2025-08-29
core: Allow EventCallback to be mutable

kp-weiwen-goh:
## Description

In all the contexts where it is used, `EventCallback` is already mutable. We expose this mutability to it so it can maintain its own state.

## Motivation and Context

This allows an `EventCallback` implementation to be stateful without interior mutability.

## How Has This Been Tested?

Tests should pass

## Types of changes

## Checklist:


kp-mariappan-ramasamy (approved):
LGTM

PR #281, opened 2025-09-15
Move .cargo inside earthly, refactor earthly targets

kp-samuel-tam:
## Types of changes

## Checklist:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,0 @@
-[target.aarch64-unknown-linux-gnu]
-linker = "aarch64-linux-gnu-gcc"
-runner = ["qemu-aarch64-static"] # use qemu user emulation for cargo run and test
-
-[target.riscv64gc-unknown-linux-gnu]
-runner = "qemu-riscv64-static -L /usr/riscv64-linux-gnu -cpu rv64"
-linker = "riscv64-linux-gnu-gcc"
--- a/Earthfile
+++ b/Earthfile
@@ -43,64 +43,56 @@ source:
     FROM +install-build-dependencies
     COPY --keep-ts Cargo.toml Cargo.lock Makefile.toml ./
     COPY --keep-ts deny.toml ./
-    COPY --keep-ts --dir lightway-core lightway-app-utils lightway-client lightway-server tests .cargo ./
+    COPY --keep-ts --dir lightway-core lightway-app-utils lightway-client lightway-server tests ./
 
-# build builds with the Cargo release profile
+# build runs cargo to build native binaries for the host platform.
+# You may use `--platform linux/[amd64|arm64]` to override the host platform, to natively compile in emulation.
 build:
     FROM +source
-    ARG ARCH=$(dpkg --print-architecture)
-    LET target = "x86_64-unknown-linux-gnu"
 
-    IF [ "$ARCH" = "arm64" ]
-        SET target = "aarch64-unknown-linux-gnu"
-    ELSE IF [ "$ARCH" = "riscv64" ]
-        SET target = "riscv64gc-unknown-linux-gnu"
-    END
+    DO lib-rust+CARGO --args="build --release --features io-uring" --output="release/lightway-(client|server)$"
+
+    SAVE ARTIFACT ./target/release/lightway-client AS LOCAL ./target/release/
+    SAVE ARTIFACT ./target/release/lightway-server AS LOCAL ./target/release/
+
+# build-cross-arm64 cross-compiles to arm64 from an amd64 host.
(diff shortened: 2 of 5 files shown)


xv-thomas-leong (commented):
Changes looks okay, but can you write in the commit message why you're making this change? It's important we document the motivation and the chain of thoughts that lead to the changes you made.

.cargo/config.toml
```diff
@@ -1,7 +0,0 @@
-[target.aarch64-unknown-linux-gnu]
-linker = "aarch64-linux-gnu-gcc"
-runner = ["qemu-aarch64-static"] # use qemu user emulation for cargo run and test
-
-[target.riscv64gc-unknown-linux-gnu]
-runner = "qemu-riscv64-static -L /usr/riscv64-linux-gnu -cpu rv64"
-linker = "riscv64-linux-gnu-gcc"
```
kp-mariappan-ramasamy, line 7: This will break local dev/testing, won't this ?
We need this file for local development.

Earthfile
```diff
@@ -43,64 +43,57 @@ source:
...
     COPY --keep-ts Cargo.toml Cargo.lock Makefile.toml ./
     COPY --keep-ts deny.toml ./
-    COPY --keep-ts --dir lightway-core lightway-app-utils lightway-client lightway-server tests .cargo ./
+    COPY --keep-ts --dir lightway-core lightway-app-utils lightway-client lightway-server tests ./
 
-# build builds with the Cargo release profile
+# Target +build always builds native binaries for the host platform by earthly.
+# You may use `--platform linux/amd64` to override the host platform, to do a native compilation in emulation.
+# E.g. `earthly --platform linux/amd64 +build` on Apple Silicon Macs to build for x86_64. Slower but almost always works.
 build:
```
kp-mariappan-ramasamy, line 50: This change breaks `earthly doc`
https://docs.earthly.dev/docs/earthly-command#description-3

The first word of the comment should be the target name

.cargo/config.toml
kp-mariappan-ramasamy: Or can you update the README about this alternate environmental variables while doing local development

Earthfile
```diff
@@ -43,64 +43,57 @@ source:
...
-    ELSE IF [ "$ARCH" = "riscv64" ]
-        SET target = "riscv64gc-unknown-linux-gnu"
-    END
+    DO lib-rust+CARGO --args="build --release --features io-uring" --output="release/lightway-(client|server)$"
+
+    SAVE ARTIFACT ./target/release/lightway-client AS LOCAL ./target/release/
+    SAVE ARTIFACT ./target/release/lightway-server AS LOCAL ./target/release/
+
+# Target like +build-cross-<variant> are for cross-compiling to non-host platforms.
+# We assume the host is either linux/amd64 or linux/arm64 for these targets.
```
kp-mariappan-ramasamy: For `linux/arm64` host, `build-cross-arm64` should be a cross compilation

Earthfile
```diff
@@ -43,64 +43,57 @@ source:
...
-    ELSE IF [ "$ARCH" = "riscv64" ]
-        SET target = "riscv64gc-unknown-linux-gnu"
-    END
-
-    DO lib-rust+CARGO --args="build --release --features kyber_only --target=$target --bin lightway-client --target-dir ./lightway-client-kyber" --output="lightway-client-kyber/$target/release/lightway-client"
-    SAVE ARTIFACT ./lightway-client-kyber/$target/release/lightway-client AS LOCAL ./target/$target/release/lightway-client-kyber
-
-# test executes all unit and integration tests via Cargo, in the host's native platform
+# Target +test compiles all unit and integration tests, natively for the host platform.
+# Again, you may use `--platform linux/amd64` to override the host platform.
```
kp-mariappan-ramasamy: Same as before.
Do not remove the target name as the first word. It will remove this target from `earthly doc`

Earthfile
kp-samuel-tam: I didn't know `earthly doc` was a thing. Thanks!

.cargo/config.toml
kp-samuel-tam: Yes, we need to add `.cargo/config.toml` ourselves if we cross-compile in local dev/testing outside of earthly. I will add this in README.

kp-mariappan-ramasamy (dismissed):

.cargo/config.toml
kp-thomas-yau, line 7: Please also update .gitignore or else we might accidentally add this toml back

kp-thomas-yau (approved):
Gucci

PR #285, opened 2025-09-19
client: Update server route on floating connections

kp-mariappan-ramasamy:
## Description

Subscribe to route changes and update server route accordingly.
For windows, additionally listen to addr changes since routes changes are not always published.

## Motivation and Context

Lightway UDP protocol supports floating clients ie. Clients can move between different networks and the connection stays online/resume without disconnecting.

When lightway client connects to a server, it creates a static route to server to override the default tunnel route. When network change, this has to be changed appropriately to float the connection to support floating connection

## How Has This Been Tested?

Verified floating connection is working in windows using Ethernet and Wifi and switching between two connections multiple times

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/route_manager.rs
+++ b/lightway-client/src/route_manager.rs
@@ -1,8 +1,11 @@
 use anyhow::Result;
-use route_manager::{AsyncRouteManager, Route, RouteManager as SyncRouteManager};
+use route_manager::{
+    AsyncRouteListener, AsyncRouteManager, Route, RouteManager as SyncRouteManager,
+};
 use serde::{Deserialize, Serialize};
 use std::net::{IpAddr, Ipv4Addr};
 use thiserror::Error;
+use tokio::task::JoinHandle;
 use tracing::warn;
 
 #[cfg(windows)]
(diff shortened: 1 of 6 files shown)


lightway-client/src/route_manager.rs
```diff
@@ -83,16 +93,71 @@ pub enum RoutingTableError {
...
     route_manager: SyncRouteManager,
     route_manager_async: AsyncRouteManager,
+    server_ip: IpAddr,
+    tun_index: u32,
+    tun_peer_ip: IpAddr,
+    tun_dns_ip: IpAddr,
     vpn_routes: Vec<Route>,
     lan_routes: Vec<Route>,
     server_route: Option<Route>,
+    monitor_task: Option<tokio::task::JoinHandle<()>>,
```
kp-brandon-chew: This seems to always be `None`, not sure if intentional?

kp-weiwen-goh (approved):
LGTM

lightway-client/src/route_manager.rs
kp-mariappan-ramasamy: Not intentional. This was before I refactor to Inner and missed to remove.
The actual task is saved inside `RouteManager`
https://github.com/expressvpn/lightway/pull/285/files#diff-b47d6885e00c3bb2aa7995f1d5769529b827db83914fb010f7d18387f59215d6R97
This can be removed.

PR #290, opened 2025-09-30
client: Add ready signal to outside_io_task

kp-thomas-yau:
## Description

Add optional oneshot channel parameter to outside_io_task to signal when the socket is ready for reading after the first successful poll.

## Motivation and Context

To allow client apps to know when the socket is actually ready to read

## How Has This Been Tested?

Tested on a local device, verified that it will indeed send a ready signal properly.

## Types of changes

## Checklist:


kp-weiwen-goh (approved):
LGTM

PR #307, opened 2025-11-06
COVPN-31: server: Add encoding_enabled metric

kp-weiwen-goh:
## Description

Adds a new `encoding_enabled` counter metric to server.

This is incremented when the server accepts a client's encoding request.

## Motivation and Context

To monitor codec enable rate

## How Has This Been Tested?

Should compile. EncodingStateChanged event handler has been tested previously.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/connection_manager.rs
+++ b/lightway-server/src/connection_manager.rs
@@ -147,6 +147,15 @@ fn handle_tls_keys_update_complete() {
 }
 
 #[instrument(level = "trace", skip_all)]
+fn handle_encoding_state_changed(enabled: bool) {
+    if enabled {
+        metrics::connection_encoding_enabled();
+    } else {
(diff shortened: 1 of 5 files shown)


lightway-server/src/connection_manager.rs
```diff
@@ -146,6 +146,13 @@ fn handle_tls_keys_update_complete() {
     metrics::connection_key_update_complete();
 }
 
+#[instrument(level = "trace", skip_all)]
+fn handle_encoding_state_changed(enabled: bool) {
+    if enabled {
+        metrics::connection_encoding_enabled();
```
kp-mariappan-ramasamy, line 152: We need to update the metric when state becomes disabled.
Might be better to change counter to guage and then increment/decrement based on encoding status

PR #308, opened 2025-11-10
server: remove average dependency

yanganto:
Remove average dependency in server crate

## Description
Remove average dependency and implement without additional crate

## Motivation and Context
This refactor PR follow the TODO hint in the code.
Two facts about this refactor.

- Macro takes compiling time, and we use average crate macro, but the lines of code does not significantly reduce.
- The macro implement inline functions to avoid performance costs, but the `fn new` will call the default functions, and these are not inline functions.
https://docs.rs/average/latest/src/average/macros.rs.html#129

## How Has This Been Tested?
This refactor obey the current test cases.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,17 +225,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "average"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "04c5a79a6c36c527e1c31bc0effa93af02e7278b7369b83e0674255290b6c1b0"
-dependencies = [
- "easy-cast",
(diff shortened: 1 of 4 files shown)


kp-mariappan-ramasamy (dismissed):
LGTM

yanganto:
@kp-mariappan-ramasamy 
Sorry for delay, just rebase and squash these two commit into one, because just notice that we are not using squash after the PR merged.  Thank for review.

kp-antonio-yang (dismissed):

kp-mariappan-ramasamy (approved):

PR #318, opened 2025-11-26
Add expresslane support

kp-mariappan-ramasamy:
## Description

This PR adds an new feature `expresslane` for UDP connections, where we split control plane and data plane.
Control plane uses existing DTLS, whereas the data plane use Expresslane

Detailed design is at docs/expresslane.md

## Motivation and Context

To make Data communication stateless and thus can be parallelized later.

## How Has This Been Tested?

Tested manually with expresslane client and server, and verified traffic flows properly.

## Types of changes

## Checklist:


kp-thomas-yau:
Should we add new e2e tests for testing connections between v1.2 and v1.3 server/clients?

PR #321, opened 2025-11-28
Keepalive fixes

kp-mariappan-ramasamy:
## Description

Refactor keepalive to be consistent/clear and fix tracer implementation for linux client
There are multiple commits, each solving a specific usecase/problem. It is explained in the commit messages.

## Motivation and Context

Tracer implementation was not working with Linux clients.

## How Has This Been Tested?

Verified connection drops out, when it detects there is a network error and not able to receive packets

## Types of changes

## Checklist:


lightway-client/src/keepalive.rs
```diff
@@ -227,17 +183,13 @@ async fn keepalive<CONFIG: SleepManager, CONNECTION: Connection>(
...
                         timeout.as_mut().set(None.into())
                     },
                     Message::NetworkChange => {
                         if !matches!(state, State::Pending) {
                             tracing::info!("sending keepalives because of {:?}", msg);
                             state = State::Needed;
                         }
-                        // Reset failure counter on new interface
-                        failed_keepalives = 0;
                     },
```
kp-thomas-yau, line 199: I think we should also reset the timeout here for network change as well right?

kp-mariappan-ramasamy: Do you mean to say timer set ? or reset ?
The timeout will be set in https://github.com/expressvpn/lightway/pull/321/files#diff-c3d0f0f201bcd5bb5e1a9cc6981e47ab37183d85912b61e444f54b3310c704d4R214 since we now set the state to State::Needed.

PLMK if i miss something, where we need to timeout during networkchange.

kp-thomas-yau: The current logic says that if we're waiting for a keepalive reply from the server, we don't reset timeout. But I think we still need to reset the timeout even we've sent a ping packet already, as previous timeout value has started counting down on the older socket/interface, but not on the new socket, so the new socket/interface timeout value got reduced in this case

PR #323, opened 2025-12-04
Add Nix cross compilation support

kp-mariappan-ramasamy:
## Description

This PR refactors the Nix flake into a modular architecture and introduces a CI workflow that builds and verifies Lightway packages across all supported platforms and architectures.

Some things to note:
- Currently aarch64 musl builds are failing from aarch64 linux machine. Commented it for now
- For now, I have kept the CI separately from regular ones, since this seems very time consuming
- Nix CI will run nightly to verify all the native/cross builds from all platforms.
- Will try to add some nix cache and will include in regular ci in future commits

## Motivation and Context

To make compiling to any target possible from major development platforms.
Ex: mac to linux. linux x64 -> linux aarch64

## How Has This Been Tested?

**Local Testing:**
  - Built all package variants on x86_64-linux: `nix flake show --all-systems`
  - Verified package names match derivation names
  - Confirmed cross-compilation produces correct binaries 
  - Verified binaries manually by running E2E tests manually and made sure it works

  **CI Testing:**
  - All 32 build jobs pass successfully
  - Binary verification confirms correct linking types (static vs dynamic)
  - Architecture checks validate cross-compilation correctness

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/nix-build.yaml
+++ b/.github/workflows/nix-build.yaml
@@ -103,16 +103,16 @@ jobs:
             build_type: "native"
 
           # Musl static builds (same arch, different libc)
-          - runner: ubuntu-24.04-arm
-            platform: aarch64-linux
-            package: lightway-client-aarch64-linux-musl
-            verify_type: "musl-static"
-            build_type: "native"
-          - runner: ubuntu-24.04-arm
-            platform: aarch64-linux
-            package: lightway-server-aarch64-linux-musl
-            verify_type: "musl-static"
-            build_type: "native"
+          # - runner: ubuntu-24.04-arm
+          #   platform: aarch64-linux
+          #   package: lightway-client-aarch64-linux-musl
+          #   verify_type: "musl-static"
+          #   build_type: "native"
+          # - runner: ubuntu-24.04-arm
+          #   platform: aarch64-linux
+          #   package: lightway-server-aarch64-linux-musl
+          #   verify_type: "musl-static"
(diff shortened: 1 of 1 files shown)


nix/default.nix
```diff
@@ -68,7 +68,22 @@ rustPlatform.buildRustPackage {
...
+  # as statically linked. But SIGSEGV crashes on aarch64
+  #
+  # Tried to also disable PIE to make it statically linked without PIE, but it didn't work.
+  # aarch64 musl always produces PIE binaries with PT_INTERP section
+  #
+  # Note 2:
+  # For cross-compilation from macOS, disable lld and use gcc directly to
+  # avoid platform_version flags
   RUSTFLAGS =
     lib.optionalString isStatic "-C target-feature=+crt-static -C link-arg=-static"
```
kp-weiwen-goh, line 97: `-static` is set in previous commit already, should be set here.

(also repeated sections in commit message)

kp-weiwen-goh (approved):
LGTM

Not sure if the cache has been evicted since we are at 10GB already, but right now it looks like it is only using 3.5GB of cache.
```
❯ gh cache list --ref refs/heads/nix-compilation --json 'sizeInBytes' --jq 'map(.sizeInBytes) | add'
3589351630
```

nix/default.nix
kp-mariappan-ramasamy: Fixed the commit message header and description

PR #330, opened 2025-12-15
Fix route manager to support IPv6 server addresses with IPv4 tunnel

kp-samuel-tam:
## Description

This ensures correct route setup when Lightway client is connected to an IPv6 endpoint. The IPv4 tunnel stays unchanged.

Changes:
- Use /128 prefix for IPv6 server routes (was incorrectly using /32)
- Use /32 prefix for IPv4 server routes (unchanged)
- Monitor IPv6 default route changes when server is IPv6, and IPv4 for IPv4
  - e.g. don't monitor IPv4 route changes when connected to an IPv6 server
- Fix LAN mode to only use gateway if address family matches route
  - e.g. when connecting to IPv6 VPN server, don't do `route add 192.168.0.0/16 via 2001:db8::1` (breaks)
- Fix `find_route` to return proper error instead of panicking on None
  - nicer error, and don't assume a default route exists
- Add mock IPv6 default route via loopback for test environments without IPv6
  - when testing environment has no IPv6 setup, just add a "mock" default route to `::1`
  - we don't need a working default route, it works as long as the tests can look up a route for VPN server address
- Add `host_prefix_len()` and `same_ip_family()` helper functions

Tests:
- Add IPv6 and IPv4 test constants, update all tests to use them explicitly
- Update `create_test_setup()` to accept server_ip parameter
- Add `test_ipv6_server_route_manager_creation()`
- Add `test_privileged_ipv6_server_initialize_route_manager()`
- Add `test_privileged_ipv6_server_route_update()`
- Mock route only looks for ::1 (not 127.0.0.1) when finding loopback

The tunnel interface remains IPv4-only (`TUN_PEER_IP`, `TUN_DNS_IP`), only the server connection endpoint supports IPv6.

## Motivation and Context

Right now when connecting to an IPv6 endpoint, a route like `8321:8bec:666c:0305:9005:9f3e:e29f:3920/32` (which is functionally `8321:8bec::/32`) will be established which is not correct; everything of `8321:8bec::/32` will leak. Only `/128` should be correct.

## How Has This Been Tested?

Wrote tests; if they work it works as intended.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/route_manager.rs
+++ b/lightway-client/src/route_manager.rs
@@ -3,7 +3,7 @@ use route_manager::{
     AsyncRouteListener, AsyncRouteManager, Route, RouteManager as SyncRouteManager,
 };
 use serde::{Deserialize, Serialize};
-use std::net::{IpAddr, Ipv4Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use thiserror::Error;
 use tokio::task::JoinHandle;
 use tracing::{trace, warn};
@@ -91,6 +91,24 @@ pub enum RoutingTableError {
     RoutingManagerError(std::io::Error),
     #[error("Server route already exists, try modifying it instead")]
     ServerRouteAlreadyExists,
+}
+
+/// Returns the host prefix length for an IP address
+/// (32 bits for IPv4, 128 bits for IPv6)
+fn host_prefix_len(ip: &IpAddr) -> u8 {
+    match ip {
+        IpAddr::V4(_) => Ipv4Addr::BITS as u8,
+        IpAddr::V6(_) => Ipv6Addr::BITS as u8,
+    }
+}
+
+/// Checks if two IP addresses are from the same address family
+/// (both IPv4 or both IPv6)
+fn same_ip_family(ip1: &IpAddr, ip2: &IpAddr) -> bool {
+    matches!(
+        (ip1, ip2),
+        (IpAddr::V4(_), IpAddr::V4(_)) | (IpAddr::V6(_), IpAddr::V6(_))
+    )
 }
 
 pub struct RouteManager {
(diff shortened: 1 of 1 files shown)


lightway-client/src/route_manager.rs
```diff
@@ -412,13 +412,22 @@ impl RouteManagerInner {
         #[cfg(windows)]
         let server_route = server_route.with_metric(0);
 
+        // For NoExec mode, store the route but don't actually add it
```
kp-mariappan-ramasamy: Not clear, why do we need to move this.
Please add changes like this in a separate commit with clear description on why it is needed.

lightway-client/src/route_manager.rs
```diff
@@ -403,7 +398,12 @@ impl RouteManagerInner {
 
         // Create server route with optional gateway - handles both direct routes (containers)
         // and routed networks (host systems with gateways)
-        let server_route = Route::new(server_ip, 32).with_if_index(default_interface_index);
+        // Use /32 for IPv4, /128 for IPv6
+        let prefix = match server_ip {
```
kp-mariappan-ramasamy: This code is repeated. Better to move this to a separate function.
And use https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#associatedconstant.BITS for prefix

lightway-client/src/route_manager.rs
kp-samuel-tam: I understood the code wrong and wrote a wrong internal state check for `NoExec`. It should be fixed now, thanks!

lightway-client/src/route_manager.rs
kp-samuel-tam: Done, for this and checking same family as well.

PR #384, opened 2026-04-10
CVPN-2496: Support `recvmmsg` on Linux/Android

kp-thomas-yau:
## Description

Use the same `BatchRecvSyscall` trait for Linux/Android platform. It should be simpler as it have existed since [glibc 2.12](https://man7.org/linux/man-pages/man2/recvmmsg.2.html#:~:text=Linux.-,HISTORY,-top), which came out in 2010-05-03.

Run the new client in e2e tests as well.

Note: `nix` crate supports [recvmmsg](https://docs.rs/nix/0.31.2/nix/sys/socket/fn.recvmmsg.html) as well, but imo it seems way too over-complicated for our use case, so I'm just using `libc` crate for now, which is acting quite similar with apple's recvmsg_x function.

## Motivation and Context

Improve download speed on Linux platform

## How Has This Been Tested?

Generally, the performance of regular Lightway Rust connection is already pretty quick on both Android/Linux platform. 

Not much improvement on Linux right now due to bottleneck on the testing server, but if I do launch my own server on a remote machine and route VPN traffic over it, we do see some performance improvement:
| (in Mbps) | Speedtest Download | Speedtest Upload |
|:---:|:---:|:---:|
| recv_from | 1107 | 775 |
| recvmmsg | 1428 | 824 |

On Android (Samsung S26 w/ 5Gbps line):
| (in Mbps) | Speedtest Download | Speedtest Upload | Speedtest (w/ Expresslane) Download | Speedtest (w/ Expresslane) Upload |
|:---:|:---:|:---:|:---:|:---:|
| recv_from | 1212 | 1216 | 1451 | 1234 |
| recvmmsg | 1240 | 1271 | 1580 | 1271 |

## Types of changes

## Checklist:

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -51,7 +51,13 @@ source:
 build:
     FROM +source
 
-    DO lib-rust+CARGO --args="build --release --features io-uring" --output="release/lightway-(client|server)$"
+    ARG BATCH_RECEIVE=false
+    LET features = "io-uring"
+    IF [ "$BATCH_RECEIVE" = "true" ]
+        SET features = "$features,batch_receive"
+    END
+
+    DO lib-rust+CARGO --args="build --release --features $features" --output="release/lightway-(client|server)$"
 
     SAVE ARTIFACT ./target/release/lightway-client AS LOCAL ./target/release/
     SAVE ARTIFACT ./target/release/lightway-server AS LOCAL ./target/release/
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -299,9 +299,8 @@ jobs:
       - name: Run clippy
         run: nix develop --command cargo make clippy
 
-      - name: Run recvmsg_x test (macOS only)
-        if: matrix.os == 'macos-26'
-        run: sudo nix develop --command cargo test batch_receiver --features batch_receive
+      - name: Run batch receiver test
+        run: nix develop --command cargo test batch_receiver --features batch_receive
 
       - name: Run privileged tests (macOS only)
         if: matrix.os == 'macos-26'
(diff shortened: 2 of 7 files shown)


Earthfile
```diff
@@ -51,7 +51,13 @@ source:
 build:
     FROM +source
 
-    DO lib-rust+CARGO --args="build --release --features io-uring" --output="release/lightway-(client|server)$"
+    ARG BATCH_RECEIVE=false
```
kp-mariappan-ramasamy, line 54: We could consider changing this batch_receive as runtime feature instead of compile feature.
Will avoid these unnecessary branches. Also it will avoid compiling two set of binaries and save CI time.

Just let the OutsideIO implementation decide whether it supports batch_receive or not

lightway-client/src/lib.rs
kp-mariappan-ramasamy, line 1: Regarding this commit `client: Update outside_io_task to clear buffer at the top of the loop`

Here we have mutable count which is been set after the successful recv_buf(s) call. If OutsideIO returns WouldBlock, then the count variable will not be set and value is the one from previous loop which is wrong.
We could reset at the beginning of the loop, but again it seems unnecessary.

I would suggest to move to older way of doing at the last.

Earthfile
kp-thomas-yau: I think I did `batch_receive` as a compile feature before just to make it explicitly opt-in and also make it obvious Windows doesn't support this.

I can make the changes in the next PR after this got merged so that it won't bloatify this PR

kp-mariappan-ramasamy (approved):
LGTM

Issue #310, opened 2025-11-12
Support for Android?


chmuhammadasim:
Hello,

I am trying to build a mobile application using **React native** that can connect to **lightway server** but I am not sure how to do that. I have read the documentation but I was not able to find the answer. I want to build `.so` file that I can connect with react native. These are the things that I tried.

* Building `.so` using lightway-core but the supported build is only for `[linux|macos|windows].`
* I tried to use lightway-client as well but I am not able to build the rust files in order to use them in the project using JNI etc.

In `lightway` repository inside `lightway-app-utils` and `lightway-client` code, I have seen multiple places where we are disabling the mobile like

    #[cfg(mobile)]
    Err(std::io::Error::from(std::io::ErrorKind::Unsupported))

Can you guys tell me what am I doing wrong? Also, Is it supported on mobile or is it there only for desktop for now?

Any help would be appreciated.

Thanks

kp-mariappan-ramasamy:
Hello @chmuhammadasim 

Thanks for trying lightway-client with react-native. 
Android and iOS as opposed to desktop platforms, is little different and unique in how the VPN is handled.

For Android, you need to go through this Android documentation to understand how VPN works:
https://developer.android.com/develop/connectivity/vpn

The VPN service mentioned in the above documentation will be the app which integrates with rust lightway client. 

> Building .so using lightway-core but the supported build is only for [linux|macos|windows]

This just means we officially support the above platforms, with the reference implementation available in this repo.
We do not have any reference Android app at the moment. 

Allow us sometime, we will see whether we can put a short guide or PR on how to compile for Android platform.
It should give you enough pointers to get started.

chmuhammadasim:
Hello @kp-mariappan-ramasamy 

Thank you for the detailed explanation and for clarifying the current platform support.

I’ve gone through the Android VPN documentation and implemented the VPNService-based approach as suggested. I’m able to successfully create the tunnel.

However, when integrating the Rust **lightway client** logic for android, I’m facing a consistent issue during the **handshake phase**:

* The VPN service starts correctly
* The tunnel/interface is created successfully
* The connection attempt reaches the server
* ❌ The Lightway handshake fails and the session is never fully established

This works as expected when connecting to the same Lightway server from Linux using the official client, so the server configuration itself seems correct.

[...]

PR #386, opened 2026-04-13
Add mobile libs

kp-antonio-yang:
## Description
- [X] Release mobile client and the build flow to roll out .aar
- [ ] Setup CI to make sure it is reproducible (required powerful CI runner we do internal only)
- [X] Expose the tun local_ip and dns_ip
- [X] Artifacts from building `aar` under target, so developer can easy to clean up
- [X] lightway as top domain in `.aar`
- [X] Building scripts align with existing build system (cargo-make)
## Motivation and Context
Driven by request #310 for OSS Lightway, this PR ships the Android mobile library (.aar) along with its build pipeline as groundwork for an upcoming demo app.

The feature branch was developed on NixOS using devshells to ensure fully reproducible builds. That said, the pipeline makes no hard assumptions about the environment — you can build on any machine with Rust, cargo-make, Java, Android SDK/NDK installed, as described in the README.

A subset of error codes and functions in the library are iOS-specific. Since an iOS OSS release is planned, these are included now rather than introduced later as a breaking change.

## How Has This Been Tested?
- [X] The .aar is build with `nix develop .#android -c build` both on my local machine and the internal CI runner
- [X] The .aar is tested in a internal android demo project, which will be released later.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.


.github/workflows/nix-build.yaml
```diff
@@ -12,6 +12,7 @@ concurrency:
 
 jobs:
   lint:
+  name: Flake check
```
kp-mariappan-ramasamy: This is unrelated change, Please move this change to previous commit.
https://github.com/expressvpn/lightway/pull/386/changes/6f1dbd33810efee17e38ca62c847f0f658c9ce5b

lightway-client/src/lib.rs
```diff
@@ -7,6 +7,12 @@ pub mod platform;
 #[cfg(desktop)]
 pub mod route_manager;
 
+#[cfg(feature = "mobile")]
```
kp-mariappan-ramasamy: We can use `#[cfg(mobile)]` here

nix/modules/devshells.nix
```diff
@@ -45,14 +45,37 @@
...
           pkgsCross.callPackage ../shell.nix ({
             rustc = rustLatest.minimal.override {
-              # TODO: automatically chose targets based on architecture
               targets = [
-                "x86_64-linux-android"
-                "i686-linux-android"
-                "armv7-linux-androideabi"
+                # "x86_64-linux-android"
+                # "i686-linux-android"
+                # "armv7-linux-androideabi"
```
kp-mariappan-ramasamy: Why are we commenting this code in this commit ?

lightway-client/build.rs
```diff
@@ -9,6 +9,7 @@ fn main() {
         // windows - supported natively
         // Mobile Platforms
         android: { target_os = "android" },
+        apple: { target_vendor = "apple" },
```
kp-mariappan-ramasamy: We have one another apple alias defined below. Considering there is no error, I believe it uses the latest. You may have to delete the below one

lightway-client/Cargo.toml
```diff
@@ -11,13 +11,13 @@ name = "lightway-client"
 path = "src/main.rs"
 
 [lib]
-name = "lightway_client_lib"
+name = "lightway_client"
```
kp-mariappan-ramasamy, line 14: Please move this change to the original commit. Do not add unrelated things to a single commit

nix/modules/devshells.nix
```diff
@@ -73,9 +73,9 @@
           pkgsCross.callPackage ../shell.nix ({
             rustc = rustLatest.minimal.override {
               targets = [
-                # "x86_64-linux-android"
-                # "i686-linux-android"
-                # "armv7-linux-androideabi"
+                "x86_64-linux-android"
+                "i686-linux-android"
+                "armv7-linux-androideabi"
```
kp-mariappan-ramasamy: I see we have again uncommented the code. Squash this change to the original code which introduced it.

nix/shell.nix
```diff
@@ -26,7 +26,7 @@ mkShell (
       # TODO: Use isEmpty after Nix 2.14+
       ${lib.optionalString (lib.length (lib.attrNames useCargoTools) > 0) ''
         ${lib.concatMapAttrsStringSep "\n" (
-          pkg: ver: "cargo install cargo-${pkg} --version ${ver}"
+          pkg: ver: "cargo install cargo-${pkg} --version ${ver} --force"
```
kp-mariappan-ramasamy: Why do we add force here? What was the issue ?

lightway-client/Cargo.toml
```diff
@@ -58,9 +58,12 @@ tracing-panic = { version = "0.1.2", optional= true }
 [dev-dependencies]
 more-asserts.workspace = true
 test-case.workspace = true
+tokio = { workspace = true, features = ["test-util"] }
 serial_test = "3.2.0"
 tun-rs.workspace = true
 tempfile = "3.24.0"
+mockall = "0.14.0"
```
kp-mariappan-ramasamy, line 65: This changes has nothing to do with this commit:
https://github.com/expressvpn/lightway/pull/386/changes/58fb72258a63bf62cf4a05f673b99b2356956cff

Please move it to appropritate commit

kp-mariappan-ramasamy (commented):
Overall changes seems okie. But my main concern is the organization of commits. Changes are scattered all over the commits instead of being logical and thus it is difficult to follow the changes. 

And I tried to build locally using nix and it failed with
```
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: build-aar
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Execute Command: "cargo" "ndk" "--bindgen" "-t" "x86_64" "-t" "x86" "-t" "armeabi-v7a" "-t" "arm64-v8a" "-o" "/work/github/expressvpn/lightway/target/aar/jniLibs" "build" "--features" "mobile" "--profile" "aar" "-p" "lightway-client" "--lib"
    Building x86_64 (x86_64-linux-android)
error[E0308]: mismatched types
   --> lightway-client/src/mobile/lightway.rs:103:45
    |
103 |                 Ok((ConnectionType::Stream, sock))
    |                                             ^^^^ expected `Arc<dyn OutsideIO>`, found `Tcp`
    |
    = note: expected struct `Arc<dyn OutsideIO>`
               found struct `outside::tcp::Tcp`

error[E0308]: mismatched types
   --> lightway-client/src/mobile/lightway.rs:115:47
    |
115 |                 Ok((ConnectionType::Datagram, sock))
    |                                               ^^^^ expected `Arc<dyn OutsideIO>`, found `Udp`
    |
    = note: expected struct `Arc<dyn OutsideIO>`
               found struct `outside::udp::Udp`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `lightway-client` (lib) due to 2 previous errors
note: If the build failed due to a missing target, you can run this command:
note:
note:     rustup target install x86_64-linux-android
Error while executing command, exit code: 101
```

When I tried to compile again, it then tries to compile cargo-make, cargo-ndk again, which seems unnecessary.

.github/workflows/nix-check.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: actions/checkout@v6
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build Android
+        run: |
+          nix develop .#aar -c build
+          ls -lh lightway-client/src/platform/lightway/build/outputs/aar/lightway-release.aar
```
kp-antonio-yang: @kp-mariappan-ramasamy Do we comment out these or delete there?

kp-mariappan-ramasamy: Lets remove this until we add beefy runners

nix/shell.nix
kp-antonio-yang: `cargo-*` tools are not project-scoped binaries — they are installed globally. If multiple projects share the same environment, the build flow may skip reinstalling the tools and silently use whatever installed by a previous project. This is a subtle reprehensibility issue I have encountered before.
Using `--force` ensures the correct version is always installed, regardless of what is already present in the environment, keeping the build fully reproducible.

kp-antonio-yang: By the way,  you can `ctrl + c` to break the shell hook, skip the `cargo-*` installation, but still in the shell.  If you 100% sure the dependency is correct.

kp-antonio-yang: These note is added to dev shell, such that developers will easy to know and skip the installation as they want

nix/modules/devshells.nix
```diff
@@ -23,8 +24,62 @@
         msrv = pkgs.callPackage ../shell.nix {
           rustc = rustMsrv.default;
         };
+      }
+      // lib.optionalAttrs (lib.hasSuffix "linux" system) {
+        android =
+          let
+            ANDROID_NDK_VERSION = "27.2.12479018";
```
kp-mariappan-ramasamy: We should better move these values to a separate file like constants.nix to make it easy to identify.

nix/shell.nix
```diff
@@ -14,38 +14,47 @@
   rustc,
   isStatic ? false,
   defaultTarget ? null,
+  useCargoTools ? { },
```
kp-mariappan-ramasamy: We can remove this unused one now

README.md
```diff
@@ -26,7 +26,7 @@ Protocol and design documentation can be found in the
 
 ## Supported platforms
 
-Lightway Rust implementation currently supports Linux OS and macOS.
+Lightway Rust implementation currently supports Linux OS, macOS and Android.
```
kp-mariappan-ramasamy: We can be little cautious here. Lets update the readme about Android, after we have an working example. And not now.

lightway-client/src/mobile/lightway.rs
kp-mariappan-ramasamy, line 1: Commit message of `client: mobile: expose local_ip and dns_ip in UserSettings` can be improved.

```
The TUN ip settings and DNS setting should match IP and DNS IP set in Android tunnel, so we expose it as a config in case the android integration choose to use different address.
```

deny.toml
```diff
@@ -19,6 +19,7 @@ allow = [
     "BSD-3-Clause",
     "ISC",
     "MIT",
+    "MPL-2.0",
```
kp-mariappan-ramasamy, line 22: Please mention which library pulls in this dependency.
I think this change belongs to previous commit where we introduce new cargo tracing libraries

README.md
```diff
@@ -70,6 +70,13 @@ linker = "riscv64-linux-gnu-gcc"
 runner = "qemu-riscv64-static -L /usr/riscv64-linux-gnu -cpu rv64"
 ```
 
+## Mobile
+The android mobile library(.aar) can build by `cargo make build-android` or `nix develop .#android -c build` under project root.
+We rely on cargo and **gradlew**, so the `cargo clean` is not enough to clean all the artifacts, the extra `gradlew clean` is needed.
```
kp-mariappan-ramasamy, line 75: Since we plan to use cargo make for building, lets also add a `cargo make clean` which will clean all.

kp-mariappan-ramasamy (commented):
We already have platform specific files in each folfer like below

```
lightway-client/src/platform
├── linux
│   ├── dns_manager
│   │   ├── direct_file.rs
│   │   ├── resolvconf.rs
│   │   └── resolvectl.rs
│   └── dns_manager.rs
├── linux.rs
├── macos
│   └── dns_manager.rs
├── macos.rs
├── windows
│   ├── addr_monitor.rs
│   ├── crypto.rs
│   ├── dns_manager.rs
│   └── utils.rs
└── windows.rs
```

We shuold move all android related files inside `lightway-client/src/platform/android` to keep it clean

nix/modules/devshells.nix
```diff
@@ -23,8 +24,65 @@
         msrv = pkgs.callPackage ../shell.nix {
           rustc = rustMsrv.default;
         };
+      }
+      // lib.optionalAttrs (lib.hasSuffix "linux" system) {
+        android =
+          let
+            androidConstants = (import ./constants.nix).android;
+            ANDROID_NDK_VERSION = "27.2.12479018";
```
kp-mariappan-ramasamy: This still uses hardcoded value instead of the one from constants.nix

kp-antonio-yang: Sorry, just fixed

kp-mariappan-ramasamy (dismissed):

kp-mariappan-ramasamy:
@kp-antonio-yang I get some error with the recent changes

```
❮ nix develop .#android -c build
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO -
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: build-android
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Execute Command: "cargo" "ndk" "--bindgen" "-t" "x86_64" "-t" "x86" "-t" "armeabi-v7a" "-t" "arm64-v8a" "-o" "/work/github/expressvpn/lightway/target/android/jniLibs" "build" "--features" "mobile" "--release" "-p" "lightway-client" "--lib"
    Building x86_64 (x86_64-linux-android)
    Building x86 (i686-linux-android)
    Building armeabi-v7a (armv7-linux-androideabi)
    Building arm64-v8a (aarch64-linux-android)
     Copying libraries to /work/github/expressvpn/lightway/target/android/jniLibs
[cargo-make] INFO - Running Task: bindgen-android

thread 'main' (2830260) panicked at /home/maari/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/uniffi-0.26.1/src/lib.rs:30:21:
called `Result::unwrap()` on an `Err` value: failed to open file `target/aarch64-linux-android/android/liblightway_client.so`

Caused by:
    No such file or directory (os error 2)
```

lightway-client/src/mobile.rs
```diff
@@ -115,10 +116,17 @@ pub struct LightwayUserSettings {
     defer_timeout: i64,
     /// Enable Expresslane for UDP connections
     pub enable_expresslane: bool,
+    /// The local ip for TUN interface in ipv4 format
+    pub local_ip: String,
+    /// The dns ip in ipv4 format
+    pub dns_ip: String,
```
xv-thomas-leong, line 122: We already have ways of passing `IpAddress` via uniffi. Can the same be used here?

https://github.com/expressvpn/lightway/blob/3e2db81bfb2257d9335c2fad633b0a2468025f3e/lightway-client/src/endpoint.rs#L44-L49

kp-antonio-yang: Unfortunately this approach won't work as simply here for a couple of reasons:

- The inner type of `IpAddress` in the endpoint is `IpAddr`, not `Ipv4Addr`, which is what `setup_tunnel_interface` or  `io::inside::Tune::new()`expects.
- A single line parser is sufficient and clear: `self.local_ip.parse::<Ipv4Addr>()`.
- Introducing another wrapper around `Ipv4Addr` would add unnecessary complexity and compile time overhead. Following KISS here might be the right call.

https://github.com/expressvpn/lightway/blob/0ee61b9aa4f1df8dbe9e21be8d55c549627014bc/lightway-client/src/io/inside/tun.rs#L25

kp-antonio-yang: By the way, we will drop `IpAddress` of other fields and use `.parse::<IpAddr>()` in the next PR on handling the configs.  If you see benefit from the current approach.

xv-thomas-leong: @kp-thomas-yau what do you think about this?

kp-thomas-yau: I think for `IpAddress`, it's mostly convenient wrapper for the client apps and rust to do less conversion whenever each-side wants to read/write it, so mostly just a convenient wrapper for us to use less `into`/`parse`

xv-thomas-leong: Thanks @kp-thomas-yau. @kp-antonio-yang how about we keep it as is and I'm okay as long as we are being consistent.

I'm happy to approve if all other issues are fixed :)

kp-antonio-yang: This diff gives a quick view of what the code looks like with fewer wrapper layers. The error types remain strongly typed, making issues straightforward to address, while removing the overhead of unnecessary deserialization, `to_string` conversions, and `uniffi::custom_type!` macro. The reduction in indirection makes the flow noticeably easier to follow:

https://github.com/expressvpn/lightway/pull/391/changes/74c7b913bc7cdcd0f3cf3650c33065ee65701b07

It is also worth noting that keeping `IpAddress` and adding yet another wrapper around `Ipv4Addr` on top of it would increase code size without any real benefit — removing it is the cleaner path.

kp-mariappan-ramasamy (approved):

lightway-client/src/mobile.rs
kp-antonio-yang: We will merge this PR, and we can discuss about the typing of the interface in later PRs.

PR #388, opened 2026-04-15
client: Add default config and JSON schema generation

kp-antonio-yang:
## Description
- [X] let `twelf` retired
- [X] Fix RUSTSEC-2024-0320
- [X] Default values with Default traits (Follow Rust Convention)
- [X] Generate config to JsonSchema, YamlConfig for client
- [X] Merge the mobile inputs(RustUserSetting, RustEndpointConfig) into configs, such that mobile and cli clients share from the same config from now

## Motivation and Context
The goal is a dynamic UI for the Android client, where the UI is generated from a JSON schema. To support this, the configuration layer needs to be addressed first — establishing an uniform configuration interface that can also generate schemas and config files as needed.

Our cli config relied on `clap` and `twelf`, it is hard to have an uniform configuration interface with mobile clients.  The first step is to retire `twelf`,  which is no longer maintained and pulls in the legacy `serde_yaml` 0.8.23. The latest `serde_yaml` is 0.9.34, but it was itself deprecated in March 2024. To move forward cleanly, `serde-env` and `struct-patch` are introduced to replace `twelf` — it decouples the deserializer, provides a flexible way to overlay configs, and is easier to maintain.
 
Then, the default values are collected via the Default trait not `clap`, making it straightforward to instantiate any config and serialize it in any format. The `.schema.json` is generated and will be as part of the mobile build flow, giving the Android client a stable schema to drive its dynamic UI. This also unifies configuration input across the clients.

## How Has This Been Tested?
- [x] test cases added for json schema parser

## Types of changes

- [X] Security issue fix (non-breaking change which fixes an issue) 

## Checklist:

- [X] No need extra document changes


Cargo.toml
```diff
@@ -44,6 +44,7 @@ clap = { version = "4.4.7", features = ["derive"] }
 ctrlc = { version = "3.4.6", features = ["termination"] }
 delegate = "0.12.0"
 educe = { version = "0.6.0", default-features = false, features = ["Debug"] }
+envious = "0.2.2"
```
kp-mariappan-ramasamy: This seems to be unmaintained for last 2 years. Have you considered alternatives?
I found [serde-env](https://crates.io/crates/serde-env) which seems to have more downloads/update recently.

Any specific reason to choose this ?

kp-antonio-yang: I believe the issue is with nested structs. `envious` handles this well with the `PREFIX_MAIN_STRUCT__NEST_STRUCT` format — double underscore as a nesting separator is a thoughtful design choice since nested configs are common in practice. I may have tried `serde-env` before and hit this limitation.

The dependency footprint of `envious` is minimal — only `serde` and `thiserror` — which reflects how focused the crate is.  `serde-env` covers the same ground but brings in significantly more dependencies, making it harder to maintain and audit going forward.
That said, if you still prefer `serde-env`, I am happy to revisit it to see if it has caught up, or alternatively look into updating the `envious` dependency directly.

kp-mariappan-ramasamy: Thanks. Makes sense.

I am just worried since it is not maintained for 3 years similar to serde_yaml crate which we are trying to replace.
We might get a CVE for envious about being unmaintained later.

Looking from the serde-env example, i thought it supports nested struct.
https://crates.io/crates/serde-env

Do you remember what was the exact issue you encountered with it?

kp-antonio-yang: Sorry, wait a second, let me check on this.

lightway-client/Cargo.toml
```diff
@@ -34,20 +34,23 @@ bytesize.workspace = true
...
+struct-patch.workspace = true
 thiserror.workspace = true
-tokio.workspace = true
 tokio-stream = { workspace = true, features = ["time"] }
 tokio-util.workspace = true
+tokio.workspace = true
 tracing = { workspace = true, features = ["attributes"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
-twelf.workspace = true
+rtrb = { version = "0.3.3", optional = true }
```
kp-mariappan-ramasamy: This crate `rtrb` seems irrelevant to the PR. May be merge conflict artifact ?
I checked locally and could not find any reference

lightway-client/src/args.rs
kp-mariappan-ramasamy, line 1: nit! Typo in commit message `docments`, `valuses` of `server: client: config: cli docs from Patch struct`

Cargo.toml
kp-antonio-yang: Agreed, `serde-env` works well.

lightway-server/src/args.rs
```diff
@@ -128,3 +130,42 @@ pub struct Config {
...
+        Config {
+            config_file: PathBuf::default(),
+            mode: ConnectionType::Tcp,
+            user_db: None,
+            token_rsa_pub_key_pem: None,
+            server_cert: PathBuf::from("./server.crt"),
+            server_key: PathBuf::from("./server.key"),
+            tun_name: None,
+            ip_pool: Ipv4Net::new(Ipv4Addr::new(10, 125, 0, 0), 16)
+                .expect("default value should corret"),
```
kp-mariappan-ramasamy: Ipv4Net::new() is a const fn. You could do following to be clear instead of runtime panic

```
const IP_POOL: Ipv4Net = match Ipv4Net::new(Ipv4Addr::new(10, 125, 0, 0), 16) {
    Ok(v) => v,
    Err(_) => panic!("Invalid compile-time constant!"),
};
```

lightway-client/src/main.rs
```diff
@@ -89,7 +89,18 @@ fn generate_config(format: ConfigFormat, config_file: &PathBuf) -> Result<()> {
...
-        ConfigFormat::JsonSchema => todo!(),
+        ConfigFormat::JsonSchema => {
+            let settings = schemars::generate::SchemaSettings::draft07().with(|s| {
+                s.inline_subschemas = true;
+            });
+            let schema = settings.into_generator().into_root_schema_for::<Config>();
+            std::fs::write(
+                config_file,
+                serde_json::to_string_pretty(&schema)
+                    .expect("Some JsonSchema field did not implement properly"),
```
kp-mariappan-ramasamy: Lets handle this error properly by returning Err.
Same for all other panics

lightway-client/src/config.rs
kp-mariappan-ramasamy, line 1: Lets also rename server args.rs to config.rs to ne consistent

lightway-client/src/config.rs
```diff
@@ -294,18 +357,29 @@ impl Default for Config {
...
             wintun_file: None,
+            #[cfg(not(feature = "mobile"))]
             tun_local_ip: Ipv4Addr::new(100, 64, 0, 6),
+            #[cfg(feature = "mobile")]
+            tun_local_ip: Ipv4Addr::new(100, 64, 100, 2),
             tun_peer_ip: Ipv4Addr::new(100, 64, 0, 5),
+            #[cfg(not(feature = "mobile"))]
             tun_dns_ip: Ipv4Addr::new(100, 64, 0, 1),
+            #[cfg(feature = "mobile")]
+            tun_dns_ip: Ipv4Addr::new(100, 64, 100, 3),
```
kp-mariappan-ramasamy: Lets have a common default for noth mobile and desktop. If some mobile app need different requirement, they can change the default value.

lightway-client/src/config.rs
```diff
@@ -294,18 +357,29 @@ impl Default for Config {
...
             keepalive_continuous: true,
             tracer_packet_timeout: NonZeroDuration::from_std_duration(StdDuration::from_secs(10)),
+            #[cfg(not(feature = "mobile"))]
             preferred_connection_wait_interval: Duration::from_std_duration(
                 StdDuration::from_secs(0),
             ),
+            #[cfg(feature = "mobile")]
+            preferred_connection_wait_interval: Duration::from_std_duration(
+                StdDuration::from_millis(2000),
+            ),
```
kp-mariappan-ramasamy: Same as here. Lets use common value here.

kp-mariappan-ramasamy (approved):
LGTM, with the one comment. 

Using lightway-client to generate json config with `-g` seems little confusing. Would be better if have subcommands like

lightway-client generate
lightway-client connect -c config ...

kp-antonio-yang:
Yeh, we can do subcommand and better in other PR.   Thanks for the review.

PR #390, opened 2026-04-16
CVPN-2511: Process multiple packets at once in outside_io_task

kp-thomas-yau:
## Description

This PR further simplify batch receive packets logic by receiving multiple packets directly in outside_io_task without the need of ring buffer. Easier to maintain + more direct

## Motivation and Context

Simplify the logic of receiving multiple packet + solve weird tokio yielding issue where it might cause an performance degradation on other platforms (Linux)

## How Has This Been Tested?

Tested this newer macOS against the older client that uses the old batch_receiver implementation, both have similar performance in general:
| (in Mbps) | Speedtest Download | Speedtest Upload | iPerf3 Download | iPerf3 Upload |
|---|---|---|---|---|
| recvmsg_x old | 1497.29 | 557.54 | 1481 | 584 |
| recvmsg_x new | 1514.38 | 558.09 | 1480 | 567 |

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside.rs
+++ b/lightway-client/src/io/outside.rs
@@ -11,6 +11,10 @@ use async_trait::async_trait;
 use lightway_core::{IOCallbackResult, OutsideIOSendCallbackArg};
 use std::{net::SocketAddr, sync::Arc};
 
+/// Maximum number of packets to receive in a single batch syscall.
+#[cfg(batch_receive)]
+pub const BATCH_RECV_SIZE: usize = 32;
+
 #[async_trait]
 pub trait OutsideIO: Sync + Send {
     fn set_send_buffer_size(&self, size: usize) -> Result<()>;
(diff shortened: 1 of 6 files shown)


lightway-client/src/io/outside.rs
```diff
@@ -11,22 +11,33 @@ use async_trait::async_trait;
 use lightway_core::{IOCallbackResult, OutsideIOSendCallbackArg};
 use std::{net::SocketAddr, sync::Arc};
 
+/// Maximum number of packets to receive in a single batch syscall.
+pub const BATCH_RECV_SIZE: usize = 32;
```
xv-thomas-leong, line 16: nit:

```suggestion
#[cfg(batch_receive)]
pub const BATCH_RECV_SIZE: usize = 32;
```

lightway-client/src/lib.rs
```diff
@@ -384,35 +386,53 @@ pub async fn outside_io_task<ExtAppState: Send + Sync>(
...
+        let count = match outside_io.recv_buf(&mut bufs[0]) {
+            IOCallbackResult::Ok(_) => 1,
+            IOCallbackResult::WouldBlock => continue,
+            IOCallbackResult::Err(err) => return Err(err.into()),
+        };
+
+        // TODO: Batch send the packets into the tun device at once
+        {
+            let mut conn = conn.lock().unwrap();
+            for b in &mut bufs[..count] {
```
xv-thomas-leong, line 422: Not sure if it will make it better or worse performance wise, but maybe we can combine the for loop here and the one below?

xv-thomas-leong (dismissed):

lightway-client/src/lib.rs
kp-thomas-yau: I reckon not much difference here hmm, my logic here is just to minimise the time we hold the lock,

xv-thomas-leong: Make sense 👍

xv-thomas-leong (approved):

PR #394, opened 2026-04-21
CVPN-2535: client: log best connection_id and interface id as INFO

kp-weiwen-goh:
## Description

Log best connection ID and interface index with INFO level as structured log

## Motivation and Context

For programmatic parsing of connection ID after best connection selected, and interface index

## How Has This Been Tested?

Verify log printed

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -1129,6 +1129,7 @@ pub async fn client<
         tracing::info!(
             message = "Interface Details",
             %device_name,
+            if_index = inside_io.if_index().ok(),
             dns_ip = %config.tun_dns_ip,
             local_ip = %config.tun_local_ip,
             peer_ip = %config.tun_peer_ip,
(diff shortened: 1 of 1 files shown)


lightway-client/src/lib.rs
```diff
@@ -1166,7 +1167,10 @@ pub async fn client<
     };
     // connect_futs dropped here — releases &config borrow
 
-    tracing::trace!("Best connection selected: {best_connection_index}");
+    tracing::info!(
+        message = "Best connection selected",
+        connection_id = best_connection_index,
+    );
```
kp-antonio-yang, line 1173: What is the advantage from the log format?

---

I am not against this, but the one line log make the code base small and easier to read.

kp-antonio-yang (approved):

lightway-client/src/lib.rs
kp-weiwen-goh: This is to support structured JSON logging. When log_format is set to json, this will print in separate fields instead of one string.

PR #396, opened 2026-04-22
client: use feature gate for mobile

kp-antonio-yang:
## Description
It is easier to tailor and conditional build with feature gate than target

## Motivation and Context
Using the target alone for conditional compilation does not cover all use cases, so we keep the `mobile` feature gate. It is more explicit and gives developers finer control over what gets included in the build.

## How Has This Been Tested?
no need

## Types of changes

- [X] Fix downstream bug

## Checklist:


kp-weiwen-goh (approved):
LGTM

PR #397, opened 2026-04-22
Improve Windows Throughput

xv-thomas-leong:
## Description

Throughput on Windows is suffering due to small socket buffer size and small wintun ring capacity. Setting sensible defaults to improve performance.

Ref: wintun's [documentation](https://github.com/tun-rs/tun-rs/blob/0331939f5bf85cdfb4c27fbaba7bc20ca1cf7030/src/builder.rs#L463-L470) on ring buffer capacity

## How Has This Been Tested?

Manually tested.

Methdology: Run speed test (speedtest.net) 3 times and compute the average. 

Speed improvements (Lightway UDP, with ExpressLane Enabled):

On a laptop with Intel i7-1355U:

|        | Download (Mbps) | Upload (Mbps) | Latency (ms) | Losss (%) |
|--------|-----------------|---------------|--------------|-----------|
| Before | 895.71           | 585.38         | 2.32           | 0%      |
| **After**  | **985.22**          | **704.66**        | **1.96**          | **0.0**       |

On a VM with Intel Xeon Gold 6132 (4 vCPUs):

|        | Download (Mbps) | Upload (Mbps) | Latency (ms) | Losss (%) |
|--------|-----------------|---------------|--------------|-----------|
| Before | 405.5          | 415.07         | 3.68          | 0.17%      |
| **After**  | **740.27**          | **659.75**        | **4.11**          | **0.0**       |

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -52,6 +52,10 @@ pub struct TunConfig {
     #[cfg(windows)]
     /// Optional wintun file path for Windows TUN interfaces
     pub wintun_file: Option<String>,
+    #[cfg(windows)]
+    /// Wintun ring buffer capacity in bytes. Larger values improve throughput.
+    /// Must be a power of two between 128KiB and 64MiB.
+    pub ring_capacity: Option<u32>,
 }
 
 impl Debug for TunConfig {
@@ -147,6 +151,20 @@ impl TunConfig {
         self
     }
 
+    /// Set the wintun ring buffer capacity in bytes (Windows only).
+    /// Must be a power of two between 128KiB and 64MiB.
+    #[cfg(windows)]
+    pub fn ring_capacity(&mut self, capacity: u32) -> Result<&mut Self> {
+        const MIN: u32 = 128 * 1024;
+        const MAX: u32 = 64 * 1024 * 1024;
+        anyhow::ensure!(
+            capacity.is_power_of_two() && (MIN..=MAX).contains(&capacity),
+            "ring capacity must be a power of two between 128KiB and 64MiB, got {capacity}"
+        );
+        self.ring_capacity = Some(capacity);
+        Ok(self)
+    }
+
     /// Creates an async device based on TunConfig
     #[cfg(desktop)]
     pub fn create_as_async(&self) -> std::io::Result<AsyncDevice> {
@@ -155,8 +173,15 @@ impl TunConfig {
             builder = builder.name(name);
         }
         #[cfg(windows)]
-        if let Some(wintun_file) = self.wintun_file.as_ref() {
-            builder = builder.wintun_file(wintun_file.clone());
+        {
+            if let Some(wintun_file) = self.wintun_file.as_ref() {
+                builder = builder.wintun_file(wintun_file.clone());
+            }
+            if let Some(ring_capacity) = self.ring_capacity {
+                builder = builder.with(|opt| {
+                    opt.ring_capacity(ring_capacity);
+                });
+            }
         }
         #[cfg(macos)]
         {
--- a/lightway-client/src/args.rs
+++ b/lightway-client/src/args.rs
@@ -192,6 +192,13 @@ pub struct Config {
     #[clap(long)]
     pub tls_debug: bool,
 
+    /// Wintun ring buffer capacity in bytes (Windows only).
+    /// Must be a power of two between 128KiB and 64MiB.
+    /// Larger values improve throughput.
+    #[cfg(windows)]
+    #[clap(long, default_value = "8MiB")]
+    pub wintun_ring_capacity: ByteSize,
+
     /// Enable DPAPI encryption/decryption for config file
     /// Only for Windows platform
     #[cfg(windows)]
--- a/lightway-client/src/args.rs
+++ b/lightway-client/src/args.rs
@@ -117,11 +117,11 @@ pub struct Config {
     pub preferred_connection_wait_interval: Duration,
 
     /// Socket send buffer size
-    #[clap(long)]
-    pub sndbuf: Option<ByteSize>,
+    #[clap(long, default_value = "8MiB")]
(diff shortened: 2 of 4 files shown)


lightway-client/src/lib.rs
```diff
@@ -746,12 +746,23 @@ pub async fn connect<
             }
         };
 
+    // Windows default socket buffers are ~8KB, far too small for VPN throughput.
+    // Linux defaults to ~212KB and macOS to ~256KB. Set explicit defaults on Windows.
+    #[cfg(windows)]
+    const DEFAULT_SOCK_BUF: usize = 8 * 1024 * 1024;
```
kp-mariappan-ramasamy: This value will be hidden, if we set inside the library.

Might be better to set the default value in https://github.com/expressvpn/lightway/blob/31f84d5591b8ae230a27aa2e3b40ec213a9528e8/lightway-client/src/args.rs#L121 

This should also benefit other platforms by making this value as default.

kp-mariappan-ramasamy: How did we come up with this default ? 
Would be nicer to update the commit message if we experimented with other values too.

lightway-client/src/args.rs
```diff
@@ -192,6 +192,13 @@ pub struct Config {
     #[clap(long)]
     pub tls_debug: bool,
 
+    /// Wintun ring buffer capacity in bytes (Windows only).
+    /// Must be a power of two between 128KB and 64MB.
```
kp-mariappan-ramasamy: This restriction is from official WinTun driver I assume https://github.com/WireGuard/wintun#wintunstartsession

But the default value `10MB` is not a power of 2. So the driver either accepts it and adjusts internally or working unpredictably.

Better we change the value to proper .
Also we should use Mebibyte (MiB) which is power of 2, instead of Megabyte (MB) which represent power of 10.

So 8MiB may be appropriate ?

lightway-client/src/args.rs
```diff
@@ -192,6 +192,13 @@ pub struct Config {
...
+    /// Larger values improve throughput.
+    #[cfg(windows)]
+    #[clap(long, default_value = "10MB")]
+    pub wintun_ring_capacity: Option<ByteSize>,
```
kp-mariappan-ramasamy: If we provide default value, we could also make this as non-optional ?

lightway-app-utils/src/tun.rs
```diff
@@ -147,6 +151,14 @@ impl TunConfig {
         self
     }
 
+    /// Set the wintun ring buffer capacity in bytes (Windows only).
+    /// Must be a power of two between 128KB and 64MB.
+    #[cfg(windows)]
+    pub fn ring_capacity(&mut self, capacity: u32) -> &mut Self {
```
kp-mariappan-ramasamy: We could add a boundary check here since we are enforcing it

lightway-client/src/lib.rs
xv-thomas-leong: If we do that, that would mean we will be overriding system default values on mac and linux, which may require more testing to make sure there is no regression. I'd like to keep the scope focused on windows for now, but I'm open to other ways of implementing this too

xv-thomas-leong: I tested different values manually on my machine and settled on this one. Values bigger than that has no positive impact, but making it smaller would slow down the connection.

Will update the commit message about this

lightway-client/src/args.rs
xv-thomas-leong: Sure!

lightway-app-utils/src/tun.rs
xv-thomas-leong: Sure!

lightway-client/src/lib.rs
kp-mariappan-ramasamy: Yes correct, I would suggest to do that for all platforms now to avoid differences.
I think a quick testing in macos and linux should be enough for asserting this is fine.

lightway-client/src/args.rs
xv-thomas-leong: Good point, I was playing with different values and I guess I landed on one that isn't power of 2. Let me confirm if 8MiB make any difference in performance

xv-thomas-leong:
@kp-mariappan-ramasamy I will do some testing to make sure the performance is okay. Can you review my updated code in the meantime? 😄

kp-mariappan-ramasamy (approved):
LGTM

lightway-client/src/lib.rs
kp-thomas-yau: FYI, I have done some benchmark on macOS with 10Gbps, maybe this would be helpful:

<details><summary>Details</summary>
<p>

Buffer Configuration (Send/Receive KiB) | Avg Download Speed (gbps) | Avg Upload Speed (gbps)
-- | -- | --
System Default (9216/786896 Byte) | 1.0902 | 0.5236
512 / 512 | 1.0043 | 0.536
512 / 1024 | 1.0518 | 0.5364
512 / 2048 | 1.0732 | 0.5412
1024 / 1024 | 1.0476 | 0.5392
1024 / 2048 | 1.0889 | 0.5325
1024 / 4096 | 1.1334 | 0.5399
1536 / 1536 | 1.0727 | 0.5418
1536 / 3072 | 1.1149 | 0.5367
1536 / 6144 | 1.1855 | 0.5398
2048 / 2048 | 1.0742 | 0.5416
2048 / 4096 | 1.1423 | 0.54
2048 / 8192 | 1.1877 | 0.5351
4096 / 4096 | 1.2236 | 0.539
4096 / 8192 | 1.194 | 0.5419

</div><div class="pm-table-sticky-sentinel-bottom" data-testid="sticky-sentinel-bottom"></div></div></span></div></div></div></div>

</p>
</details>

xv-thomas-leong:
Thanks @kp-mariappan-ramasamy. I've done retest on all platforms and there is no regression. Merging this PR.

PR #399, opened 2026-04-23
Remove `batch_receive` from feature flag of client crate

kp-thomas-yau:
## Description

Remove this feature flag from `lightway-client`. Instead, we can use cfg_alias to define which platform explicitly supports batch_receive to guard this feature so that we don't need to build 2 binaries for testing.

## Motivation and Context

https://github.com/expressvpn/lightway/pull/384#discussion_r3128161536

## How Has This Been Tested?

Passed CI

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -299,9 +299,6 @@ jobs:
       - name: Run clippy
         run: nix develop --command cargo make clippy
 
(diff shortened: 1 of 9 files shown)


kp-mariappan-ramasamy (approved):
I was thinking to move as a run time feature. But this works too to select features based on platform.

LGTM

PR #402, opened 2026-04-24
client: add fixed GUID option for Wintun

kp-weiwen-goh:
## Description

Add optional `device_guid` field for Windows only, to specify the GUID for the tun. The device GUID is logged when set if specified.

## Motivation and Context

Using a stable GUID ensures that adapter creation retries reuse the same device node rather than leaking duplicates.

## How Has This Been Tested?

Tested that it sets the GUID and logs correctly.

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1646,6 +1646,7 @@ dependencies = [
  "tun-rs",
  "twelf",
  "uniffi",
+ "uuid",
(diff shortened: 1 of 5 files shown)


xv-thomas-leong (approved):
Tested it on my side too, LGTM

PR #408, opened 2026-04-29
CVPN-2528: client: Sighup reload config

kp-weiwen-goh:
## Description

BREAKING: `enable_inside_pkt_encoding_at_connect` arg is renamed to `enable_inside_pkt_encoding`

Add runtime reloadable config infrastructure for lightway-client. For now, only `enable_inside_pkt_encoding` is reloadable. The existing `enable_inside_pkt_encoding_at_connect` is renamed to `enable_inside_pkt_encoding` to be more accurate.

When a SIGHUP is received, lightway-client reloads the config from YAML and diffs it with the previous state. If any fields that are not reloadable are changed, it gives a warning.

CLI args and env variables will still take priority over the YAML config. (i.e. they will cause any config changes to fail if specified)

## Motivation and Context

Allows lightway configuration to be updated at runtime. For now, toggling inside pkt codec can be done by updating the config and sending a SIGHUP.

## How Has This Been Tested?

Tested that SIGHUP is properly handled, `enable_inside_pkt_encoding` field is properly parsed, and warn if other fields changed

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/main.rs
+++ b/lightway-client/src/main.rs
@@ -6,6 +6,7 @@ use anyhow::{Context, Result, anyhow};
 use futures::future::join_all;
 use lightway_core::{Event, EventCallback};
 use tokio::fs::read_to_string;
+use tokio::sync::mpsc;
 
 use lightway_app_utils::{
     TunConfig, Validate,
@@ -127,7 +128,7 @@ async fn main() -> Result<()> {
     config.apply(load_patch(&options, &config_file).await?);
     #[cfg(not(windows))]
     config.apply(serde_saphyr::from_str::<ConfigPatch>(
-        &read_to_string(config_file).await?,
+        &read_to_string(&config_file).await?,
     )?);
     config.apply(serde_env::from_env_with_prefix("LW_CLIENT")?);
     config.apply(options);
@@ -180,6 +181,8 @@ async fn main() -> Result<()> {
             tracing::warn!("Failed to send Ctrl-C signal: {err:?}");
         }
     })?;
+
+    let config_reload_signal = spawn_sighup_reload_handler(&config, config_file.clone());
 
     let inside_io: Option<Arc<dyn InsideIO<()>>> = None;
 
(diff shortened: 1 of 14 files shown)


lightway-client/src/main.rs
```diff
@@ -174,13 +177,32 @@ async fn main() -> Result<()> {
...
     ctrlc::set_handler(move || {
-        if let Some(Err(err)) = ctrlc_tx.take().map(|tx| tx.send(())) {
-            tracing::warn!("Failed to send Ctrl-C signal: {err:?}");
+        if let Some(tx) = ctrlc_tx_clone.lock().unwrap().take() {
+            let _ = tx.send(());
         }
     })?;
 
+    let config_reload_signal =
+        spawn_sighup_reload_handler(&config, config_file.clone(), env_patch, cli_patch);
```
xv-thomas-leong: minor bug: when `tun_name` is set in config, the first Sighup received will always show this warning

```
WARN lightway_client: Non-reloadable config fields changed (requires restart to take effect)
```

I did a diff when this happened and it showed `tun_name` as `None` initially, and after it re-reads the config, it has `Some(tunname)` set correctly. I suspect that's because we have `config.tun_name.take()` above

lightway-client/src/main.rs
```diff
@@ -174,13 +177,32 @@ async fn main() -> Result<()> {
...
+
+    // SIGTERM: ctrlc without `termination` only handles SIGINT
```
kp-mariappan-ramasamy: Since we have handler for SIGTERM, do we still need ctrlc library ?

Also how do we handle this for Windows ?

kp-weiwen-goh: ctrlc supports Windows, so we need to keep it. I can gate it to Windows only and use only sig handlers for unix.

Reloading for Windows is not supported here.

xv-thomas-leong (approved):

PR #391, opened 2026-04-16
Feat/unify initializing flow

kp-antonio-yang:
## Description
Mobile and CLI clients currently diverge in their state management, configuration structure, and naming conventions. This PR aligns the two, unifying the network initialization flow and establishing a consistent interface across both clients.

## Motivation and Context
- [x] LightwayClientConnectArgs -> ClientConnectionConfig
- [x] Collect event handlers as module
- [X] Collect mobile inside io from mobile
- [X] Collect mobile outside io from mobile
- [X] Normalize config when taking server configures
Following are the context and flow chart.
client-initialization-flow

## How Has This Been Tested?
Follow current test cases

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] The correct base branch is being used, if not `main`

PR #411, opened 2026-04-30
client: mobile: accept config content

kp-antonio-yang:
This PR splits the mobile interface commits out from #391, allowing the mobile project to iterate independently without being blocked by the ongoing internal config unification work.

## Description

- [x] Enriched Json Schema
- [x] Accept config content in mobile
- [x] Simplified command usage in the dev shell by automatically resolving the script location
- [X] Rename `EventHandlers`, which is also an interface in Koltlin
- [X] Unify auth fields in config (change `username` field under server to `user`, and the same field name in global level)
- [X] remove strong type input for mobile client

## Motivation and Context

By decoupling the mobile project from the lightway project, the mobile client can move faster and more easily adopt a dynamic UI.
This PR is split from #391. It is for 3M part of overall flow chart, and the meanwhile the 2M is dropped. 
client-initialization-flow

## How Has This Been Tested?

Following existing tests.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
  - Enriched JSON schema with more details about format, attributes, and examples
  - Modified mobile interface to accept string content in YAML config
  - Rename field user from username in ConnectionConfig (the config of servers)

## Checklist:


lightway-client/src/lib.rs
```diff
@@ -116,6 +116,8 @@ pub enum LightwayError {
     Unauthorized,
     #[error("Config Error: `{0}`")]
     ConfigError(#[from] crate::config::Error),
+    #[error("Config Format Error: `{0}`")]
+    ConfigFormatError(#[from] serde_saphyr::Error),
```
xv-thomas-leong, line 118: What's the difference between these 2 errors?

lightway-client/src/config.rs
```diff
@@ -25,6 +25,12 @@ use std::net::IpAddr;
 // values follow the Rust convention in Default trait, such that we are able
 // to serialized out any kind of configure from the Config::default(), such
 // that our library will be nice and easy to further do integrations.
+//
+// NOTE for JsonSchema
+// - Use feature gate not cfg, such that cli client can generate schema based
+// on features for other target if needed
```
xv-thomas-leong: Can you share more about this? It has always been possible to cross compile to other targets

lightway-client/src/lib.rs
kp-antonio-yang: - `ConfigFormatError` covers errors related to the format itself — for example, invalid indentation in YAML or unclosed `{}` in JSON(if we accept multiple style config input).
- `ConfigError` covers errors related to the value content — these are more domain-specific.

In short, `ConfigFormatError` is about structural/syntactic issues, while `ConfigError` is about semantic/domain validation.

lightway-client/src/config.rs
kp-antonio-yang: A Windows/macOS GUI client can still be cross-compiled from a Linux machine. In that case, the developer would run the client on Linux to generate the JSON schema, which would then be used for the Windows/macOS build. This is just a note to keep in mind for future cross-compilation scenarios.

xv-thomas-leong: Okay, but can you share more about why we should use feature gate instead of cfg? Do you mean we shouldn't use Rust's conditional compilation based on platform, but use cargo features for platform specific configs?

kp-antonio-yang: Yeh, `cfg` with target on functions are good and common in our repo, and I did not against it, so I just add a note on the struct for the `#[cfg(feature = ...)]`exception.

This is a exceptional note in for the `struct` to show the JSON schema.   Guide a developer need add fields and also considering to generate a schema for other platform.

lightway-client/src/config.rs
```diff
@@ -34,23 +41,27 @@ pub struct Config {
...
     #[serde(skip_serializing)]
+    #[schemars(skip)]
     pub config_file: PathBuf,
 
     /// Servers to attempt to connect to. Configuration is only supported in
     /// config file, not command line or environment variable
     #[patch(attribute(clap(skip)))]
     #[serde(default)]
     #[serde(skip_serializing)]
+    #[schemars(skip)]
```
xv-thomas-leong, line 48: Is this intentional for the `servers` field?

lightway-client/src/config.rs
```diff
@@ -25,6 +25,13 @@ use std::net::IpAddr;
 // values follow the Rust convention in Default trait, such that we are able
 // to serialized out any kind of configure from the Config::default(), such
 // that our library will be nice and easy to further do integrations.
+//
+// NOTE for JsonSchema
+// - Exceptionally use feature gate not cfg on Config struct, such that cli
+//   client can generate schema based on features for other target if needed
+// - `x-cfg` is an attr of field the config specified for the target, such that
+//   the client can optionally render UI
+// - Doc string will be the description for fields in JsonSchema
```
xv-thomas-leong: Suggested rewrite for clarity _(Awaiting agreement on the first point)_:

```suggestion
// JsonSchema conventions:                                          
// - Prefer `#[cfg(feature = "...")]` over target cfgs (e.g. `cfg(windows)`)                                                                                               
//   so CLI on any host can generate schema for other targets via cargo features.
// - `x-cfg` annotates the platform a field belongs to; UI uses it to decide rendering.                                                                                    
// - Field doc-comments become the schema `description`. 
```

lightway-client/src/config.rs
xv-thomas-leong: This is still unclear to me. Allow me to try to paraphrase:

```
Prefer `#[cfg(feature = "...")]` over target cfgs (e.g. `cfg(windows)`) 
so CLI on any host can generate schema for other targets via cargo features.
```

Did I capture what you're trying to say?

lightway-client/src/mobile.rs
```diff
@@ -147,15 +147,13 @@ impl RustVpnConnection {
         endpoints: Vec<crate::config::MobileConnectionConfig>,
         event_handler: Arc<dyn RustEventHandlers>,
         raw_tun_fd: i32,
-        mobile_config: crate::config::MobileConfig,
+        mobile_config: Option<crate::config::MobileConfig>,
```
xv-thomas-leong: Should we remove mobile_config entirely?

We don't have to consider backwards compatibility yet because we're developing both the library and the app end to end.

lightway-client/src/mobile.rs
```diff
@@ -170,7 +168,30 @@ impl RustVpnConnection {
...
+        if !config_content.is_empty() {
+            use struct_patch::Patch;
+            config.apply(serde_saphyr::from_str(&config_content)?);
+            config.servers.push(crate::config::ConnectionConfig {
+                server: config.server.clone(),
+                mode: config.mode.clone(),
+                server_dn: config.server_dn.take(),
+                cipher: config.cipher,
+                outside_mtu: config.outside_mtu,
+                username: config.user.take(),
```
xv-thomas-leong: > `username: config.user.take(),`

Probably not introduced in this PR, but can we add a commit to unify this?

nix/modules/devshells.nix
```diff
@@ -39,9 +39,11 @@
             };
             androidSdk = androidComposition.androidsdk;
             buildScript = pkgs.writeShellScriptBin "build" ''
+              cd $(git rev-parse --show-toplevel 2>/dev/null)
```
xv-thomas-leong: Bash best practice:

```suggestion
              cd "$(git rev-parse --show-toplevel 2>/dev/null)"
```

xv-thomas-leong (commented):
Some more comments

lightway-client/src/config.rs
kp-antonio-yang: Yes, it is intentional for the servers,  and also the servers will not public in next PR. #420

lightway-client/src/config.rs
kp-antonio-yang: Yes, the note is changed.

lightway-client/src/config.rs
kp-antonio-yang: Do we want to expose multiple servers on mobile at this stage? I would lean towards holding off for now, for a few reasons:

- Multiple server support has not been tested — only the single OSS server with one Android client has been manually verified.
- There are open design questions around multiple server config that need more discussion, for example: do we use a shared global auth and CA across all servers, or per-server credentials?  What is the way in mobile if we want to enable multiple by UI or by passing a yaml file?

Happy to do these in another PR.

lightway-client/src/config.rs
xv-thomas-leong: Okay, going back to my original question. Why do we want to use `#[cfg(feature = "...")]` over target cfgs (e.g. `cfg(windows)`) for platform specific configs?

lightway-client/src/config.rs
xv-thomas-leong: Sure, no problem. Just wanna check and make sure 👍

lightway-client/src/config.rs
kp-antonio-yang: A developer generating a JSON schema from the CLI does not need to be running on the target platform.

As a concrete example: a UI designer + React frontend developer (frontend fullstack common in start up teams) working on a Windows/Android client, but using macOS day-to-day, can simply run the CLI on his Mac with the relevant features enabled to generate the schema files as he need.   From there they can design and build the UI without needing access to a Windows machine or any other target-specific environment.

xv-thomas-leong: That's a niche use case, but it could be mitigated by including JSON schema of different platforms as CI artifacts so they can be downloaded.

The downside to using features instead of platform target configs is that we will create drifts between the schema and the compiled code, which would do more harm than good.

kp-antonio-yang: Theoretically, we could download artifacts from CI, run across multiple platforms, and store the JSON schema as artifacts. In practice however, this does not always hold up:

- The CLI cannot run on mobile clients, so there is no way to generate the JSON schema from an Android device. Putting the CLI on an Android device purely to produce a schema in the CI flow does not make sense. This also means there is no consistent approach to config and JSON schema generation if some fields are gated by target and others by feature.
- As mentioned in my earlier example, a fronted full stack developer cannot easily obtain the JSON schema for all platforms from a working feature branch on a fork repo without all platform supported CI.

There will likely be more edge cases along these lines. I believe keeping this simple makes the library significantly easier to use as an open source project, and these small non-ideal tradeoffs are worth tolerating.

The practical approach with the least friction for everyone is:

1. Feature gates should be on the **fields** of the config struct.
2. cfg target attributes are well suited for **functions**, as mentioned previously — just don't break the 1.

xv-thomas-leong: Good point about not being able to generate a schema on mobile via CLI. But I'm still confused about what you're proposing. Can you see if the following matches your intent? Please feel free to write if none of them captures exactly what you mean

#### A. Swap one Rust gate for another
- Replace every occurrence of `#[cfg(windows)]` with `#[cfg(feature = "windows")]`
- That would involve adding a `windows` Cargo feature. Schema is regenerated per feature combo on the host.

#### B. Drop the Rust gate on the field; use `x-cfg` only
- Target-varying fields stay unconditionally compiled; `#[schemars(extend("x-cfg" = "..."))]` carries the platform meaning for the UI. 
- One schema covers all platforms, generated once from any host. 
- Cargo features (`postquantum`, `debug`, etc.) keep their `#[cfg(feature = "...")]` gates.

kp-antonio-yang: **Give the biggest tailoring power to developer via schema from _ONE_ cli**

> B. Drop the Rust gate on the field _of Config_; use x-cfg only
> - Target-varying fields stay unconditionally compiled; #[schemars(extend("x-cfg" = "..."))] carries the > platform meaning for the UI.
> - One Schema cover all platforms generated from host which can run cli.
> - Cargo features (postquantum, debug, etc.) keep their #[cfg(feature = "...")] gates.

What is **biggest tailoring power**?

As an open source project, the greatest power means the lowest barrier — easiest to use, with details that are clear without requiring code changes:

- No code modifications or special machines needed. The schema can target any platform with the desired feature flags (postquantum, debug, etc.).
- All details are discoverable from the schema itself without reading source code — conveyed via x-cfg, format, and other JSON schema attributes.

Why does **tailoring power** matter?

By lowering down the integration efforts, developers can easy to integrate to their tools, then use it, eventually contribute back.  It is about interface, so it is consumer facing and it is spec.  As a spec of lib or an opensource project here, the consumer is developer.

---

Off-topic for this PR but related — my mental model of `feature` vs `cfg`:

There is a set of functions sharing the same signature, where `#[cfg(target)]` automatically selects the correct implementation at compile time. This means a Windows developer and an Android developer working in the same project speak almost the same domain language:
  
```rust
struct Config {
   #[cfg(feature="windows")]
   #[schemars(extend("x-cfg" = "windows"))]
   win_only_field: usize,
   
   #[cfg(feature="android")]
   #[schemars(extend("x-cfg" = "android"))]
   android_only_field: usize,
   // ...
}

fn main () {
   let config = Config::load();
   connect(config)
}

#[cfg(window)]
fn connect(config: Config) {
     let Config {
       win_only_field,
       ..
     } = config;
    
    if win_only_filed > 256 {
       // ...
    }
}

#[cfg(android)]
fn connect(config: Config) {
     let Config {
       android_only_field,
       ..
     } = config;
    let tun = Tun::new(android_only_field);
}

```

xv-thomas-leong: You lost me again. I have 2 questions for you and I would like you to answer them directly:

1. Does the current comment, prefer `#[cfg(feature = "...")]` over target cfgs (e.g. `cfg(windows)`), captures what you meant to say?
2. What's the primary technical reason behind your comment? You've listed several social reasons above, but I don't think they hold water.

kp-antonio-yang: 1. Yes, the preference is `#[cfg(feature = "...")]` over target-based gates on Config struct fields. Other target gates like `#[cfg(windows)]` are not touched — this PR is specifically about mobile accepting config content, and other targets are out of scope here. Once other clients start adopting JSON schema, their PRs can modify the fields they need, following the pattern established here and guided by the note.

2. It is true that there is no cli for mobile, and the mobile target needs a feature gate to generate the schema.

If there is no consensus on whether to prefer #[cfg(feature = "...")] over target gates as a general guide, we can leave that question open, remove the note for now, and let the mobile changes land as they are. We can revisit the broader convention in a follow-up once alignment is reached.

xv-thomas-leong (approved):

kp-antonio-yang:
Thanks for the reviews.  Will add document under /doc for this in #433

PR #412, opened 2026-05-04
CVPN-2519: Expresslane and InsidePktCodec mutually exclusive

kp-weiwen-goh:
## Description

Adds a `DataPathMode` enum that enforces that ExpressLane and InsidePktCodec are never both enabled.

ExpressLane cannot be enabled if `DataPathMode` is `InsidePktCodec`.

If ExpressLane is enabled when an encoding request is sent or received, ExpressLane is set to `Inactive` and is only re-enabled when encoding is then disabled.

## Motivation and Context

They are mutually exclusive. Previously, this was not enforced by code and relies on correct user configuration.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -78,6 +78,24 @@ const MAX_RETRANSMISSION_ATTEMPTS: u8 = 5;
 /// Maximum number of retransmissions attempts for each encoding request packet.
 const ENCODING_REQUEST_PKT_MAX_RETRANSMISSION_ATTEMPTS: u8 = 5;
 
+/// Which mutually exclusive data path optimization is active.
+///
+/// ExpressLane and InsidePktCodec cannot be active simultaneously.
+/// This enum is the single source of truth for which mode governs
+/// the data path.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum DataPathMode {
+    /// Standard data path (i.e. non-ExpressLane, non-InsidePktCodec)
+    #[default]
+    Standard,
+
+    /// ExpressLane
+    ExpressLane,
+
+    /// InsidePktCodec
+    InsidePktCodec,
+}
+
 /// Connection state
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[repr(u8)]
@@ -436,6 +454,9 @@ pub struct Connection<AppState: Send = ()> {
 
     // Expresslane state, config exchange, health monitoring, wire crypto, and callbacks
     expresslane: expresslane::Expresslane<AppState>,
+
+    /// Which mutually exclusive data path optimization is active
+    data_path_mode: DataPathMode,
 }
 
 /// Information about the new session being established with a new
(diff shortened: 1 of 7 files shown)


lightway-core/src/connection.rs
```diff
@@ -1688,12 +1688,28 @@ impl<AppState: Send> Connection<AppState> {
...
+            return;
+        }
         self.expresslane.state = new_state;
         self.event(Event::ExpresslaneStateChanged(new_state));
+
+        if new_state == ExpresslaneState::Active {
+            self.set_data_path_mode(DataPathMode::ExpressLane);
+        } else if self.data_path_mode == DataPathMode::ExpressLane {
+            self.set_data_path_mode(DataPathMode::Standard);
+        }
```
kp-antonio-yang, line 1720: May we use match here, such that compiler can help us to go through each state, and also future proof, when a new state comes in.
```rust
match (new_state, self.data_path_mode) {
  (_, _) if self.expresslane.state == new_state => return,
  (ExpresslaneState::Active, DataPathMode::InsidePktCodec) => {
    warn!("Blocking expresslane activation: InsidePktCodec is active");
    return;
  }
  (ExpresslaneState::Active, ) => {
    self.set_data_path_mode(DataPathMode::ExpressLane);
  }
  (, DataPathMode::ExpressLane) => {
    self.set_data_path_mode(DataPathMode::Standard);
  }
  (ExpresslaneState::Disabled, _)
    | (ExpresslaneState::WaitingForClient, _)
    | (ExpresslaneState::Inactive, _)
    | (ExpresslaneState::Degraded, _) => {}
}
self.expresslane.state = new_state;
self.event(Event::ExpresslaneStateChanged(new_state));

```

lightway-core/src/connection.rs
```diff
@@ -1879,14 +1856,21 @@ impl<AppState: Send> Connection<AppState> {
...
+        // with the peer, use that. Otherwise advertise our local max
         let version = match self.expresslane.data.version {
             ExpresslaneVersion::Unknown => ExpresslaneVersion::MAX,
             v => v,
         };
 
         self.expresslane.config_counter += 1;
         let config = wire::ExpresslaneConfig {
-            enabled: false,
+            enabled,
```
kp-antonio-yang, line 1898: nit:
Do we have any using case `send_expresslane_config(true, ExpresslaneKey::INVALID)` or `send_expresslane_config(false, key)`?

Why not pass a valid key to enable? 
```rust
ExpresslangConfig {
   enable: key.is_valid()
}
```

May we add `ExpresslaneKey::is_valid(&self) -> bool` and drop `ExpresslaneKey::is_invalid(&self) -> bool`?
The code will be less and strait forward.

kp-antonio-yang (changes requested):

lightway-core/src/connection.rs
kp-weiwen-goh: `enabled` can be false if `update_next_self_key()` fails, even if the key is valid. Then, we will send that expresslane should be enabled when it should not be. I don't think this is correct.

lightway-core/src/connection.rs
kp-weiwen-goh: I replaced the second branch with a match, but kept the first check to maintain the same order (set new state and emit event, then set data path).

This way, if there is a new state, it will cause a compile error.

kp-antonio-yang: Yeh, that is what I means,  If there is a new state, then compile error, and developer will take a look on the function, and make sure the logic is correct.  Do agree with this?

PR #413, opened 2026-05-04
CVPN-2346 Implement GSO offload on lightway-server

kp-samuel-tam:
## Description

Implement GSO on server side on DTLS and Expresslane, specifically on bulk server->client traffic. This consistently halves the total syscalls used during bulk transfers, and also improves aggregated server throughput by 2x for multiple clients doing transfers.

When `--enable-tun-offload` is set, the server reads TSO superpackets from the TUN with `IFF_VNET_HDR`, segments them in userspace, and emits each superpacket as a single `sendmsg(UDP_SEGMENT)` instead of N per-segment syscalls. On a single-flow iperf3 reverse test the kernel UDP send path collapses near-completely: `udp_sendmsg` 0.71% → ~0.05%, `sock_alloc_send_pskb` 2.61% → ~0.13%, `mlx5e_xmit` 1.88% → ~0%.

Trade-off: kernel work is replaced with userspace work (per-segment IP/TCP/UDP checksum recomputation, segment assembly). Kernel-side wins are clear and measurable; userspace cost is now the dominant factor.

Pacing: each `sendmsg(UDP_SEGMENT)` produces a NIC burst of up to N segments. This can exceed receiver socket buffer depth and increase tail drops at peak rates. We will need to revisit better TX pacing under congested links.

Future work will focus on compatibility with TUN backends like io_uring, GRO on the server side, and full GSO/GRO on client side, where single-flow workloads should see the biggest visible speedup not in this PR.

## Motivation and Context

See ticket CVPN-2346.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -316,9 +316,18 @@ impl Tun {
         }
     }
 
-    /// Raw read from `Tun`, returning the full virtio frame (header + payload).
+    /// Recv a GSO frame from `Tun` into `buf`, stripping and decoding the
+    /// leading `virtio_net_hdr`.
+    ///
+    /// On success `buf` holds the IP payload (header already advanced past)
+    /// and the returned tuple is `(buf.len(), hdr)`. Short reads and headers
+    /// that fail to decode are reported as [`IOCallbackResult::WouldBlock`] so
+    /// the caller's recv loop retries instead of treating them as hard errors.
     #[cfg(target_os = "linux")]
-    pub async fn recv_gso(&self, buf: &mut [u8]) -> IOCallbackResult<usize> {
+    pub async fn recv_gso(
+        &self,
+        buf: &mut BytesMut,
+    ) -> IOCallbackResult<(usize, lightway_core::VirtioNetHdr)> {
         match self {
             Tun::Direct(t) => t.recv_gso(buf).await,
             #[cfg(feature = "io-uring")]
@@ -433,18 +442,73 @@ impl TunDirect {
         }
     }
 
-    /// Raw read from Tun, returning the full virtio frame (header + payload).
+    /// Recv a GSO frame into `buf`. See [`Tun::recv_gso`] for the
+    /// buffer/result contract.
     #[cfg(target_os = "linux")]
-    pub async fn recv_gso(&self, buf: &mut [u8]) -> IOCallbackResult<usize> {
+    pub async fn recv_gso(
+        &self,
+        buf: &mut BytesMut,
+    ) -> IOCallbackResult<(usize, lightway_core::VirtioNetHdr)> {
+        use bytes::Buf;
+        use lightway_core::gso::VIRTIO_NET_HDR_LEN;
+
         let tun = self.tun.as_ref().unwrap();
-        match tun.recv(buf).await {
-            Ok(0) => IOCallbackResult::WouldBlock,
-            Ok(n) => IOCallbackResult::Ok(n),
+
+        // Read directly into the spare capacity. BytesMut's
+        // spare_capacity_mut returns &mut [MaybeUninit<u8>] so there's
+        // no zero-init pass on the hot path.
+        let spare = buf.spare_capacity_mut();
+        // SAFETY: `tun_rs::AsyncDevice::recv` takes `&mut [u8]` and forwards
+        // to `libc::read(2)`. The kernel only writes — it never dereferences
+        // userspace memory for reading — so handing it our uninitialized slab
+        // is sound at the syscall boundary. The unsoundness lives in *Rust*:
+        // constructing a `&mut [u8]` over uninitialized bytes is UB per strict
+        // aliasing rules, even if no one reads them. This cast is the only
+        // place we paper over that gap. Delete it once `tun-rs` exposes a
+        // `MaybeUninit`-aware recv.
+        #[allow(unsafe_code)]
+        let raw =
+            unsafe { std::slice::from_raw_parts_mut(spare.as_mut_ptr().cast::<u8>(), spare.len()) };
+
+        let n = match tun.recv(raw).await {
+            Ok(0) => return IOCallbackResult::WouldBlock,
+            Ok(n) => n,
             Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
-                IOCallbackResult::WouldBlock
+                return IOCallbackResult::WouldBlock;
             }
-            Err(err) => IOCallbackResult::Err(err),
+            Err(err) => return IOCallbackResult::Err(err),
+        };
+
+        if n <= VIRTIO_NET_HDR_LEN {
+            tracing::warn!(n, "tun recv_gso: read shorter than virtio header");
+            crate::metrics::tun_recv_gso_short_read();
+            // Discard the partial read (buf is untouched — no set_len)
+            // and return WouldBlock so the caller's recv loop retries
+            // immediately instead of treating this as a hard error.
+            return IOCallbackResult::WouldBlock;
         }
+
+        // SAFETY: the kernel wrote exactly `n` bytes into the spare
+        // slab; `n <= buf.capacity()` because the kernel wrote into a
+        // slice of that length.
+        #[allow(unsafe_code)]
+        unsafe {
+            buf.set_len(n);
+        }
+
+        // SAFETY for VirtioNetHdr::from_bytes: BytesMut is heap-backed
+        // and 8-byte aligned; `n > VIRTIO_NET_HDR_LEN` was just checked.
+        let hdr = match lightway_core::VirtioNetHdr::from_bytes(&buf[..VIRTIO_NET_HDR_LEN]) {
+            Ok(h) => *h,
+            Err(e) => {
+                tracing::warn!(?e, "tun recv_gso: virtio header decode failed");
+                buf.clear();
+                return IOCallbackResult::WouldBlock;
+            }
+        };
+        buf.advance(VIRTIO_NET_HDR_LEN);
+
+        IOCallbackResult::Ok((buf.len(), hdr))
     }
 
     /// Try write from Tun
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -49,6 +49,11 @@ pub struct TunConfig {
     #[cfg(unix)]
     #[educe(Default = true)]
     pub close_fd_on_drop: bool,
+    /// Enable TUN offload (`IFF_VNET_HDR`) so reads/writes carry a
+    /// `virtio_net_hdr` and the kernel performs GRO/GSO across the
+    /// device. Required for the GSO inside-IO path.
+    #[cfg(target_os = "linux")]
+    pub offload: bool,
     #[cfg(windows)]
     /// Optional wintun file path for Windows TUN interfaces
     pub wintun_file: Option<String>,
@@ -224,6 +229,10 @@ impl TunConfig {
             {
                 builder = builder.associate_route(false);
             }
+            #[cfg(target_os = "linux")]
+            if self.offload {
+                builder = builder.offload(true);
+            }
             let device = builder.build_async()?;
 
             if let Some(mtu) = self.mtu {
@@ -307,6 +316,18 @@ impl Tun {
         }
     }
 
+    /// Raw read from `Tun`, returning the full virtio frame (header + payload).
+    #[cfg(target_os = "linux")]
+    pub async fn recv_gso(&self, buf: &mut [u8]) -> IOCallbackResult<usize> {
+        match self {
+            Tun::Direct(t) => t.recv_gso(buf).await,
+            #[cfg(feature = "io-uring")]
+            Tun::IoUring(_) => {
+                IOCallbackResult::Err(std::io::Error::from(std::io::ErrorKind::Unsupported))
+            }
+        }
+    }
+
     /// Send a packet to `Tun`
     pub fn try_send(&self, buf: BytesMut) -> IOCallbackResult<usize> {
         match self {
@@ -363,6 +384,10 @@ pub struct TunDirect {
     fd: RawFd,
     #[cfg(unix)]
     close_fd_on_drop: bool,
+    /// `IFF_VNET_HDR` enabled — sends must be prefixed with a 12-byte
+    /// `virtio_net_hdr`, reads include it.
+    #[cfg(target_os = "linux")]
+    vnet_hdr: bool,
 }
 
 impl TunDirect {
@@ -385,6 +410,8 @@ impl TunDirect {
             fd,
             #[cfg(unix)]
             close_fd_on_drop: config.close_fd_on_drop,
+            #[cfg(target_os = "linux")]
+            vnet_hdr: config.offload,
         })
     }
 
@@ -406,10 +433,38 @@ impl TunDirect {
         }
     }
 
+    /// Raw read from Tun, returning the full virtio frame (header + payload).
+    #[cfg(target_os = "linux")]
+    pub async fn recv_gso(&self, buf: &mut [u8]) -> IOCallbackResult<usize> {
+        let tun = self.tun.as_ref().unwrap();
+        match tun.recv(buf).await {
+            Ok(0) => IOCallbackResult::WouldBlock,
+            Ok(n) => IOCallbackResult::Ok(n),
+            Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
+                IOCallbackResult::WouldBlock
+            }
+            Err(err) => IOCallbackResult::Err(err),
+        }
+    }
+
     /// Try write from Tun
     pub fn try_send(&self, buf: BytesMut) -> IOCallbackResult<usize> {
         let tun = self.tun.as_ref().unwrap();
-        match tun.try_send(&buf[..]) {
+        #[cfg(target_os = "linux")]
+        let res = if self.vnet_hdr {
+            // IFF_VNET_HDR requires a zeroed `virtio_net_hdr` prefix
+            // on every write (NEEDS_CSUM=0, GSO_NONE).
+            let hdr_len = tun_rs::VIRTIO_NET_HDR_LEN;
+            let mut prefixed = bytes::BytesMut::zeroed(hdr_len);
+            prefixed.extend_from_slice(&buf[..]);
+            tun.try_send(&prefixed[..])
+                .map(|n| n.saturating_sub(hdr_len))
+        } else {
+            tun.try_send(&buf[..])
+        };
+        #[cfg(not(target_os = "linux"))]
+        let res = tun.try_send(&buf[..]);
+        match res {
             Ok(nr) => IOCallbackResult::Ok(nr),
             Err(err) if matches!(err.kind(), std::io::ErrorKind::WouldBlock) => {
                 IOCallbackResult::WouldBlock
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -110,6 +110,13 @@ pub enum InvalidPacketError {
     /// Packet size greater than MAX_MTU
     #[error("Packet size greater than MAX_MTU")]
     InvalidPacketSize,
+
+    /// GSO superpacket failed validation (bad virtio_net_hdr,
+    /// per-segment build failure, etc.). The specific cause is
+    /// reflected in the `gso_dropped_*` and `gso_build_segment_failed`
+    /// metrics.
+    #[error("Invalid GSO superpacket")]
+    InvalidGsoPacket,
 }
 
 /// An error from an operation on a [`Connection`]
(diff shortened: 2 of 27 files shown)


lightway-core/src/plugin.rs
```diff
@@ -53,6 +53,11 @@ pub enum PluginResult {
 }
 
 impl PluginList {
+    #[cfg(target_os = "linux")]
```
kp-mariappan-ramasamy: This need not be a linux specific method. Looks generic

lightway-server/src/lib.rs
```diff
@@ -258,6 +264,146 @@ pub(crate) fn handle_inside_io_error(conn: Arc<Connection>, result: ConnectionRe
...
+    loop {
+        // Reset the window to start of slab (cheap; no-op while still
+        // at slab start, pointer-only reset after `advance()` has
+        // drifted us).
+        pkt.reserve(initial_cap);
+
+        // Expose the full slab to `recv_gso` as `&mut [u8]`.
+        // SAFETY: every byte of the slab was zero-initialized at
+        // construction; subsequent iters only ever shrunk `len` or
+        // overwrote bytes. We never hand out uninitialized memory.
```
kp-mariappan-ramasamy: This does not sound safe, as pkt is mutable we could also create new BytesMut and replace it.
So now if you reserve, it might be unintialized.

I think what you want is https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.spare_capacity_mut which gives pointer to spare buffer which you can sent to recv_gso

lightway-server/src/lib.rs
```diff
@@ -318,6 +468,39 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
...
+                            "link",
+                            "set",
+                            &name,
+                            "gso_max_size",
+                            &max_gso_buf_size.to_string(),
+                        ])
+                        .output();
+                }
+            }
+
```
kp-mariappan-ramasamy: Better to make this a config and move the responsibility of setting gso size, outside of server to avoid needing ADMIN capability

kp-mariappan-ramasamy: I see this is to reduce the default GSO size:

```
❮ ip -det link show dev lightway
2: lightway: <NO-CARRIER,POINTOPOINT,MULTICAST,NOARP,UP> mtu 1350 qdisc fq_codel state DOWN mode DEFAULT group default qlen 500
    link/none  promiscuity 0 allmulti 0 minmtu 68 maxmtu 65535
    tun type tun pi off vnet_hdr off persist on addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535 tso_max_size 65536 tso_max_segs 65535 gro_max_size 65536 gso_ipv4_max_size 65536 gro_ipv4_max_size 65536
❮ ip -det link show dev lightway
2: lightway: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1350 qdisc fq_codel state UP mode DEFAULT group default qlen 500
    link/none  promiscuity 0 allmulti 0 minmtu 68 maxmtu 65535
    tun type tun pi off vnet_hdr on persist on addrgenmode random numtxqueues 1 numrxqueues 1 gso_max_size 60300 gso_max_segs 65535 tso_max_size 65536 tso_max_segs 65535 gro_max_size 65536 gso_ipv4_max_size 60300 gro_ipv4_max_size 65536
```

let me go through all the changes

kp-samuel-tam: Yep, that's all that part is doing, to not get a "message too long" at sendmsg later

lightway-server/src/lib.rs
kp-samuel-tam: Because `spare_capacity_mut(&mut self) -> &mut [MaybeUninit<u8>]`, I think we ultimately still need a unsafe cast on the `&mut [MaybeUninit<u8>]` because `tun_rs::AsyncDevice::recv()` only takes `&mut [u8]` in the end. I will think about it.

Interestingly there's also `std::io::Read::read_buf` in the nightly std which takes `MaybeUninit<u8>` (which tun-rs could use) but it is never stabilized.

Edit: I changed recv signatures to accept `MaybeUninit` and do one unsafe cast in the end, I am hoping to drop it once tun-rs accepts `MaybeUninit`. Please take a look at  https://github.com/expressvpn/lightway/commit/e995ffa878c98f578b2a1baa44f2e6b20f41de4b 👀

lightway-core/src/plugin.rs
kp-samuel-tam: Failed lint because the function is unused on other platforms, needed to add back `#[allow(dead_code)]`.

lightway-app-utils/src/tun.rs
```diff
@@ -307,6 +316,18 @@ impl Tun {
...
     }
 
+    /// Raw read from `Tun`, returning the full virtio frame (header + payload).
+    #[cfg(target_os = "linux")]
+    pub async fn recv_gso(&self, buf: &mut [std::mem::MaybeUninit<u8>]) -> IOCallbackResult<usize> {
+        match self {
+            Tun::Direct(t) => t.recv_gso(buf).await,
+            #[cfg(feature = "io-uring")]
+            Tun::IoUring(_) => {
+                IOCallbackResult::Err(std::io::Error::from(std::io::ErrorKind::Unsupported))
```
kp-samuel-tam, line 335: I changed from `todo!()` to this.

lightway-app-utils/src/tun.rs
```diff
@@ -406,10 +433,51 @@ impl TunDirect {
...
+        // to `libc::read(2)`. The kernel only writes — it never dereferences
+        // userspace memory for reading — so handing it our uninitialized slab
+        // is sound at the syscall boundary. The unsoundness lives in *Rust*:
+        // constructing a `&mut [u8]` over uninitialized bytes is UB per strict
+        // aliasing rules, even if no one reads them. This cast is the only
+        // place we paper over that gap. Delete it (and revert this signature)
+        // once `tun-rs` exposes a `MaybeUninit`-aware recv.
+        #[allow(unsafe_code)]
+        let raw =
+            unsafe { std::slice::from_raw_parts_mut(buf.as_mut_ptr().cast::<u8>(), buf.len()) };
```
kp-samuel-tam: This is new to cast `MaybeUninit` away for `tun.recv()` which doesn't support `MaybeUninit` yet.

lightway-app-utils/src/tun.rs
```diff
@@ -406,10 +433,51 @@ impl TunDirect {
         }
     }
 
+    /// Raw read from Tun, returning the full virtio frame (header + payload).
+    #[cfg(target_os = "linux")]
+    pub async fn recv_gso(&self, buf: &mut [std::mem::MaybeUninit<u8>]) -> IOCallbackResult<usize> {
```
kp-samuel-tam: A lot of read/recv signatures are changed from `&mut [u8]` to `&mut [MaybeUninit<u8>]`.

lightway-app-utils/src/tun.rs
```diff
@@ -406,10 +433,51 @@ impl TunDirect {
...
-        match tun.try_send(&buf[..]) {
+        #[cfg(target_os = "linux")]
+        let res = if self.vnet_hdr {
+            // IFF_VNET_HDR requires a zeroed `virtio_net_hdr` prefix
+            // on every write (NEEDS_CSUM=0, GSO_NONE).
+            let hdr_len = tun_rs::VIRTIO_NET_HDR_LEN;
+            let mut prefixed = bytes::BytesMut::zeroed(hdr_len);
+            prefixed.extend_from_slice(&buf[..]);
+            tun.try_send(&prefixed[..])
+                .map(|n| n.saturating_sub(hdr_len))
```
kp-samuel-tam, line 525: This can be optimized to be `writev` but I'd defer it for now.

lightway-core/src/connection/io_adapter.rs
```diff
@@ -100,6 +100,16 @@ pub(crate) struct TlsIOAdapter {
     /// This buffer will be used to save the remaining data, to be sent in next call.
     pub(crate) send_buf: SendBuffer,
 
+    /// When `Some`, the wolfssl IO callback `send()` appends raw
+    /// encrypted segments here instead of sending to the socket.
+    /// After all segments are collected, `udp_send_gso` wraps each
+    /// with `wire::Header`, runs plugins, and sends via `send_gso`.
+    pub(crate) gso_buf: Option<BytesMut>,
```
kp-samuel-tam: This is not new but this is a buffer to "intercept" wolfssl doing send immediately after the packet is encrypted. Now when this is `Some(_)` it will store packets here and return early; for us to do GSO instead of shooting individual packets out right away.

Edit: A future plan is to do streaming/in-place encryption on the original superpacket, so this would not be needed... but deferred

lightway-core/src/connection/io_adapter.rs
```diff
@@ -205,6 +215,114 @@ impl TlsIOAdapter {
...
+        let hdr = wire::Header {
+            version: self.protocol_version,
+            aggressive_mode: false,
+            expresslane_data,
+            session: self.session_id,
+        };
+        let mut hdr_buf = BytesMut::with_capacity(wire::Header::WIRE_SIZE);
+        hdr.append_to_wire(&mut hdr_buf);
+
+        // Fast path: no outside plugins. Segments are not mutated, so
```
kp-samuel-tam, line 356: Why there are a fast path and a slow path - is because how plugins are designed. Plugins always need to see the packet in one consecutive buffer, but our fast path doesn't do that because it separates header construction and payload itself, so we can avoid a copy on the payload.

lightway-core/src/connection.rs
```diff
@@ -985,6 +985,132 @@ impl<AppState: Send> Connection<AppState> {
...
+        // No MTU check — GSO superpacket is intentionally oversized
+        if !ipv4_is_valid_packet(pkt.as_ref()) {
+            return Err(InvalidInsidePacket(InvalidIpv4Packet));
+        }
+
+        let _ = self.rotate_expresslane_key();
+
+        // No tcp_clamp_mss — GSO packets are never SYN
+
+        // Plugins see the whole superpacket as one packet
```
kp-samuel-tam, line 1032: This might break plugins because they will be exposed to GSO superpackets, but I don't think this can be avoided... But they should still see the usual transport headers for routing decisions so personally I am not too worried.

lightway-core/src/connection.rs
```diff
@@ -985,6 +985,132 @@ impl<AppState: Send> Connection<AppState> {
...
+                break;
+            }
+        }
+
+        if result.is_ok() {
+            self.activity.last_data_traffic_from_peer = Instant::now();
+            match self.session.io_cb_mut().udp_send_gso(n, expresslane) {
+                IOCallbackResult::Ok(_) | IOCallbackResult::WouldBlock => {}
+                IOCallbackResult::Err(e) => {
+                    tracing::error!(error = %e, segs = n, "udp_send_gso failed");
```
kp-samuel-tam: I promoted this from `tracing::warn` to `tracing::error`.

Edit: reverted this rookie mistake, this will flood Sentry, abort abort abort 🚨✋

lightway-core/src/plugin.rs
```diff
@@ -53,6 +53,11 @@ pub enum PluginResult {
 }
 
 impl PluginList {
+    #[allow(dead_code)]
```
kp-samuel-tam, line 56: Changed from `#[cfg(target_os = "linux")]`.

lightway-server/src/io/outside/udp.rs
```diff
@@ -48,32 +49,45 @@ impl std::fmt::Display for BindMode {
 
 fn send_to_socket(
```
kp-samuel-tam, line 51: Not new, but did additional refactoring on `fn send_to_socket`, now `.with_control()` is unconditional but kernel handles a `cmsg_len = 0` cmsg just fine.

lightway-server/src/io/outside/udp.rs
```diff
@@ -96,7 +110,24 @@ struct UdpSocket {
 impl OutsideIOSendCallback for UdpSocket {
     fn send(&self, buf: &[u8]) -> IOCallbackResult<usize> {
         let peer_addr = self.peer_addr.read().unwrap();
-        send_to_socket(&self.sock, buf, &peer_addr.1, self.reply_pktinfo)
+        send_to_socket(
+            &self.sock,
+            &[IoSlice::new(buf)],
```
kp-samuel-tam, line 116: Not new, but this is changed `buf -> &[IoSlice::new(buf)]` because `send_to_socket` signature changed for `iovec` aka scatter/gather IO.

lightway-server/src/lib.rs
```diff
@@ -276,6 +282,139 @@ pub(crate) fn handle_inside_io_error(conn: Arc<Connection>, result: ConnectionRe
     }
 }
 
+async fn inside_io_loop_default(
```
kp-samuel-tam, line 287: This is pulled directly from the `inside_io_loop` `tokio::spawn` code and unchanged.

lightway-server/src/lib.rs
```diff
@@ -276,6 +282,139 @@ pub(crate) fn handle_inside_io_error(conn: Arc<Connection>, result: ConnectionRe
...
+            let result = conn.inside_data_received(&mut buf);
+            handle_inside_io_error(conn, result);
+        } else {
+            metrics::tun_rejected_packet_no_connection();
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+async fn inside_io_loop_gso(
```
kp-samuel-tam, line 324: This is the GSO replacement for original `inside_io_loop` to handle the `vnet_hdr` and other GSO stuff.

tests/server/docker-entrypoint.sh
```diff
@@ -14,6 +14,10 @@ echo "====================================================================="
 
 ip tuntap add mode tun dev "${tunname}"
 ip link set dev "${tunname}" mtu 1350
+# Cap GSO superpacket intake. The default 65536 plus Lightway + TLS/expresslane
+# headers per segment can exceed what sendmsg(UDP_SEGMENT) accepts, causing the
+# kernel to return EMSGSIZE ("Message too long") on the outside socket.
+ip link set dev "${tunname}" gso_max_size 60300
 ip link set dev "${tunname}" up
```
kp-samuel-tam, line 21: This is new to cap `gso_max_size` to not blow up `sendmsg(UDP_SEGMENT)` later. We also need this in deployment scripts because it's not in the lightway code anymore.

kp-samuel-tam (commented):
Some updates and comments to help with reviews.

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,650 @@
...
+
+/// Build segment `gso_idx` from the superpacket into `out`.
+///
+/// Resets `out` and writes header template + payload slice into its
+/// spare capacity, applies all per-segment fixups (IP ID, TCP seq,
+/// checksums), then commits the segment via `set_len`. On return,
+/// `out` holds exactly the one segment's wire bytes.
+///
+/// `out.capacity()` must be ≥ one segment's maximum wire length.
+pub(crate) fn build_segment(
```
kp-samuel-tam, line 301: I added some new tests below for this function - this is run to construct a complete, checksum-computed packet for wolfssl to encrypt. Could be written in IO scatter/gather style and avoid the copies if we use the wolfssl streaming/in-place encrypt APIs.

lightway-core/src/connection/io_adapter.rs
```diff
@@ -205,6 +215,114 @@ impl TlsIOAdapter {
...
+                // Drop semantics on the GSO path are all-or-nothing:
+                // skipping a middle segment would break UDP_SEGMENT's
+                // uniform stride requirement.
+                PluginResult::Drop | PluginResult::DropWithReply(_) => {
+                    return IOCallbackResult::Ok(tun_buf.len());
+                }
+                PluginResult::Error(e) => {
+                    return IOCallbackResult::Err(Error::other(e));
+                }
+            }
```
kp-thomas-yau, line 392: So if i understand correctly, if only one of the packets here gets rejected by the plugin, it will just drop the `tun_buf` here and none of segments here will be sent out? Should we skip the dropped packets (or cache the error from the plugin if we've received an error here) and only return the result after we sent out all the segments?

kp-mariappan-ramasamy:
I tried to give it a go in my dev environments. 
With namespace based dev setup, I am getting the following errors when running with tun-offload. Still the traffic passes. 
We raise all error alerts to sentry. So in general, we will not have any error log in data path which will flood sentry.
Please use warn for datapath errors.

```
2026-06-02T08:10:04.394214Z ERROR lightway_core::connection: udp_send_gso failed error=Message too large (os error 90) segs=49
2026-06-02T08:10:04.590807Z ERROR lightway_core::connection: udp_send_gso failed error=Message too large (os error 90) segs=50            2026-06-02T08:10:04.802131Z ERROR lightway_core::connection: udp_send_gso failed error=Message too large (os error 90) segs=49
2026-06-02T08:10:05.058434Z ERROR lightway_core::connection: udp_send_gso failed error=Message too large (os error 90) segs=50
2026-06-02T08:10:05.216928Z ERROR lightway_core::connection: udp_send_gso failed error=Message too large (os error 90) segs=48
```

And then I tried to test in 2 VMs, where server panic'd with the following after starting the iperf3 test.
```
thread 'main' (324242) panicked at lightway-core/src/gso.rs:263:37:
range end index 2648 out of range for slice of length 1350
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Inside IO loop panicked

Caused by:
    task 7 panicked with message "range end index 2648 out of range for slice of length 1350"
```
I set the tun gso properly similar to earthly script. Debugging whether it is my setup issue.
in case if you face this while dev, let me know.

lightway-core/src/connection/io_adapter.rs
kp-samuel-tam: Great catch, I overlooked this because of how I did the inside plugins; it should just work here because they are encrypted, individual packets already. Changed.

kp-samuel-tam:
Thank you all for the reviews 🙌 I have updated the PR but mistakenly rebased on a wrong branch, so now even if I fixed it, the force-push `Compare` button above will show a mess.

To make the review easier, please see https://github.com/expressvpn/lightway/compare/39cf489..057ded1 for changes since May 28. You can verify the two commit hashes and history with 
```sh
curl -s "https://api.github.com/repos/expressvpn/lightway/issues/413/events?per_page=100" | jq '[.[] | select(.event == "head_ref_force_pushed") | {created_at, actor: .actor.login, commit_id}]'
```

---

And for the crash that @kp-mariappan-ramasamy raised, that is on me. That was not a setup issue but a subtle bug in the `gso::build_segment` code that I should have caught in tests, but missed when iterating the PR. It is now addressed in: https://github.com/expressvpn/lightway/blob/cd00ef84521608cbbe2dbe45fe0a18b53a6a580b/lightway-core/src/gso.rs#L195-L200

It's subtle but well-documented that kernel breaks the virtio-net spec on `hdr_len`; in our case `vhdr.hdr_len = 1350`, and it _cannot_ be trusted or used as the protocol header length. The same issue was surfaced in a different way, see
- kernel source: https://github.com/torvalds/linux/blob/6f3ed7fec72fc8979b2a8c7219c0a9fcfc8d07b5/include/linux/virtio_net.h#L254-L256
- wireguard-go: https://github.com/WireGuard/wireguard-go/blob/ecfc5a8d54462e18e13c72173e2623d16d8e25a0/tun/tun_linux.go#L417-L420
- tun-rs: https://github.com/tun-rs/tun-rs/blob/744e41c18a2c90cc1b87149952a1e4f013e1c564/src/platform/linux/device.rs#L625-L628

lightway-core/src/connection.rs
kp-mariappan-ramasamy: There can be some plugins to reject/dropwithreply individual packet, ex dns requests. This will break that.

lightway-core/src/connection/io_adapter.rs
```diff
@@ -205,6 +215,115 @@ impl TlsIOAdapter {
...
+    ) -> IOCallbackResult<usize> {
+        use std::io::{Error, IoSlice};
+
+        // Consume the accumulated encrypted segments. Leaves
+        // `gso_buf = None` so GSO mode is off until the next batch
+        // explicitly re-enables it.
+        let tun_buf = self
+            .gso_buf
+            .take()
+            .expect("udp_send_gso called without gso_buf");
```
kp-mariappan-ramasamy: We could handle this gracefully instead of panic'ing ?

lightway-server/src/lib.rs
```diff
@@ -398,39 +548,28 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
...
+            ))
         }
-    });
+        #[cfg(not(target_os = "linux"))]
+        tokio::spawn(inside_io_loop_default(
+            inside_io,
+            ip_manager.clone(),
+            config.lightway_client_ip,
+        ))
+    };
```
kp-mariappan-ramasamy, line 541: You can do the following to avoid redundant _default tasks:
```rust
let inside_io_loop: JoinHandle<anyhow::Result<()>> = {
    #[cfg(target_os = "linux")]
    let gso = config.enable_tun_offload;
    #[cfg(not(target_os = "linux"))]
    let gso = false;

    if gso {
        tokio::spawn(inside_io_loop_gso(
            inside_io,
            ip_manager.clone(),
            config.lightway_client_ip,
        ))
    } else {
        tokio::spawn(inside_io_loop_default(
            inside_io,
            ip_manager.clone(),
            config.lightway_client_ip,
        ))
    }
};
```

lightway-server/src/lib.rs
```diff
@@ -277,6 +282,139 @@ pub(crate) fn handle_inside_io_error(conn: Arc<Connection>, result: ConnectionRe
...
+        let hdr = match VirtioNetHdr::from_bytes(&pkt[..VIRTIO_NET_HDR_LEN]) {
+            Ok(hdr) => *hdr,
+            Err(err) => {
+                tracing::warn!("Failed to decode virtio header: {err}");
+                pkt.clear();
+                continue;
+            }
+        };
+        // Strip the virtio header — `pkt` is now the IP payload.
+        pkt.advance(VIRTIO_NET_HDR_LEN);
```
kp-mariappan-ramasamy: I think these lines can be moved inside recv_gso and make the api return hdr and len along with bytesMut

```
async fn recv_gso(&self, buf: &mut BytesMut) -> IOCallbackResult<(usize, VirtioNetHdr)>;
```

Better to keep unsafe code inside the library abstractions instead of from main library code

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,702 @@
...
+///
+/// Linux's TUN driver writes `skb_headlen` (a hint about linearity) into
+/// `virtio_net_hdr.hdr_len` — NOT the protocol header length the virtio-net
+/// spec calls for. For multi-segment GSO aggregates the linearity hint is
+/// roughly the size of the first segment (≈ MTU), not the headers, so any
+/// code that copies a per-segment header template based on `vhdr.hdr_len`
+/// will get a wildly wrong value. Parse the real length from the packet.
+///
+/// Returns `None` for truncated buffers or protocols we don't decode.
+pub(crate) fn calc_hdr_len(pkt: &[u8]) -> Option<usize> {
```
kp-mariappan-ramasamy: This function better return Result<usize> so that caller knows exactly what is the issue

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,702 @@
...
+                return None;
+            }
+            let doff = (pkt[ip_hdr_len + TCP_DATA_OFFSET_BYTE] >> 4) as usize * 4;
+            if doff < TCP_MIN_HDR_LEN {
+                return None;
+            }
+            doff
+        }
+        IPPROTO_UDP => UDP_HDR_LEN,
+        _ => return None,
```
kp-mariappan-ramasamy: I think we could simply by using pnet_packet Packet abstractions without too low level access

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,702 @@
...
+    };
+    Some(ip_hdr_len + l4_hdr_len)
+}
+
+/// Number of segments in a GSO superpacket.
+pub(crate) fn calc_gso_segs(pkt_len: usize, hdr_len: usize, gso_size: usize) -> usize {
+    let payload_len = pkt_len.saturating_sub(hdr_len);
+    payload_len.div_ceil(gso_size)
```
kp-mariappan-ramasamy, line 287: This fn will panic if gso_size is zero, we need to make sure gso_size is not zero here.
Handle this properly

lightway-core/src/connection.rs
```diff
@@ -985,6 +985,166 @@ impl<AppState: Send> Connection<AppState> {
...
+        if result.is_ok() {
+            self.activity.last_data_traffic_from_peer = Instant::now();
+            match self.session.io_cb_mut().udp_send_gso(gso_segs, expresslane) {
+                IOCallbackResult::Ok(_) | IOCallbackResult::WouldBlock => {}
+                IOCallbackResult::Err(e) => {
+                    tracing::warn!(error = %e, gso_segs, "udp_send_gso failed");
+                }
+            }
+        } else {
+            self.session.io_cb_mut().gso_buf.take();
```
kp-mariappan-ramasamy: We need to clear this gso_buf whether it is success or failure, is not it ?
So it hsuld be outside of if else case ?

lightway-core/src/connection.rs
```diff
@@ -985,6 +985,166 @@ impl<AppState: Send> Connection<AppState> {
...
+        // initializes bytes 0..total_len through `MaybeUninit` and
+        // returns the prefix it wrote; we only ever expose that prefix
+        // via `set_len`. Same shape as `inside_io_loop_gso`'s recv buf.
+        let mut segment = BytesMut::with_capacity(mtu);
+
+        // Enable GSO wire buffer — IO callback will append here.
+        // Both DTLS and expresslane paths detect this and append
+        // encrypted segments instead of sending immediately.
+        self.session.io_cb_mut().gso_buf =
+            Some(BytesMut::with_capacity(gso_segs * self.outside_mtu));
```
kp-mariappan-ramasamy: This could be a function `prepare_gso_buf` in io_adapter to prevent leaking abstraction.

lightway-core/src/connection.rs
kp-mariappan-ramasamy: Also same as `prepare_gso_buf`api.
Could add a function `finalize_gso_buf` or something to remove the current gso_buf and reset size

lightway-core/src/connection/io_adapter.rs
```diff
@@ -291,7 +410,19 @@ impl crate::tls::IOCallbacks for TlsIOAdapter {
...
-            ConnectionType::Datagram => self.udp_send(buf, false),
+            ConnectionType::Datagram => {
+                if let Some(gso_buf) = self.gso_buf.as_mut() {
+                    // GSO: buffer raw encrypted data; udp_send_gso
+                    // will wrap each segment with wire::Header later.
+                    if gso_buf.is_empty() {
+                        self.gso_size = buf.len();
+                    }
+                    gso_buf.extend_from_slice(buf);
+                    IOCallbackResult::Ok(buf.len())
```
kp-mariappan-ramasamy: We can move this part inside self.udp_send() so expresslane handling in connection.rs::send_expresslane_data() does not need to handle this separately.

lightway-core/src/connection.rs
```diff
@@ -985,6 +985,166 @@ impl<AppState: Send> Connection<AppState> {
...
+
+        // Parse the protocol header length from the packet itself. We
+        // can't trust `hdr.hdr_len` — Linux's TUN puts `skb_headlen`
+        // (~MTU for multi-segment aggregates) there, not the protocol
+        // header length the virtio-net spec calls for.
+        let Some(hdr_len) = gso::calc_hdr_len(pkt.as_ref()) else {
+            tracing::warn!(
+                pkt_len = pkt.len(),
+                "unable to parse hdr_len from superpacket"
+            );
```
kp-mariappan-ramasamy: We would need metrics for these error cases to track in production. We will not have logs

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,702 @@
...
+const TCP_DATA_OFFSET_BYTE: usize = 12;
+const UDP_HDR_LEN: usize = 8;
+
+fn is_v6(pkt: &[u8]) -> bool {
+    pkt[0] >> 4 == 6
+}
+
+fn is_tcp(hdr: &VirtioNetHdr) -> bool {
+    hdr.gso_type == VIRTIO_NET_HDR_GSO_TCPV4 || hdr.gso_type == VIRTIO_NET_HDR_GSO_TCPV6
+}
```
kp-mariappan-ramasamy: This should be an associated method of VirtioNetHdr like VirtioNetHdr::is_tcp()

lightway-core/src/gso.rs
```diff
@@ -0,0 +1,702 @@
...
+/// Resets `out` and writes header template + payload slice into its
+/// spare capacity, applies all per-segment fixups (IP ID, TCP seq,
+/// checksums), then commits the segment via `set_len`. On return,
+/// `out` holds exactly the one segment's wire bytes.
+///
+/// `hdr_len` is the real header length the caller derived once via
+/// [`calc_hdr_len`] for the whole superpacket.
+///
+/// `out.capacity()` must be ≥ one segment's maximum wire length.
+pub(crate) fn build_segment(
```
kp-mariappan-ramasamy, line 301: I think it is possible to write this function using safe rust code using pnet_packet crate. 

The function writes header, segment, then udpate ip id, checksum. Should be possible to do it with BytesMut update and then pnet_packet::IpPacket

Is it any specific reason not to use safe code here?

kp-mariappan-ramasamy:
There are 2 other issues i found:

1. The dev namespace environment is broken. iperf3 fails from client to remote if i start the server without tun offload. PTAL. dev env will be created by sudo ./tests/setup.sh. Refer repo README
2. Cargo.lock file needs some update. Rebuild and push to push lock file changes

lightway-core/src/connection.rs
kp-samuel-tam: I would say current presentation is faithful, that is, we are reflecting what we receive from TUN (but now without the vnet_hdr so it doesn't totally break plugins that look at TCP/IP headers). The reject individual packet part is technically achievable if you drop some segments from the superpacket and `Accept` it. I agree that it's not compatible with the current plugin interface, and would like to raise a new PR on a redesign (also for I/O scatter/gather, if possible) but would like to keep it as is for now. What do you think?

kp-mariappan-ramasamy: Yes okie, leave this for now, with a TODO on the trait about this GSO quirk

lightway-app-utils/src/tun.rs
```diff
@@ -263,6 +286,103 @@ impl TunConfig {
...
+/// device-wide, so the subsequent real attach (without `IFF_VNET_HDR`)
+/// sees the kernel deliver raw IP packets again.
+///
+/// Best-effort: errors are logged and swallowed.
+#[cfg(target_os = "linux")]
+fn clear_tun_offloads(name: &str) {
+    use std::ffi::CString;
+
+    // _IOW('T', 202, int) — same as tun-rs uses for TUNSETIFF.
+    const TUNSETIFF: libc::c_ulong = 0x400454ca;
```
kp-mariappan-ramasamy: Use libc::Ioctl type. Otherwise it fails for other arch

Example: Compiling for musl fails with this error
```
   Compiling heck v0.4.1
error[E0308]: mismatched types
    --> lightway-app-utils/src/tun.rs:362:39
     |
 362 |     let rc = unsafe { libc::ioctl(fd, TUNSETIFF, &mut req as *mut IfReqShort) };
     |                       -----------     ^^^^^^^^^ expected `i32`, found `u64`
     |                       |
     |                       arguments to this function are incorrect
     |
note: function defined here
    --> /home/maari/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/unix/linux_like/mod.rs:1731:20
     |
1731 |             pub fn ioctl(fd: c_int, request: Ioctl, ...) -> c_int;
     |                    ^^^^^
help: you can convert a `u64` to an `i32` and panic if the converted value doesn't fit
     |
 362 |     let rc = unsafe { libc::ioctl(fd, TUNSETIFF.try_into().unwrap(), &mut req as *mut IfReqShort) };
     |                                                ++++++++++++++++++++

error[E0308]: mismatched types
    --> lightway-app-utils/src/tun.rs:378:39
     |
 378 |     let rc = unsafe { libc::ioctl(fd, TUNSETOFFLOAD, std::ptr::null::<libc::c_int>()) };
     |                       -----------     ^^^^^^^^^^^^^ expected `i32`, found `u64`
     |                       |
     |                       arguments to this function are incorrect
     |
note: function defined here
    --> /home/maari/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.186/src/unix/linux_like/mod.rs:1731:20
     |
1731 |             pub fn ioctl(fd: c_int, request: Ioctl, ...) -> c_int;
     |                    ^^^^^
help: you can convert a `u64` to an `i32` and panic if the converted value doesn't fit
     |
 378 |     let rc = unsafe { libc::ioctl(fd, TUNSETOFFLOAD.try_into().unwrap(), std::ptr::null::<libc::c_int>()) };
     |                                                    ++++++++++++++++++++

   Compiling oneshot-uniffi v0.1.6
   Compiling signature v2.2.0
```

kp-samuel-tam: Thanks, will fix 🙌

lightway-server/src/io/inside/tun.rs
```diff
@@ -53,6 +57,52 @@ impl InsideIORecv for Tun {
...
+                buf.clear();
+                return IOCallbackResult::WouldBlock;
+            }
+        };
+        buf.advance(VIRTIO_NET_HDR_LEN);
+
+        // Note: payload bytes (post-virtio-strip), not raw kernel
+        // bytes — see metrics::tun_to_client doc.
+        metrics::tun_to_client(buf.len());
+        IOCallbackResult::Ok((buf.len(), hdr))
```
kp-mariappan-ramasamy: This is a low level processing. It should be moved inside app_utils::tun::recv_gso

lightway-app-utils/src/tun.rs
```diff
@@ -263,6 +286,103 @@ impl TunConfig {
...
+/// `TUNSETOFFLOAD` is `_IOW('T', 208, c_int)` and is rejected (EINVAL)
+/// on fds that don't have `IFF_VNET_HDR` set. We therefore open a
+/// short-lived probe fd that attaches to the existing TUN *with*
+/// `IFF_VNET_HDR`, issue the clear, and close it. The mask is
+/// device-wide, so the subsequent real attach (without `IFF_VNET_HDR`)
+/// sees the kernel deliver raw IP packets again.
+///
+/// Best-effort: errors are logged and swallowed.
+#[cfg(target_os = "linux")]
+fn clear_tun_offloads(name: &str) {
```
kp-mariappan-ramasamy: This does not look like proper way to clear it. We can create a issue in tun crate to reset the headers. Also we can fork the repo and use it meanwhile??

kp-samuel-tam: Merged upstream and dropped here: https://github.com/tun-rs/tun-rs/pull/141

kp-mariappan-ramasamy:
@kp-samuel-tam Did you test the local namespace based env. 
I am facing an issue where switching between enable-run-offload/ disable makes disable mode failure to transfer traffic ?

Please let me know if you test the following cases:
- In local dev env, run without offload, do a iperf test, then with offload, iperf test and repeat without the offload to make sure offload does not impact the orig environment.
- Repeat the same tests using two VMs

lightway-app-utils/src/tun.rs
```diff
@@ -307,6 +316,69 @@ impl Tun {
...
+        // and 8-byte aligned; `n > VIRTIO_NET_HDR_LEN` was just checked.
+        let hdr = match lightway_core::VirtioNetHdr::from_bytes(&buf[..VIRTIO_NET_HDR_LEN]) {
+            Ok(h) => *h,
+            Err(e) => {
+                tracing::warn!(?e, "tun recv_gso: virtio header decode failed");
+                buf.clear();
+                return IOCallbackResult::WouldBlock;
+            }
+        };
+        buf.advance(VIRTIO_NET_HDR_LEN);
```
kp-mariappan-ramasamy: Can you move this implementation even further down to TunDirect ?

kp-samuel-tam:
@kp-mariappan-ramasamy I bumped tun-rs 2.5.1 -> 2.8.5 in this branch, which should resolve the issue properly now. Tested the offload off → on → off cycle in both real wire and local dev env.

kp-mariappan-ramasamy (approved):
Thanks for the PR. LGTM

PR #416, opened 2026-05-05
CVPN-2515: TUN `recv_buf` to take in BytesMut

kp-thomas-yau:
## Description

We currently keep allocating a new BytesMut whenever we run `recv_buf`. Updating it to take in a BytesMut from the outer-loop instead, so that we can just clear + resize BytesMut whenever we wish to receive a packet.

## Motivation and Context

Reduce unnecessary memory allocation + unify/interface between this and batch receiving/sending packets to TUN

## How Has This Been Tested?

Passed CI

Some very minor speed improvements as well, on a M2 macBook Pro with a 10Gbps line:
Download: 1469.32 Mbps -> 1505.16 Mbps
Upload: 536.96 Mbps -> 534.32 Mbps (so similar speed here, nothing much)

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -509,9 +509,13 @@ pub async fn inside_io_task<ExtAppState: Send + Sync>(
             .unwrap_or(DEFAULT_TRACER_TRIGGER_TIMEOUT)
     };
     let mut tracer_timeout_last_outside_data_rcvd: Option<Instant> = None;
+    let mtu = inside_io.mtu();
+    let mut buf = BytesMut::with_capacity(mtu);
     loop {
-        let mut buf = match inside_io.recv_buf().await {
-            IOCallbackResult::Ok(buf) => buf,
+        buf.clear();
+        buf.resize(mtu, 0);
+        match inside_io.recv_buf(&mut buf).await {
+            IOCallbackResult::Ok(_n) => {}
             IOCallbackResult::WouldBlock => continue, // Spuriously failed to read, keep waiting
             IOCallbackResult::Err(err) => {
                 // Fatal error
--- a/lightway-server/src/lib.rs
+++ b/lightway-server/src/lib.rs
@@ -20,6 +20,7 @@ pub use lightway_core::{
 pub type ServerEventCbType = Arc<dyn Fn(SessionId, &Event) + Send + Sync>;
 
 use anyhow::{Context, Result, anyhow};
+use bytes::BytesMut;
 use ipnet::Ipv4Net;
 use lightway_app_utils::{PacketCodecFactoryType, TunConfig, connection_ticker_cb};
 use lightway_core::{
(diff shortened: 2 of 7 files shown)


lightway-client/src/lib.rs
```diff
@@ -509,9 +509,13 @@ pub async fn inside_io_task<ExtAppState: Send + Sync>(
             .unwrap_or(DEFAULT_TRACER_TRIGGER_TIMEOUT)
     };
     let mut tracer_timeout_last_outside_data_rcvd: Option<Instant> = None;
+    let mtu = inside_io.mtu();
+    let mut buf = BytesMut::with_capacity(mtu);
     loop {
-        let mut buf = match inside_io.recv_buf().await {
-            IOCallbackResult::Ok(buf) => buf,
+        buf.clear();
```
kp-samuel-tam, line 515: Not sure if `buf.clear()` is needed. For me just truncate and resize is enough on GSO.

lightway-server/src/lib.rs
```diff
@@ -369,9 +370,13 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
     };
 
     let inside_io_loop: JoinHandle<anyhow::Result<()>> = tokio::spawn(async move {
+        let mtu = inside_io.mtu();
+        let mut buf = BytesMut::with_capacity(mtu);
         loop {
-            let mut buf = match inside_io.recv_buf().await {
-                IOCallbackResult::Ok(buf) => buf,
+            buf.clear();
```
kp-samuel-tam, line 376: same same

lightway-server/src/lib.rs
```diff
@@ -369,9 +370,13 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
...
+            buf.resize(mtu, 0);
+            match inside_io.recv_buf(&mut buf).await {
+                IOCallbackResult::Ok(_n) => {}
```
kp-samuel-tam, line 379:
```suggestion
                IOCallbackResult::Ok(_) => {}
```

kp-samuel-tam (commented):
alloc -> reuse is gucci

lightway-server/src/lib.rs
kp-thomas-yau: I think this is just a matter of preference as well, just to acknowledge that we're receiving byte size

lightway-client/src/lib.rs
kp-thomas-yau: I would say it's a matter of preference? In theory `tun-rs` would not read what we provide but returning a bytesMut with a previous packet here is not ideal imo

lightway-server/src/lib.rs
kp-samuel-tam: Then it's fine

lightway-client/src/lib.rs
kp-samuel-tam: The way it's coded is already "not ideal", might as well drop the decoration. `buf.clear()` is just `self.set_len(0)` and later `buf.resize(mtu)` is `self.set_len(mtu)` so the data never left.

Also zeroing bytes is very expensive on the hot path, I will trade dirty for speed anytime.

kp-thomas-yau: > buf.resize(mtu) is self.set_len(mtu)

this is wrong, if the length is set to zero via `clear()` beforehand, resize will then overwrite the entire BytesMut (spare_capacity_mut returns a pointer that starts at ptr + self.len)
```
        let dst = self.spare_capacity_mut().as_mut_ptr();
        // SAFETY: `spare_capacity_mut` returns a valid, properly aligned pointer and we've
        // reserved enough space to write `additional` bytes.
        unsafe { ptr::write_bytes(dst, value, additional) };
```

kp-samuel-tam: Right, my bad, re-read again and you are right. The point on expensive zeroing still holds, but that's up for debate.

kp-samuel-tam (approved):

PR #417, opened 2026-05-05
CVPN-2549: Centralise TLS imports

kp-shawn-thiah:
Introduce lightway-core::tls as the single re-export point for TLS types used across the workspace. In-tree crates now import TLS types from crate::tls instead of from wolfssl crate directly, keeping TLS-related imports in one place.


lightway-app-utils/src/tracing.rs
kp-thomas-yau, line 1: Do other libraries provides a tracing callback with the same interface as well? If WolfSSL's ones are different, I think we might want to keep this name

kp-mariappan-ramasamy (approved):
LGTM

PR #419, opened 2026-05-08
Bump wolfssl-rs to 7.0.0

kp-shawn-thiah:
## Description
In 7.0.0, wolfssl-rs exposes a safe rust API (install_logging_callback) to register the tracing callback function and automatically enables debugging
- updated relevant functions to use the new signature for setting logging callbacks

## How Has This Been Tested?
Ran with debug feature and checked logs


lightway-client/src/lib.rs
```diff
@@ -847,7 +847,7 @@ pub async fn connect<
 
     #[cfg(feature = "debug")]
     if config.tls_debug {
-        set_logging_callback(Some(wolfssl_tracing_callback));
+        set_logging_callback(wolfssl_tracing_callback);
```
kp-mariappan-ramasamy: I think we could remove the file tracing.rs completely and simplify to:

```suggestion
        set_logging_callback(|m: &str| tracing::debug!(m));
```

Prev I added as a file since it has unsafe code, now it does not looks like it has any purpose.

kp-thomas-yau: nit: I think we can also set `target: "ssl_debug"` (or other name) in the debug event as well

kp-shawn-thiah: removed `wolfssl_tracing.rs` and added `target: "ssl_debug"` in debug event

kp-mariappan-ramasamy (approved):

PR #420, opened 2026-05-11
client: move auth and ca on connection level config

kp-antonio-yang:
## Description
- [X] add auth and ca_content on `ClientConnectionConfig`
- [X] add `Config::take_servers` to normalize `ConnectionConfig` with auth and cert information
- [X] rm `auth` and `root_ca_cert` of `ClientConfig`
- [X] doc for [Connection Setting](https://github.com/expressvpn/lightway/blob/feat/auth-in-endpoint/docs/connection_setting.md)

## Motivation and Context
Currently, auth and ca settings are in global config level in desktop clients, but the mobile client used in connection level.  The input of ca_cert can be a content or a path of ca cert file in desktop clients.

It is possible to have different auth and ca with multiple servers in desktop, so we move these in `ConnectionConfig`, also provide a `take_servers` method and keep `servers` private to load the ca content and normalize it.

This PR is split from #391(the pink red arrow part of overall flow chart), and now it is documented under [/doc/connection_setting.md](https://github.com/expressvpn/lightway/blob/feat/auth-in-endpoint/docs/connection_setting.md). 

## How Has This Been Tested?
Follow current test case

## Types of changes

- [X] Refactoring with internal config struct change

## Checklist:

### DIFF
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -46,7 +46,7 @@ pub struct Config {
     #[serde(default)]
     #[serde(skip_serializing)]
     #[schemars(skip)]
-    pub servers: Vec<ConnectionConfig>,
+    servers: Vec<ConnectionConfig>,
 
     #[patch(attribute(clap(short, long)))]
     #[patch(attribute(doc = r#"Server to connect to in `<hostname>:<port>` format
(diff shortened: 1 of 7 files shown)


lightway-client/src/config.rs
```diff
@@ -314,30 +314,68 @@ pub struct Config {
...
+            self.servers.len() + 1
+        }
     }
 
-    /// Try build CA from ca_crt
-    pub fn load_ca(&self) -> Result<lightway_core::tls::RootCertificate<'_>, Error> {
-        load_ca(&self.ca_cert)
+    /// Check if there is any server setting
+    pub fn is_empty(&self) -> bool {
+        self.servers.is_empty() & self.server.is_empty()
```
kp-weiwen-goh:
```suggestion
        self.servers.is_empty() && self.server.is_empty()
```

lightway-client/src/config.rs
```diff
@@ -314,30 +314,68 @@ pub struct Config {
...
+        let ca_content = if check_cert_header(&self.ca_cert) {
+            Some(self.ca_cert.clone())
+        } else {
+            // NOTE: we support a path input on desktop
+            if cfg!(desktop) {
+                std::fs::read_to_string(&self.ca_cert).ok()
+            } else {
+                None
+            }
+        };
```
kp-weiwen-goh, line 354: This logic is repeated in a different way below. I think we can keep the `load_ca(cert_or_path) -> Result<cert>` and put the logic in there

kp-antonio-yang: There is a little different between these, so we can not make a load_ca for the Config and also for the ConnectionConfig.

We accept the global config without CA, so the `std::fs::read_to_string(&self.ca_cert).ok()`  and for the connection level the error should raise `*ca_cert = std::fs::read_to_string(&mut *ca_cert).map_err(|_| Error::CaFileNotFound)?`

kp-antonio-yang: More details note added to point out the difference.

kp-weiwen-goh (approved):

kp-antonio-yang:
Thanks for reviews.

PR #425, opened 2026-05-14
CVPN-2559: Handle expresslane flags and session id tampering

kp-mariappan-ramasamy:
## Description

**Changes:**
- Add expresslane flags in AAD to detect the flags tampering and update expresslane version to v2
- Make the Protocol Version mismatch as nonfatal since it is a plaintext header
- Make expresslane keys rotation interval as configurable
- Make connection expiration interval and stats report interval also as configurable

## Motivation and Context

Since expresslane flags are not included in AAD, a onpath attacker can modfiy expresslane flags without being detected and close the connection. We got notified by a researcher about this bug.

Also the protocol version in UDP lightway packets can be tampered by on-path attacker and make server close the client connection. Ignore protocol version error too since it is a plaintext header.

## How Has This Been Tested?

Added unit test to verify this tampering

## Types of changes

- [x] Breaking change. This would break compat with existing clients. But we have not deploy any clients yet, also clients will fallback to DTLS.

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -1796,11 +1796,18 @@ impl<AppState: Send> Connection<AppState> {
         // If updating key failed, send disabled to peer
         let enabled = self.expresslane.data.update_next_self_key(key).is_ok();
 
+        // Outgoing version: if we have already negotiated a version
+        // with the peer, use that. Otherwise advertise our local max
+        let version = match self.expresslane.data.version {
+            ExpresslaneVersion::Unknown => ExpresslaneVersion::MAX,
+            v => v,
+        };
+
         self.expresslane.config_counter += 1;
         let config = wire::ExpresslaneConfig {
             enabled,
             key,
-            version: ExpresslaneVersion::MAX,
+            version,
             ack: false,
             counter: self.expresslane.config_counter,
         };
@@ -1826,11 +1833,16 @@ impl<AppState: Send> Connection<AppState> {
         let key = ExpresslaneKey::INVALID;
         let _ = self.expresslane.data.update_next_self_key(key);
 
+        let version = match self.expresslane.data.version {
+            ExpresslaneVersion::Unknown => ExpresslaneVersion::MAX,
+            v => v,
+        };
+
         self.expresslane.config_counter += 1;
         let config = wire::ExpresslaneConfig {
             enabled: false,
             key,
-            version: ExpresslaneVersion::MAX,
+            version,
             ack: false,
             counter: self.expresslane.config_counter,
         };
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -465,6 +465,7 @@ struct NewConnectionArgs<AppState> {
     expresslane: bool,
     expresslane_cb: Option<expresslane::ExpresslaneCbType<AppState>>,
     expresslane_metrics: Option<expresslane::ExpresslaneMetricsType>,
+    expresslane_keys_rotation_interval: std::time::Duration,
 }
 
 impl<AppState: Send> Connection<AppState> {
@@ -528,6 +529,7 @@ impl<AppState: Send> Connection<AppState> {
                 expresslane_state,
                 args.expresslane_cb,
                 args.expresslane_metrics,
+                args.expresslane_keys_rotation_interval,
             ),
         };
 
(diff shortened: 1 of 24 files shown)


lightway-core/src/connection.rs
```diff
@@ -1786,11 +1786,18 @@ impl<AppState: Send> Connection<AppState> {
         // If updating key failed, send disabled to peer
         let enabled = self.expresslane.data.update_next_self_key(key).is_ok();
 
+        // Outgoing version: if we have already negotiated a version
+        // with the peer, use that. Otherwise advertise our local max
+        let version = match self.expresslane.data.version {
+            ExpresslaneVersion::Unknown => ExpresslaneVersion::MAX,
+            v => v,
+        };
```
kp-thomas-yau, line 1804: should we add a warning trace if we have negotiated a lower expresslane version?

lightway-core/src/connection.rs
```diff
@@ -270,8 +264,10 @@ impl ConnectionError {
                     WireError(wire::FromWireError::InsufficientData) => false,
                     WireError(wire::FromWireError::InvalidExpressData) => false,
                     WireError(wire::FromWireError::ReplayedExpressData) => false,
+                    WireError(wire::FromWireError::InvalidProtocolVersion(..)) => false,
                     WireError(_) => true,
 
+                    InvalidProtocolVersion => false,
                     InvalidState => false, // Can be due to out of order or repeated messages
                     InvalidInsideIo => false, // Can be used for test only test control plane
                     UnknownSessionID => false,
```
kp-thomas-yau, line 273: I'm a bit concerned about this one - does this mean older Lightway client can now connect to a newer server without any issue in the future if we bumped Lightway version to a newer one?

lightway-core/src/connection.rs
kp-thomas-yau: Seems like if we've received a new config from peer and we have negotiated a lower version here, we're still sending an acknowledgement packet with expresslane version unchanged? Should we modify the config version after negotitaion()?

lightway-core/src/connection.rs
kp-mariappan-ramasamy: Yes, it is a general rule to make server backward compatible with client. Even now, without expresslane older clients 1.2 will still be able to connect to 1.3 server.
This change should not impact that, and is targeted to ignore tampered protocol version.

I did not understand your exact concern though, can you explain a bit more

lightway-core/src/connection.rs
kp-mariappan-ramasamy: Nice catch! Yes will update it.
Even without update, it currently works cause server again send config message with nego version and client updates it. But would be better to update in ACK itself since it is already known.

lightway-core/src/connection.rs
kp-mariappan-ramasamy: Added info logs/ This is with older v1 clients

```
2026-05-18T04:03:00.033941Z  INFO lightway_server::connection_manager: State changed for 192.168.0.2:38437 session=e119ab6a62f7cb56 state=Online
2026-05-18T04:03:00.034132Z  INFO lightway_core::connection: Negotiated expresslane version: Version1 [max:Version2 peer:Version1]
2026-05-18T04:03:00.034399Z  INFO lightway_server::connection_manager: Setting expresslane state to Inactive
2026-05-18T04:03:00.034562Z  INFO lightway_server::connection_manager: Setting expresslane state to Active
```

lightway-core/src/wire/expresslane_config.rs
```diff
@@ -4,18 +4,54 @@ use bytes::{Buf, BufMut, BytesMut};
...
-    Unknown = 0,
     #[default]
+    // Not used
+    Unknown = 0,
+    // Initial Expresslane format
     Version1 = 1,
+    /// same wire layout as V1, but the expresslane flags field
+    /// is bound into the AEAD AAD. Incompatible with V1 builds
+    Version2 = 2,
+}
```
kp-thomas-yau, line 25: nit: Can mark this as `non_exhaustive`

kp-thomas-yau (approved):

lightway-core/src/wire/expresslane_config.rs
kp-mariappan-ramasamy: Incrementing the version implies breaking change.

Marking it as non-exhaustive will make downstream lib users to have catch all case, which might hide issues. I would prefer to keep it as exhaustive so that new version is visible.

Note that is the compile time thing, so it will not impact existing binaries. Only newer builds will have to handle the new version.

PR #426, opened 2026-05-18
CVPN-2576 Fix missing postquantum cfg on tests

kp-shawn-thiah:
## Description
connection.rs referenced `KeyShare` and `with_pq_crypto` unconditionally, but both are gated on postquantum feature in the crate, as a result it fails to compile -- `cargo test -p lightway-core`

This was missed in CI because postquantum was enabled.

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


xv-thomas-leong (approved):
Can we have a test in CI such that the same issue can be detected in the future?

PR #427, opened 2026-05-18
CVPN-1597 Reduce fragment map size

kp-shawn-thiah:
## Description
Previously max size was 65536 (uint16), which can take up to 8GB of memory per connection, assuming maximum fragment size. Reduced to 4000.

## Types of changes

## 

### DIFF
--- a/lightway-core/src/connection/fragment_map.rs
+++ b/lightway-core/src/connection/fragment_map.rs
@@ -20,7 +20,7 @@ pub(crate) enum FragmentMapResult {
 pub(crate) struct FragmentMap(LruCache<u16, FragmentedPacket>);
 
 impl FragmentMap {
-    pub(crate) const DEFAULT_MAX_ENTRIES: NonZeroU16 = NonZeroU16::MAX;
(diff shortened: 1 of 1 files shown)


lightway-core/src/connection/fragment_map.rs
```diff
@@ -20,7 +20,7 @@ pub(crate) enum FragmentMapResult {
 pub(crate) struct FragmentMap(LruCache<u16, FragmentedPacket>);
 
 impl FragmentMap {
-    pub(crate) const DEFAULT_MAX_ENTRIES: NonZeroU16 = NonZeroU16::MAX;
+    pub(crate) const DEFAULT_MAX_ENTRIES: NonZeroU16 = NonZeroU16::new(4000).unwrap();
```
kp-mariappan-ramasamy: Its a Hashmap underneath. Could better be 2**x value like 4096 ?

kp-shawn-thiah: okies, updated to 4096.

kp-mariappan-ramasamy (approved):

PR #428, opened 2026-05-18
tun: Use given tunnel fd on desktop platform

kp-thomas-yau:
## Description

Use the given tunnel FD on desktop as well

## Motivation and Context

On macOS (and other desktop unix targets), TunConfig::create_as_async() previously ignored self.fd entirely and always called DeviceBuilder::new().build_async(), which creates a new TUN device from scratch. This requires elevated privileges or specific entitlements and fails with "operation not supported" in contexts where the TUN device is pre-created externally (e.g. a macOS Network Extension).

## How Has This Been Tested?

Tested this change on macOS network extension, verified we're able to use the tunnel device created by network extension properly.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -12,7 +12,7 @@ use std::{
     net::{IpAddr, Ipv4Addr},
 };
 
-#[cfg(mobile)]
+#[cfg(unix)]
 use std::os::fd::FromRawFd;
 #[cfg(feature = "io-uring")]
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


lightway-app-utils/src/tun.rs
```diff
@@ -183,6 +183,17 @@ impl TunConfig {
...
+        // If a fd was provided (e.g., macOS Network Extension), wrap it directly
+        // instead of creating a new TUN device, which would require elevated privileges.
+        #[cfg(unix)]
+        if let Some(fd) = self.fd {
+            // SAFETY: The caller must ensure `fd` is a valid TUN device file descriptor
+            // and transfer exclusive ownership to this function. The AsyncDevice will
+            // properly close the fd when dropped (unless close_fd_on_drop is false).
+            #[allow(unsafe_code)]
+            return Ok(unsafe { tun_rs::AsyncDevice::from_raw_fd(fd) });
+        }
```
kp-mariappan-ramasamy: Guess we could merge `create_as_async` into one function for both mobile and desktop and return error only for mobile if there is no fd

kp-thomas-yau: updated

kp-mariappan-ramasamy (approved):
LGTM

PR #432, opened 2026-05-20
CVPN-109: Token auth with version

kp-mariappan-ramasamy:
## Description

Add a new auth type TokenWithVersion which has lightway versoin in addition to token in auth message.

Note that for now, i just added the support for server. We will update the clients to use this auth once all servers are updated.

## Motivation and Context

CVPN-109

With UDP lightway, server knows clients lightway version from Wire::Header.
But there is no way for tcp client to negotiate Lightway version with server, since this header is not present in TCP.

Add lightway version with AuthRequest message, so server knows client lightway version on first message.

## How Has This Been Tested?

Added nuit/integration tests for new auth method

## Types of changes

## Checklist:


kp-weiwen-goh (approved):
LGTM

PR #433, opened 2026-05-21
doc: document unified Config design, override flow, and lifecycle

kp-antonio-yang:
## Description

Add a new [`docs/general_config.md`](https://github.com/expressvpn/lightway/blob/doc/client-config/docs/general_config.md) (easy to read with the link) document covering how client configuration is determined, with a Mermaid flow diagram, guidance for introducing new config fields, and inline descriptions of the three key config methods:

- `validate()` — checks the resolved `Config` for conflicts and invalid values (warns on sndbuf/rcvbuf/enable_pmtud on non-macOS TCP; errors on invalid `device_guid` UUID or out-of-range `wintun_ring_capacity`).
- `take_servers()` — normalizes the server list, promotes single-server top-level fields into `Vec<ConnectionConfig>`, resolves `ca_cert`, and propagates shared auth credentials.
- `try_from_reload_sig_and_config()` — the canonical `ClientConfig` constructor; builds `TunConfig` and maps all remaining `Config` fields.

`config.rs` in both `lightway-client` and `lightway-server` now carries a module-level doc link pointing to the new document. `docs/design_overview.md` is updated to reference it.

As a side fix, `config.validate()` is moved from inside `ClientConfig::try_from_reload_sig_and_config()` to after all patches (env and CLI) are applied in `main.rs` and `mobile.rs`, so validation reflects the actual final configuration.

## Motivation and Context

There was no single place describing where config values originate or how precedence works across config files, environment variables, and CLI flags. The new document fills that gap and serves as the reference when introducing new config fields.

The validation fix is a correctness improvement surfaced while writing the docs: validation was running on a partially-configured state before patches were applied, which could reject a valid final config or silently pass an invalid one.

## How Has This Been Tested?

- documentation no need test
- validate in a good flow before take_server out, already having unittest

## Types of changes

## Checklist:

### DIFF
--- a/docs/design_overview.md
+++ b/docs/design_overview.md
@@ -16,13 +16,15 @@ desktop or an iPhone.
 
 ## lightway-client
 
-lightway-client is a Linux implementation for a fully working Lightway client with both TCP and UDP support.
+lightway-client is a fully working Lightway client implementation with TCP and UDP support across multiple platforms, including Windows, Linux, macOS, and mobile devices.
 
-All clients possible work with mulitple server connections with different auth settings, and it still possible to set one auth for all connection, please refer [Connection Setting.](/docs/connection_setting.md)
+All client settings are centralized in a single [Config](/lightway-client/src/config.rs) struct, designed to cover multiple use cases simultaneously, and can work with multiple server connections with different auth settings, and it still possible to set one auth for all connections, please refer to [Connection Setting.](/docs/connection_setting.md)
 
 ## lightway-server
 
 lightway-server is a Linux implementation for a fully working Lightway server with both TCP and UDP support.
+
+All server settings are centralized in a single Config struct, please refer to [General Config Design Rationale](/docs/general_config.md)
 
 # Terminology
 
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -32,7 +32,7 @@ Identical to the CLI client flow, but all parameters use `SERVER` keywords — e
 
 ```mermaid
 flowchart TB
-    clientFn("client(config)")
+    clientFn("client(ClientConfig,...)")
     E0@{ shape: paper-tape, label: "SchemaFile" } --> A0
     0 -. "JsonSchema" .-> E0
     0 -. "Serialize" .-> E1
@@ -62,7 +62,7 @@ flowchart TB
     end
 
     0.1 -. "Deserialize" .-> 2
-    6 ==> clientFn
+    6 == "Config Lifecycle Methods" ==> clientFn
 ```
 
 ### Design Principles
@@ -107,6 +107,34 @@ fn client(config: Config) {
     let tun = Tun::new(android_only_field);
 }
 ```
+
+### Config Lifecycle Methods
+
+After the config is fully determined (step 6), three methods bridge it into the client runtime. `validate()` and `take_servers()` are methods on `Config`; `try_from_reload_sig_and_config()` is a constructor on `ClientConfig`.
+
+Typical call site order in `main.rs`:
+
+```rust
+config.validate()?;                             // fail fast before logging or other setup
+let servers = config.take_servers()?;           // normalize servers, transfer ownership
+let client_config = ClientConfig::<()>::try_from_reload_sig_and_config(
+    config_reload_signal,
+    config,
+)?;
+// servers and client_config are then used independently
+```
+
+#### `validate()`
+
+Checks the fully determined `Config` for conflicts and invalid values (mismatched socket-buffer or PMTUD settings on TCP, invalid Windows TUN parameters). Called independently in `main.rs` as the first step after config is determined, so validation fails fast before any further setup (logging, server normalization, runtime construction).
+
+#### `take_servers()`
+
+Must be called **before** constructing `ClientConfig`. It normalizes the config's flexible server representation — promoting single-server top-level fields, resolving CA certificates, and propagating auth credentials — and transfers ownership of the resulting `Vec<ConnectionConfig>` out of `Config`. All downstream consumers work against a uniform server list regardless of how the original config was expressed.
+
+#### `try_from_reload_sig_and_config()`
+
+The canonical constructor for `ClientConfig`. It consumes `config` (after `validate()` and `take_servers()` have been called), builds `TunConfig`, and wires up the optional `config_reload_signal` for hot-reload support.
 
 ---
 
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -1,0 +1,169 @@
+# General Config Design Rationale
+
+## Introduction
+
+The config generation is unified across all clients and the server. For the server, the scope is simple — it only supports native builds on Linux. For the client side, real-world usage is considerably broader, particularly when generating JSON schema for any platform target from a macOS machine, which is a common working scenario for a small team.
+
+This document is not meant to impose restrictions on development. Instead, it addresses development pain points ahead of time by providing guides that help you leverage what is currently available.
+
+---
+
+## Methods
+
+### Config as Single Source of Truth
+
+`Config` is the single source of truth for all clients — all user inputs, whether from a UI or a file, flow through it. JSON schema generation from the CLI is designed to be a general-purpose mechanism for all client tooling.
+
+`Config` derives both `ConfigPatch` and `JsonSchema`, which enables the layered override flow described below.
+
+### Config Flow
+
+**Desktop client flow (steps 0–6):**
+
+Starting from default values (0), the config evolves through each step along the bold lines. `ConfigPatch` plays a central role along the dot lines, generating patches by deserializing from a file, environment variables, and CLI options — each applied as a layered override in sequence.
+
+**Mobile client flow (steps 0, 1, 2, 6):**
+
+Mobile takes a shorter path, skipping the intermediate steps after `2.ConfigContent`. Rather than reading from a file, config content comes from a Dynamic UI driven by a JSON schema file generated at compile time from the same `Config` struct via the CLI client. Both desktop and mobile ultimately share the same `Config` source of truth, with the mobile flow being a streamlined subset.
+
+**Server flow (steps 0–6):**
+
+Identical to the CLI client flow, but all parameters use `SERVER` keywords — e.g. `LW_CLIENT_*` becomes `LW_SERVER_*`.
+
+```mermaid
+flowchart TB
+    clientFn("client(config)")
+    E0@{ shape: paper-tape, label: "SchemaFile" } --> A0
+    0 -. "JsonSchema" .-> E0
+    0 -. "Serialize" .-> E1
+
+    subgraph "android (foreign)"
+      A0@{ shape: manual-input, label: "Dynamic UI"}
+    end
+
+    A0 --"mobile"--> 2
+    2  --"mobile shortcut" --> 6
+    E1@{ shape: paper-tape, label: "ConfigFile" } --"cli"--> 2
+
+    subgraph main.rs or mobile.rs
+      1("1: Config::default()") ==> 2
+      2("2: ConfigContent") ==> 3
+      3("3:Envars(LW_CLIENT_*)") ==> 4
+      4("4: CLI Option") ==> 5
+      5("5: Special Envars(LW_CLIENT_RUST_LOG)") ==> 6
+      6("6: Config (determined)")
+    end
+
+    subgraph config.rs
+      0@{ shape: braces, label: "0: Config" } == "Default" ==>1
+      0 -. "Patch" .-> 0.1
+      0.1 -. "Deserialize" .-> 3
+      0.1@{ shape: braces, label: "ConfigPatch" } -. "Parser" .-> 4
+    end
+
+    0.1 -. "Deserialize" .-> 2
+    6 ==> clientFn
+```
+
+### Design Principles
+
+When adding a platform-specific field, a feature gate may be required — e.g. `#[cfg(feature = "...")]` — with platform intent communicated via `x-cfg` and `format` attributes in the JSON schema. Critically, `#[cfg(target)]` must **not** be applied to fields: if it were, those fields would be absent when generating schema on a non-matching host, making it impossible to generate all schemas from any desktop machine.
+
+To keep things clean, the practical approach with the least friction is:
+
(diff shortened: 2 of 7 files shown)


docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
 
 ## lightway-client
 
-lightway-client is a Linux implementation for a fully working Lightway client with both TCP and UDP support.
+lightway-client is a fully working Lightway client implementation with TCP and UDP support across multiple platforms, including Windows, Linux, macOS, and mobile devices.
+
+All client settings are centralized in a single [Config](/lightway-client/src/config.rs) struct, designed to cover multiple use cases simultaneously. `Config` derives both `ConfigPatch` and JsonSchema file.
```
kp-samuel-tam: From this line onwards, I think a 6-paragraph explanation of "how `Config` works" is definitely out of place in `design_overview.md`. The original markdown is more of a starting point for developers to grasp how we segment Lightway into Rust crates. I suggest moving the edits into its own doc and having a link in `design_overview.md` to point to it.

docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
...
+      
+      0.1 -. "Deserialize" .-> 3
+      0.1@{ shape: braces, label: "ConfigPatch" } -. "Parser" .-> 4
+    end
+
+    0.1 -. "Deserialize" .-> 2
+    6 ==> clientFn
+```
+
+As shown, `Config` is the single source of truth for all clients — all user inputs, whether from a UI or a file,
```
kp-samuel-tam: From this line onwards there's a clear disjoint between the convention you're documenting and how `Config` actually ships today. Looking at the merged PR #411:

- `wintun_file`, `enable_dpapi` → `#[cfg(windows)]` (target gate on field)
- `route_mode`, `dns_config_mode`, `enable_pmtud`, `enable_tun_iouring`, `enable_inside_pkt_encoding` → `#[cfg(desktop)]` (custom cfg on field)
- `enable_batch_receive` → `#[cfg(batch_receive)]` (custom cfg on field)

The doc says "feature gates belong on fields" but there's 8+ fields in the prior PR being documented use target/custom cfg on fields.

Even if we're aligned on the direction, the code needs to follow the rule before we declare it a developer practice in the docs.

docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
...
+}
+
+#[cfg(android)]
+fn client(config: Config) {
+     let Config {
+       android_only_field,
+       ..
+     } = config;
+    let tun = Tun::new(android_only_field);
+}
```
kp-samuel-tam: Looks like this pseudocode was copy-pasted from https://github.com/expressvpn/lightway/pull/411#discussion_r3263321514 with typos - `#[cfg(window)]` is missing the `s`, and `win_only_filed` should be `win_only_field`.

Also, this code actively encourages the paired use of `#[cfg(windows)]` and `#[cfg(feature="windows")]`. This is unsound because we can't ever guarantee they are in sync or supported, adding to the drift Thomas was trying to talk about.

docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
...
+even though the broader approach to consuming JSON schema remains an open question in the [discussion](https://github.com/expressvpn/lightway/pull/411#discussion_r3166422937).
+
+To add a platform-specific field, a feature gate is required — e.g. `#[cfg(feature = "...")]`
+— with platform intent communicated via x-cfg and format attributes in the JSON schema.
+This makes it easy to tailor the schema on the client side while still being generatable from the CLI.
+That said, introducing more feature gates alongside existing target gates risks making the repo harder to follow.
+To keep things clean, the practical approach with the least friction is:
+
+1. Feature gates belong on **fields** of the `Config` struct.
+2. `cfg` target attributes belong on **functions**.
```
kp-samuel-tam: This contradicts with line 72.

> [...] the broader approach to consuming JSON schema remains an open question in the [discussion](https://github.com/expressvpn/lightway/pull/411#discussion_r3166422937).

In addition, you said in https://github.com/expressvpn/lightway/pull/411#discussion_r3271003699:

> If there is no consensus on whether to prefer #[cfg(feature = "...")] over target gates as a general guide, we can leave that question open, remove the note for now, and let the mobile changes land as they are. We can revisit the broader convention in a follow-up once alignment is reached.

But you are writing the exact same "note" here. I am not aware of any further discussions from that point on, so this edit needs to happen *after* we have that conversation. I'd suggest pinging @xv-thomas-leong specifically for this.

docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
...
+
+#[cfg(android)]
+fn client(config: Config) {
+     let Config {
+       android_only_field,
+       ..
+     } = config;
+    let tun = Tun::new(android_only_field);
+}
+```
```
kp-samuel-tam: Commit is titled `doc: add description for JsonSchema` but the `JsonSchema` description isn't in the room with us. I'd expect an example `JsonSchema` how-to code -> example JSON config in here as a developer.

docs/design_overview.md
```diff
@@ -16,7 +16,111 @@ desktop or an iPhone.
...
-lightway-client is a Linux implementation for a fully working Lightway client with both TCP and UDP support.
+lightway-client is a fully working Lightway client implementation with TCP and UDP support across multiple platforms, including Windows, Linux, macOS, and mobile devices.
+
+All client settings are centralized in a single [Config](/lightway-client/src/config.rs) struct, designed to cover multiple use cases simultaneously. `Config` derives both `ConfigPatch` and JsonSchema file.
+
+**Desktop client flow (steps 0, 1, 2, 3, 4, 5, 6):**
+
+Starting from default values (0), the config evolves through each step along the bold lines. Meanwhile, ConfigPatch plays a central role along the dot lines, generating patches by deserializing from a file, environment variables, and CLI options — each applied as a layered override in sequence.
+
+**Mobile client flow (steps 0, 1, 2, 6):**
```
kp-samuel-tam: There is no visible mobile-only edge from 2 → 6 in the graph. Either add `2 -. "mobile" .-> 6` to the mermaid or update the prose to match what the diagram actually shows.

kp-samuel-tam (changes requested):
It's a great write-up overall, the flow chart captures the layering really well. I would like to raise a few blockers.

docs/design_overview.md
kp-antonio-yang: Yeh, this make sense.  I will move these to a separated section.

docs/design_overview.md
kp-antonio-yang: Yeh, not all the code following the design and only mobile does.

The reason it when we use json schema on windows/macos client and we will refactor it to follow this design.  For now, we do the least change on the code when mobile introduce.  And it will make really sense when developer working on that target and test it at the same time, so we do not touch these target now.

docs/design_overview.md
kp-antonio-yang: It will always need to enable the corresponding feature gate for targets with this design,  like current mobile feature.  

Also, we can make sure it always sync and supported via `cargo make`, and also that is the reason we have script or make file in the project.

docs/design_overview.md
kp-antonio-yang: Will this better to note the `(cli only)`  on step 3, 4, 5?   The chart shown that all the client follow the same flow, and the mobile only skip some steps.

docs/design_overview.md
kp-antonio-yang: Agreed, this doc is asked from him in a video call and the #411 merged in the video call.  

I believe he will like every reviewer can make decision and rolling our project to share the ownership, and he will back to office with a reviewed version.  :smile:

docs/design_overview.md
kp-antonio-yang: There are only `format` and `x-cfg` fields for now, but there is more how about when a feature gate goes in and not only the `x-cfg`.   The commit message might misleading, so I squash this commit with previous one.

kp-antonio-yang:
Hi @kp-samuel-tam 
The docs is move to a split one general_config.md, and also an general introduction of this topic is added with two possible option as open question, such that we are not enforce to use `#[cfg(feature = ...)]` at this moment and keep it open.  May you review again with your suggestion on these?  Thank you. :pray:

docs/general_config.md
```diff
@@ -0,0 +1,144 @@
+# General Config
+
+## Use cases by developer persona
+
+The general config centralizes all settings into a single `Config` struct and aligns the config generation flow across clients. However, only the mobile feature has been implemented with this design so far — existing clients may not adopt it until a larger refactor happens.
+There are two possible directions going forward:
```
xv-thomas-leong: Could you clarify when to use each direction? 

It's not clear what the differences are or which scenarios call for which approach. If they're effectively the same, let's pick one as the standard going forward.

docs/general_config.md
```diff
@@ -0,0 +1,144 @@
+# General Config
```
xv-thomas-leong: Suggesting a more concise title:

```suggestion
# Client Config Design Rationale
```

Can you also link this doc in `lightway-client/src/config.rs`, such that people are aware that there is a relevant doc here?

docs/design_overview.md
xv-thomas-leong: The lingering concerns are a signal that the documentation itself isn't clear enough. Ideally, after reading your doc, people should walk away feeling that what you're proposing makes sense.

The current document covers the _what_ in depth, but doesn't really address the _why_. This is especially important in your case because you're deviating from the standard, which naturally raises concerns. Addressing those concerns head-on is exactly what this change needs to do, and it's also why I suggested writing this up in the first place. 

In a technical context, explaining the _why_ well means laying out your motivation, the technical constraints you're working within, and how this design directly addresses those challenges. I'd suggest following this exact structure in your doc for clarity.

docs/general_config.md
```diff
@@ -0,0 +1,144 @@
...
+    0 -. "Serialize" .-> E1
+    
+    subgraph "android (foreign)"
+      A0@{ shape: manual-input, label: "Dynamic UI"}
+    end
+
+    A0 --"mobile"--> 2
+    E1@{ shape: paper-tape, label: "ConfigFile" } --"cli"--> 2
+
+    subgraph main.rs or mobile.rs
```
xv-thomas-leong, line 48: You mentioned:

> Desktop client flow (steps 0, 1, 2, 3, 4, 5, 6):

> Mobile client flow (steps 0, 1, 2, 6):

It seems the difference is step 3, 4, and 5. Would it be clearer if we can reflect that on the chart directly?

docs/design_overview.md
xv-thomas-leong: How about we make this less controversial by only applying this rule on mobile for now? i.e.:

* Mobile -> use `#[cfg(feature="mobile")]` to allow schema generation on desktop
* Desktop -> follow Rust's default convention

As you noted in the doc, desktop won't adopt this convention until the larger refactor lands. Any new desktop config added before then would deviate from the proposal anyway, which somewhat undercuts its value if we roll it out broadly now.

docs/design_overview.md
xv-thomas-leong: Thanks Samuel for reviewing!

docs/general_config.md
kp-antonio-yang: It is `General Config`.

I should doc the server part, although it is really similar with `cli` part of the client.

docs/general_config.md
kp-antonio-yang: A mobile shortcut path is added

docs/design_overview.md
kp-antonio-yang: Yes, the why is the important and making focus.  

A introduction is added, and also the using scenario put more details about why the scope is boarder than before.  This issue should not be a problem.

docs/design_overview.md
kp-antonio-yang: The intent was never to lay down strict rules — it started as a brief three-line comment to guide developers when they need to revisit the JSON schema. As it evolved into a document, it made more sense to capture not just the current state but also the details we will likely encounter going forward, especially since the discussion is already open.  Current document put introduction and easier for reader to have the focus.

docs/general_config.md
kp-antonio-yang: If we can reach a decision on `Option 1`, I can drop the redundant documentation and keep things lean. Once that lands, a follow-up PR can enforce the same design across the desktop clients.

Personally, I lean towards `Option 1` with feature gates. Since we are already using `Makefile.toml`, passing --features is not an extra burden — it is a common pattern in our build flow for each target. However, since there seems to be resistance towards `--feature=target`, so I am leaving this as an open question for now.

kp-antonio-yang:
Hi @xv-thomas-leong 

Please take a look at the updated doc when you get a chance. 🙏
Here is a summary of the major changes since the last review:

- Added server-side description.
- Added doc links in the source code.
- Explained the reasoning behind the broader persona scope and called it out explicitly.

On the options section — if we would rather not leave it open-ended in the doc, I am happy to implement either Option 1 or Option 2 in a follow-up PR. But we need to reach consensus here first before I proceed.

docs/design_overview.md
xv-thomas-leong: > The intent was never to lay down strict rules

No, by writing down what people should do, you _are_ setting rules, no matter what you call it.

If you want to keep your vision for the desktop config, you may put it in a section labeled "Future Plans", but it must be distinctive that it's not meant to be followed.

docs/general_config.md
xv-thomas-leong: Then I would suggest `Client/Server Config Design Rationale`. The word "General" doesn't really mean anything.

I see you've extended the diagram to include server as well, but that has become a bit messy. Can you separate it into its own flow chart?

kp-antonio-yang:
Will update docs with #457 and #451, because these are also related to the Config.

kp-antonio-yang:
Will update the changes in #511

kp-antonio-yang:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed in 7 days if no further activity occurs.

I will back to this, thanks for reminds.

PR #442, opened 2026-05-29
CVPN-2571: Server support batched receive (recvmmsg)

kp-thomas-yau:
## Description

Similar to client (https://github.com/expressvpn/lightway/pull/378 and https://github.com/expressvpn/lightway/pull/384), the PR aims to bring recvmmsg/recvmsg_x to the server as well. Since `recvmsg_x` is not included in macOS libc and we have defined the headers already in the client binary, I have moved message header/function header to app-utils so it can be shared between client + server. 

Unlike client which only takes in a BytesMut in the batch receive function, it will take in a `BatchRecvSlot` in, to store all the relevant info (socket address/cmsg).

## Motivation and Context

Fewer syscall + faster performance

## How Has This Been Tested?

e2e tests, unit tests

Tested on a 40Gbps Linux server with M5 Pro MacBook Pro on the client-side (with a 10Gbps line): 
| (in Mbps) | DL Without Expreslane |    UL   | With Expresslane |    UL   |
|:---------:|:---------------------:|:-------:|:----------------:|:-------:|
|  recvmsg  |        1649.22        | 1222.03 |      1846.05     | 1302.49 |
|  recvmmsg |        1790.45        | 1266.35 |      1859.61     | 1234.31 |

Another test is done on 2 Linux machines with 40Gbps line:
| (in Mbps) | DL Without Expreslane |    UL    | With Expresslane |    UL    |
|:---------:|:---------------------:|:--------:|:----------------:|:--------:|
|  recvmsg  |        9930.17        | 31953.70 |     11378.48     | 36010.30 |
|  recvmmsg |        11304.62       | 37168.78 |     14362.08     | 35953.04 |

## Types of changes

## Checklist:


kp-antonio-yang:
It seems we are not using refactor: as a label.
Will following to new commit titile better?
- `refactor: Move msghdr_x and recvmsg_x declaration to app-utils`  -> `app-utils: collect msghdr_x and recvmsg_x from ...`
- `refactor: Replace peer_addr with take_peer_addr to consume stored address and update associated logic`  -> `server: use take_peer_addr to consume stored address`

lightway-server/src/io/outside/udp.rs
kp-antonio-yang, line 494: It seems the first return parameter `usize` is the length of  `Vec<BatchRecvMetadata>)>`, because every error will raise and return.   It should not be used as error code holder.   It might be written by AI without Rust convention.

Will this a better return type having meaningful fields since it was complicated and asked by inter?

```rust
BatchRecvMetadata {
    peer: ...,
    local: ...,
    reply_pktinfo: Option<....>,
}
```

Developer still easy to have the length by `.len()` on the vector, and the field will be much easier to read, since there are two `SocketAddr`, ex: `.peer` and `.local` are clear than `.0` and `.1`

kp-thomas-yau: > because every error will raise and return. It should not be used as error code holder. It might be written by AI without Rust convention.

Not sure I understand you here, `recv_multiple_with_metadata` returns how many packets have been received via `Ok` result and not the error code? `recv_multiple_with_metadata` will return an `Err` result already if the syscall returned an error here. Plus, I have used `?` operator here already....

But I do agree that returning with an actual struct will be easier to understand.

lightway-app-utils/src/recvmsg_x.rs
```diff
@@ -0,0 +1,100 @@
...
+    unsafe { !libc::dlsym(libc::RTLD_DEFAULT, name.as_ptr()).is_null() }
+}
+
+/// Check whether the platform supports batch receiving via `recvmsg_x`.
+///
+/// On Apple platforms `recvmsg_x` is a private syscall that may not be present
+/// on every OS version, so callers should consult this probe before relying on
+/// the syscall.
+pub fn is_batch_receive_available() -> bool {
+    *RECVMSG_X_AVAILABLE
```
kp-antonio-yang, line 39: May this function also available for other platform but we do the `#[cfg(apple)]` in side the function and in a small scale?  It is straightforward when we already have a good document of the function.  When we put `#[cfg(apple)]` outside,  there is no doc,  it might be confusing why other platform no need a mod `recvmsg_x`.

```rust
     if cfg!(macos) {
         if ! *RECVMSG_X_AVAILABLE {
             warn!("batch receive (recvmsg_x) not available on this system, batch receive disabled");
         }
         *RECVMSG_X_AVAILABLE
     } else {
         true
     }
```

If this is available for every platform, our code reduce a lot of `#[cfg(apple)]`, and the confusing thereof.
Besides, the code will be trim less, for example.

original:
```rust
        #[cfg(batch_receive)]
        let batch_receive_enabled = if enable_batch_receive {
            #[cfg(macos)]
            if lightway_app_utils::recvmsg_x::is_batch_receive_available() {
                info!("Using batch receiver");
                true
            } else {
                warn!(
                    "batch receive (recvmsg_x) not available on this system, batch receive disabled"
                );
                false
            }
            #[cfg(linux)]
            {
                info!("Using batch receiver");
                true
            }
```

replaced:
```rust
let batch_receive_enabled = enable_batch_receive && is_batch_receive_available(); 
if batch_receive_enabled {  
      // NOTE: if there other feature conflicts in the future, we still easy to handle this message 
      info!("Using batch receiver");
}
```
And the warning message still comes out as we need, and the successful message is keep outer, we still can easy to modify it if there are other conflict feature we need to handle.

kp-thomas-yau: No, there's no need for this function to exist on other platforms at all. `recvmmsg` has been introduced in Linux 2.6.33 already, which is released in 2010 and basically all Linux platform that we support already have this built-in. The reason this is needed on Apple platform is already embedded in the docs of `is_batch_receive_available()` 

<details><summary>Docs</summary>
<p>

```
/// Check whether the platform supports batch receiving via `recvmsg_x`.
///
/// On Apple platforms `recvmsg_x` is a private syscall that may not be present
/// on every OS version, so callers should consult this probe before relying on
/// the syscall.
```

</p>
</details> 

`recvmsg_x()` is technically a private API and Apple may remove this in a future version (Highly likely not, but just to play it safe). I really don't see a need for the same function to exist on Linux.

kp-antonio-yang: To clarify — this is not about whether the function exists natively on the platform. The idea is to provide the function on all platforms with consistent behavior, keeping the code slim and free of verbose target gates.

On Linux, `is_batch_receive_available` always returns `true` and will be optimized away by the compiler. Other platforms might return `false`, so the same call site works everywhere without any `#[cfg(...)]` branching out side the function.

lightway-server/src/io/outside/udp.rs
kp-antonio-yang: Yes, your understanding is correct, I can see the exact code change from [6d16ac3](https://github.com/expressvpn/lightway/commit/6d16ac39180f8ac023605b1bccc61dfb811e0dcc) to [74a2722](https://github.com/expressvpn/lightway/commit/74a2722f4ac643a05392bccf77b671df22fbfd71)

The issue I pointed out is the return parameter `n`, which in other languages is commonly used as an error code holder. Since this function already returns a `Result`, carrying n as a return value is redundant and an anti-pattern in Rust — the error is already handled by the `Result` type.

kp-thomas-yau: I'm still confused - if I'm returning a Result here, why would you think `n` is a error code holder then? I'm pretty sure n/i usually refers to a count or byte count or sth when I'm returning Ok()

lightway-app-utils/src/recvmsg_x.rs
kp-thomas-yau: I really don't think there is "verbose" target gates here... It's literally just few extra lines here:
```
#[cfg(linux)]
let batch_receive_enabled = enable_batch_receive;
#[cfg(macos)]
let batch_receive_enabled = if enable_batch_receive {
    if lightway_app_utils::recvmsg_x::is_batch_receive_available() {
        true
    } else {
        warn!(
            "batch receive (recvmsg_x) not available on this system, batch receive disabled"
        );
        false
    }
} else {
    false
};
```
I personally not a fan of unifying the target gates just for the sake of it. This really is a macOS-only behaviour....

kp-antonio-yang: Okay.

kp-antonio-yang:
I am not familiar with MacOSX, so I only do suggestions on the language syntax part.   

Wish I can make inputs on this PR since taking time on this, not really want to be picky.  It looks good to me.

kp-antonio-yang (dismissed):

lightway-server/src/lib.rs
```diff
@@ -384,6 +387,8 @@ pub async fn server<SA: for<'a> ServerAuth<AuthState<'a>> + Sync + Send + 'stati
                 conn_manager.clone(),
                 config.bind_address,
                 config.udp_buffer_size,
+                #[cfg(batch_receive)]
```
kp-mariappan-ramasamy: This entire #[cfg(batch_receive)] is not needed I think. We could remove it in all places

kp-thomas-yau: I somehow always think we fully support Lightway server on Windows lol, removed it

kp-mariappan-ramasamy (approved):

PR #445, opened 2026-06-01
[CVPN-2251] client: dns_manager for Windows

xv-piotr-haber:
## Description
Add implementation of dns_manager for Windows platform.
Using NRPT (Name Resolution Policy Table) to always use configured DNS resolver.

Based on work of geekjy
[https://github.com/geekjy/lightway/blob/ef21c1567548e1608ef2b99e1acffd84310e30ee/lightway-client/src/platform/windows/dns_manager.rs]
## Motivation and Context
We want to prevent DNS leaks on Windows.
For that we need to be sure Windows uses a DNS resolver over VPN tunnel.

## How Has This Been Tested?
Unit tests.
End-to-end test with Windows client and Linux server.

## Types of changes

## Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -23,3 +23,7 @@
 
 * [Wireshark lightway plugin](wireshark.md)
 
+# Platform specific info
+
+* [Windows](./windows.md)
+
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -1,0 +1,13 @@
+# Windows specific info
+
+## DNS name resolution
+
+To prevent DNS leaks lightway client uses [NRTP](https://www.island.io/blog/man-7-windows-nrpt) (Name Resolution Policy Table)
+rule to try and route all DNS queries via tunnel.
+
+Unfortunately some programs (ex. nslookup) do not obey this which might still lead to DNS leaks.
+
+Another issue is that when lightway client crashes the NRPT rule might be left leading to name resolution problems.
+The old rule will be deleted on next client connection.
+In order to remove the rule manually following PowerShell (run as Administrator) command might be used:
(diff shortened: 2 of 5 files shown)


kp-mariappan-ramasamy (commented):
LGTM

I found some apps like nslookup might skip NPRT rule
https://stackoverflow.com/questions/60144350/why-dont-nrpt-rules-affect-dns-queries-on-my-windows-10-box
https://www.island.io/blog/man-7-windows-nrpt#the-mystery-of-the-missing-rules

I think we could document it, since it might confuse users

xv-piotr-haber:
> LGTM
> 
> I found some apps like nslookup might skip NPRT rule https://stackoverflow.com/questions/60144350/why-dont-nrpt-rules-affect-dns-queries-on-my-windows-10-box https://www.island.io/blog/man-7-windows-nrpt#the-mystery-of-the-missing-rules
> 
> I think we could document it, since it might confuse users

Where should I document this?
Another, more serious issue is that when lightway client dies ungracefully the NRPT rule is left and affects name resolving.

kp-mariappan-ramasamy:
> Where should I document this?

We can add a doc inside https://github.com/expressvpn/lightway/tree/main/docs explaining it.

> Another, more serious issue is that when lightway client dies ungracefully the NRPT rule is left and affects name resolving.

Yeah true. We try not to have panic and handle errors gracefully, but still we use c libraries so its possible.
Should we add a client sub-command to do it ? or do you have any other ideas

xv-piotr-haber:
> Yeah true. We try not to have panic and handle errors gracefully, but still we use c libraries so its possible.
> Should we add a client sub-command to do it ? or do you have any other ideas

The best I could think of was adding a removal of existing NRPT rule before setting-up a new one. That way when the client crashes and is restarted by Qt app we will have a sane setup.

xv-thomas-leong (approved):

PR #451, opened 2026-06-09
server: config: expose ServerConfig::try_from method

kp-antonio-yang:
## Description
- [X] collect `ServerConfig::try_from_auth_and_config`
- [X] add `validate` method of `Config` 
- [X] adding validation for `enable_expresslane`, `proxy_protocol`
- [X] validate config when building `ServerConfig`

## Motivation and Context
Provide reusable methods, including validations and casting, for `Config`, such that these method can easier to reused by any downstream projects.  Track with ticket CVPN-2602.

## How Has This Been Tested?
- [X] testcases for new validation functions

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/args/connection_type.rs
+++ b/lightway-app-utils/src/args/connection_type.rs
@@ -19,7 +19,7 @@ pub enum ConnectionType {
 }
 
 impl ConnectionType {
-    /// A helper function easier to use especially in mobile
+    #[allow(missing_docs)]
     pub fn is_tcp(&self) -> bool {
         *self == ConnectionType::Tcp
     }
--- a/lightway-app-utils/src/args/connection_type.rs
+++ b/lightway-app-utils/src/args/connection_type.rs
@@ -23,6 +23,11 @@ impl ConnectionType {
     pub fn is_tcp(&self) -> bool {
         *self == ConnectionType::Tcp
     }
+
+    #[allow(missing_docs)]
+    pub fn is_udp(&self) -> bool {
+        *self == ConnectionType::Udp
+    }
 }
 
 impl From<ConnectionType> for LWConnectionType {
--- a/lightway-server/src/config.rs
+++ b/lightway-server/src/config.rs
@@ -233,6 +233,13 @@ impl Config {
             anyhow::ensure!(self.mode.is_udp(), "Expresslane only work in udp mode")
         }
 
+        if self.proxy_protocol {
+            anyhow::ensure!(
+                self.mode.is_tcp(),
+                "Proxy protocol only support with tcp mode"
+            )
+        }
+
         Ok(())
     }
 }
@@ -254,6 +261,14 @@ mod tests {
         let mut config = Config::default();
         config.mode = ConnectionType::Tcp;
         config.enable_expresslane = true;
+        assert!(config.validate().is_err());
+    }
+
+    #[test]
+    fn validate_proxy_protocol() {
+        let mut config = Config::default();
+        config.mode = ConnectionType::Udp;
+        config.proxy_protocol = true;
         assert!(config.validate().is_err());
     }
 }
--- a/lightway-server/src/config.rs
+++ b/lightway-server/src/config.rs
@@ -229,6 +229,10 @@ impl Default for Config {
 impl Config {
     /// Ensure the config is validated, and alerted when there's a conflict in the settings.
     pub fn validate(&self) -> anyhow::Result<()> {
+        if self.enable_expresslane {
+            anyhow::ensure!(self.mode.is_udp(), "Expresslane only work in udp mode")
+        }
+
         Ok(())
     }
 }
@@ -243,5 +247,13 @@ mod tests {
     fn validate_default_config() {
         let config = Config::default();
         assert!(config.validate().is_ok());
+    }
+
+    #[test]
+    fn validate_enable_expresslane() {
+        let mut config = Config::default();
+        config.mode = ConnectionType::Tcp;
+        config.enable_expresslane = true;
+        assert!(config.validate().is_err());
     }
 }
--- a/lightway-server/src/config.rs
+++ b/lightway-server/src/config.rs
@@ -225,3 +225,23 @@ impl Default for Config {
         }
     }
 }
+
+impl Config {
+    /// Ensure the config is validated, and alerted when there's a conflict in the settings.
+    pub fn validate(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
+}
+
+// Note it easier to see what is different from default in each testcase
+#[allow(clippy::field_reassign_with_default)]
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn validate_default_config() {
+        let config = Config::default();
+        assert!(config.validate().is_ok());
+    }
+}
--- a/lightway-server/src/lib.rs
+++ b/lightway-server/src/lib.rs
@@ -262,6 +262,65 @@ pub struct ServerConfig<SA: for<'a> ServerAuth<AuthState<'a>>> {
     pub randomize_ippool: bool,
 }
 
+impl<SA: for<'a> ServerAuth<AuthState<'a>>> ServerConfig<SA> {
+    pub fn try_from_auth_and_config(auth: SA, config: config::Config) -> Result<Self> {
+        let mut tun_config = lightway_app_utils::TunConfig::default();
+        if let Some(tun_name) = config.tun_name {
+            tun_config.tun_name(tun_name);
+        }
+        tun_config.up();
+
+        Ok(crate::ServerConfig {
+            mode: match config.mode {
+                lightway_app_utils::args::ConnectionType::Udp => {
+                    crate::ServerConnectionMode::Datagram(None)
+                }
+                lightway_app_utils::args::ConnectionType::Tcp => {
+                    crate::ServerConnectionMode::Stream(None)
+                }
+            },
+            auth,
+            server_cert: config.server_cert,
+            server_key: config.server_key,
+            tun_config,
+            ip_pool: config.ip_pool,
+            ip_map: config.ip_map.unwrap_or_default().try_into()?,
+            inside_io: None,
+            tun_ip: config.tun_ip,
+            lightway_server_ip: config.lightway_server_ip,
+            lightway_client_ip: config.lightway_client_ip,
+            lightway_dns_ip: config.lightway_dns_ip,
+            use_dynamic_client_ip: false,
+            enable_expresslane: config.enable_expresslane,
+            expresslane_keys_rotation_interval: config.expresslane_keys_rotation_interval.into(),
+            expresslane_cb: None,
+            expresslane_metrics: None,
+            event_cb: None,
+            enable_pqc: config.enable_pqc,
+            #[cfg(target_os = "linux")]
+            enable_tun_offload: config.enable_tun_offload,
+            #[cfg(feature = "io-uring")]
+            enable_tun_iouring: config.enable_tun_iouring,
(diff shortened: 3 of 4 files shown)


lightway-app-utils/src/args/connection_type.rs
```diff
@@ -19,10 +19,15 @@ pub enum ConnectionType {
 }
 
 impl ConnectionType {
-    /// A helper function easier to use especially in mobile
+    /// A helper function easier to use
```
kp-samuel-tam: Two comments of
```
/// A helper function easier to use
```
don't have value here, I suggest removing it.

lightway-server/src/config.rs
```diff
@@ -128,7 +128,7 @@ pub struct Config {
 
     #[patch(attribute(clap(long)))]
     #[patch(attribute(doc = "The key update interval for DTLS/TLS 1.3 connections"))]
-    pub key_update_interval: NonZeroDuration,
+    pub key_update_interval: Duration,
```
kp-samuel-tam: Why was `key_update_interval` relaxed from `NonZeroDuration` to `Duration`? The commit message didn't explain why.

lightway-server/src/lib.rs
```diff
@@ -258,6 +259,95 @@ pub struct ServerConfig<SA: for<'a> ServerAuth<AuthState<'a>>> {
...
+            mode: match mode {
+                lightway_app_utils::args::ConnectionType::Udp => {
+                    crate::ServerConnectionMode::Datagram(None)
+                }
+                lightway_app_utils::args::ConnectionType::Tcp => {
+                    crate::ServerConnectionMode::Stream(None)
+                }
+            },
+            auth,
+            server_cert,
```
kp-samuel-tam: I think the original `server_cert: config.server_cert` is easier to follow than to have the long list of options repeat itself twice.

lightway-server/src/config.rs
```diff
@@ -218,3 +216,109 @@ impl Default for Config {
         }
     }
 }
+
+impl Config {
+    /// Validatate but skip some fields, such that downstream can easy to use the validate function
+    /// but skip some field using in different way
```
kp-samuel-tam:
```suggestion
    /// Allow downstream to skip validation on some fields
```

Please remember to proofread in future PRs.

lightway-server/src/config.rs
```diff
@@ -218,3 +216,109 @@ impl Default for Config {
...
+    pub fn validate_without(&self, skips: &[&'static str]) -> anyhow::Result<()> {
```
kp-samuel-tam: The bigger problem is that we are adding `config.validate()?;` as an undocumented contract. Unlike the original code that does validation upfront, now anyone reimplementing `lightway-server/src/main.rs` must remember to call `config.validate()?;`... which is not that obvious.

A skip list for validation also seems over-engineered right now, do we have good reason why we'd want to skip validation? Besides, the skip list being stringly-typed makes it prone to typos, use enums.

lightway-server/src/config.rs
```diff
@@ -218,3 +216,109 @@ impl Default for Config {
...
+                "Expresslane only work in udp mode"
+            )
+        }
+
+        if self.proxy_protocol && !skips.contains(&"proxy_protocol") {
+            anyhow::ensure!(
+                self.proxy_protocol == self.mode.is_tcp(), // XNOR
+                "Proxy protocol only support with tcp mode"
+            )
+        }
```
kp-samuel-tam: These should not hard fail. Especially for `enable_expresslane` it's only a mere suggestion to enable Expresslane and it is not enforced; right now if we are using TCP, Expresslane code will just no op.

Plus, the XNOR comment is confusing for a config check... I'd remove.

lightway-server/src/config.rs
```diff
@@ -218,3 +216,109 @@ impl Default for Config {
...
+            validate_configuration_file_path(user_db, Validate::OwnerOnly)
+                .with_context(|| format!("Invalid user db file {}", user_db.display()))?;
+        }
+
+        if !skips.contains(&"key_update_interval") {
+            anyhow::ensure!(
+                !self.key_update_interval.is_zero(),
+                "key_update_interval should greater than zero"
+            )
+        }
```
kp-samuel-tam: This could be avoided if we don't touch `NonZeroDuration`.

lightway-server/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+pub mod auth;
```
kp-samuel-tam: Do we really need to expose the crate as public?

lightway-server/src/auth.rs
```diff
@@ -10,7 +10,7 @@ use jsonwebtoken::{Algorithm, DecodingKey, Validation};
 use lightway_core::LightwayFeature;
 use pwhash::unix;
 
-use lightway_server::{ServerAuth, ServerAuthHandle, ServerAuthResult};
+use crate::{ServerAuth, ServerAuthHandle, ServerAuthResult};
```
kp-samuel-tam: Maybe we can stick to original `lightway_server::` instead of `crate::`? Then we can leave this file untouched.

kp-samuel-tam (commented):
Some comments, please review.

lightway-server/src/auth.rs
kp-antonio-yang: The `auth`  mod move to lib, it will be part of crate.  So we can not use `lightway_server::` anymore.   You can double check the result.  
```console
error[E0432]: unresolved import `lightway_server`
  --> lightway-server/src/auth.rs:13:5
   |
13 | use lightway_server::{ServerAuth, ServerAuthHandle, ServerAuthResult};
   |     ^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `lightway_server`
   |
help: there is a crate or module with a similar name
   |
13 - use lightway_server::{ServerAuth, ServerAuthHandle, ServerAuthResult};
13 + use lightway_core::{ServerAuth, ServerAuthHandle, ServerAuthResult};

```

Here is the explanation.
In previous status, the `auth` is part of bin, so it can get `ServerAuth` from a crate named `lightway_server::`, and now the `auth` is part of lib, so it can get `ServerAuth` from exactly the same `crate`, and it is not allow to write `lightway_server::` here.

lightway-server/src/lib.rs
kp-antonio-yang: Yes, it is required.  How can bin use `Auth`, if the `Auth` is not public in lib?

lightway-server/src/config.rs
kp-antonio-yang: Yeh, I will drop the commit.  
It is a border topic and no really related of this PR, might have discussion later.

lightway-server/src/config.rs
kp-antonio-yang: Sorry about confusing, it is clear now.

lightway-app-utils/src/args/connection_type.rs
kp-antonio-yang: Agree with you but there is a missing doc checker in our CI.   I still need to write something, even the function name is strait forward.
```
warning: missing documentation for a method
  --> lightway-app-utils/src/args/connection_type.rs:22:5
   |
22 |     pub fn is_tcp(&self) -> bool {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the lint level is defined here
```

lightway-server/src/config.rs
kp-antonio-yang: It will be a border topic I feel, I drop the commit first, and let the PR small enough.

And this is the reason why the commit shown at first.  We did not have validation function which can cross the repos, so we use `NonZeroDuration` to do the validation stuff cross repo and avoid the redundant code.  The validation introduced in the PR, it was good timing to retire it on server side if there were no discussion.

lightway-server/src/lib.rs
kp-antonio-yang: Agreed.   

I love to see what is changed or not by the previous syntax, but why not if we can have a smaller codes.  :smile:

lightway-server/src/config.rs
kp-antonio-yang: No, the validation is not a contract.   No mater we have the reusable `valiate*()` functions or not, now anyone reimplementing `lightway-server/src/main.rs` still must remember to implement functions/statements for validating.

 It is a reusable helper and it is better than rules only shown in docs.  And developer can easy to apply part of it by `validate_without`.

lightway-server/src/lib.rs
xv-thomas-leong: Can we just keep `mod auth;` in lightway-server/src/main.rs?

lightway-server/src/auth.rs
xv-thomas-leong: Why do we want to move the `auth` mod to lib?

lightway-app-utils/src/args/connection_type.rs
xv-thomas-leong: In this case, I'd suggest using `#[allow(missing_docs)]` instead

lightway-server/src/main.rs
```diff
@@ -86,15 +86,7 @@ async fn main() -> Result<()> {
...
-    validate_configuration_file_path(&config.server_key, Validate::OwnerOnly)
-        .with_context(|| format!("Invalid server key file {}", config.server_key.display()))?;
-    validate_configuration_file_path(&config.server_cert, Validate::AllowWorldRead)
-        .with_context(|| format!("Invalid server cert file {}", config.server_cert.display()))?;
-
-    if let Some(user_db) = &config.user_db {
-        validate_configuration_file_path(user_db, Validate::OwnerOnly)
-            .with_context(|| format!("Invalid user db file {}", user_db.display()))?;
-    }
+    config.validate()?;
```
xv-thomas-leong: I don't see much benefits by collecting these methods. On the contrary, it made the code more complicated than it has to too. 

Can we keep it as it was before?

lightway-server/src/lib.rs
kp-antonio-yang: It is required.  Is there any concern to public this mod?

lightway-server/src/auth.rs
kp-antonio-yang: If we want the function reusable, it should be public, and then input parameters also need public.

lightway-server/src/main.rs
kp-antonio-yang: Agreed, `user_db`, `server_key` and `server_cert` are keep as before.

lightway-server/src/lib.rs
kp-antonio-yang: Sorry about that it is not really required, the ServerAuth is already public.

lightway-server/src/auth.rs
kp-antonio-yang: Sorry about this, the input parameters, `ServerAuth` is from other crate.

lightway-server/src/config.rs
kp-antonio-yang: Only `validate` now, due to the files part check keep in the main.

lightway-app-utils/src/args/connection_type.rs
kp-mariappan-ramasamy, line 1: Typo in commit message `expressland` -> `expresslane`
in `server: config: validate enable_expresslane`

lightway-server/src/lib.rs
kp-mariappan-ramasamy, line 1: There were some typos in the commit message of `server: config: validate when ServerConfig build`

Could be simplified to

```
Move the config validation to try_* method, to make downstream apps also run validation
```

kp-mariappan-ramasamy (approved):

PR #452, opened 2026-06-11
CVPN-2505: Network change monitor

kp-weiwen-goh:
## Description

Exposes the network change detection logic within route_manager in app_utils. Multiple consumers can now subscribe to the change notification. route_manager is changed to subscribe to this monitor task instead.

Users will pass a network_change_signal (from `NetworkChangeManager::subscribe()`, or their own implementation) if they wish to own their own NetworkChangeManager. If none is provided, the client will create its own and use it.

## Motivation and Context

Previously, the network change detection was done only within route_manager. To allow consumers to also reuse this logic, this is exposed in app_utils instead.

## How Has This Been Tested?

Change is broadcasted

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1579,6 +1579,7 @@ dependencies = [
  "metrics",
  "pnet_packet",
  "regex",
+ "route_manager",
  "schemars 1.2.1",
  "serde",
  "serde_with",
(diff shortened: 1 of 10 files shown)


kp-shawn-thiah (approved):
lgtm 👍

PR #453, opened 2026-06-11
CVPN-2609: Bound UDP batch receive iovecs by buffer capacity

kp-thomas-yau:
## Description
The UDP batch receive paths (client and server) advertised a fixed `MAX_OUTSIDE_MTU` iovec length to the kernel regardless of the destination buffer's actual spare capacity. Each iovec is now bounded by the buffer's spare capacity, so undersized buffers result in kernel-side truncation —matching the non-batch receive path — rather than writes past the buffer.

- **client**: iovecs bounded by spare capacity (buffers are sized from the configured `outside_mtu`, which can be below `MAX_OUTSIDE_MTU`); logs one warning per batch when `recvmmsg` reports `MSG_TRUNC`.
- **server**: same bounding, capped at `MAX_OUTSIDE_MTU`, with a `debug_assert` documenting the spare-capacity contract.

Note: current XNU doesn't report per-message `MSG_TRUNC` via `recvmsg_x` (despite its header docs), so truncation isn't observable on Apple platforms; this is documented in code and the related test assertions are Linux-gated.

## Motivation and Context
The fixed iovec length didn't always reflect the real buffer size on the client. Bounding by spare capacity makes the batch path safer for any buffer size, and the warning makes MTU misconfiguration diagnosable instead of silent packet loss.

## How Has This Been Tested?
New unit tests: truncation of oversized datagrams on client and server, and the client truncation warning (via `tracing-test`, Linux/Android only)

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp/batch_receive.rs
+++ b/lightway-client/src/io/outside/udp/batch_receive.rs
@@ -34,7 +34,7 @@ pub(crate) fn recv_multiple(
 mod apple {
     use bytes::BytesMut;
     use lightway_app_utils::recvmsg_x::{msghdr_x, recvmsg_x};
-    use lightway_core::{MAX_IO_BATCH_SIZE, MAX_OUTSIDE_MTU};
+    use lightway_core::MAX_IO_BATCH_SIZE;
     use std::{io, mem};
 
     pub(crate) struct RecvmsgX;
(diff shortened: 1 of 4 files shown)


lightway-client/src/io/outside/udp/batch_receive.rs
```diff
@@ -240,6 +262,65 @@ mod tests {
         assert_eq!(&bufs[0][..], &payload[..capacity]);
     }
 
+    /// Linux/Android only: current XNU never copies MSG_TRUNC into
+    /// recvmsg_x's per-message msg_flags, so the warning cannot fire on
+    /// Apple platforms.
+    #[cfg(any(linux, android))]
+    #[tokio::test]
+    #[tracing_test::traced_test]
```
kp-weiwen-goh, line 270: Is this needed? We can add it to all tests if we are going to include it as a dependency. There is also `test_log` which has more downloads

kp-thomas-yau: tracing-test here actually checks what logs are emitted here, so it is different from regular test_log

kp-weiwen-goh (approved):

PR #456, opened 2026-06-19
[CVPN-2539] remove IP addresses on TUN interface drop

xv-piotr-haber:
## Description
Tun-rs tries to open existing TUN interface with specified name if it already exists. But when adding IP address to it an error may occur if that address is already assigned to the interface. To allow lightway to re-use existing TUN interface remove all addresses on close, so when restarted lightway can assign them successfully again.
The error only occurs on windows.

## Motivation and Context
We want to be able to control TUN interface life cycle outside of lightway.

## How Has This Been Tested?
Created WinTUN interface in external program and started lightway to verify interface is used and usable.
Tested repeated use of the same interface by new lightway instances.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -251,6 +251,9 @@ impl TunConfig {
                         // Windows if destination provided create a default route with
                         // high priority
                         if cfg!(windows) {
+                            // remove address before adding it to prevent error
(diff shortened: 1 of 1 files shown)


kp-anuz-tomar (approved):
Seems okay to me.

PR #457, opened 2026-06-19
client: config: expose `try_from_*` methods

kp-antonio-yang:
## Description

This PR makes `lightway-client`'s configuration types reusable as a library, so downstream projects can build a client without duplicating the glue code that previously lived in `main.rs`:

- Expose the `config` module (`Config`, `ConnectionConfig`) from the `lightway-client` library and derive `struct_patch::Substrate`.
- Add `Config::validate()`, which rejects conflicting settings:
  - `sndbuf` / `rcvbuf` cannot be overridden in TCP mode — the kernel reserves these for autotuning (non-macOS)
  - `enable_pmtud` is only supported for UDP connections
  - `device_guid` must be a valid UUID (Windows)
  - `wintun_ring_capacity` must be a power of two between 128 KiB and 64 MiB (Windows)
- Add `ClientConfig::try_from_reload_sig_and_config()`, which validates a `config::Config` and converts it into a ready-to-use `ClientConfig`, including the `TunConfig` setup previously inlined in `main.rs`.
- Add `ClientConnectionConfig::try_from_event_handler_and_connection_config()`, which resolves the server address and converts a `config::ConnectionConfig` into a `ClientConnectionConfig` (replaces `make_client_connection_config` in `main.rs`).
- Move `From<&Config> for ReloadableClientConfig` from the binary into the library.
- Simplify `Config::server_dn` from `Option<String>` to `String` (an empty string means "not set").
- `main.rs` now just parses/patches the config and delegates to the new library methods.
- CI: drop `--enable-pmtud` from the TCP single-threaded e2e test, since the new validation rejects PMTUD in TCP mode.

## Motivation and Context

The logic that turns `Config` / `ConnectionConfig` into `ClientConfig` / `ClientConnectionConfig` was private to the client binary, so any downstream project embedding lightway-client had to copy it. Exposing these types and conversions from the library — together with upfront validation — makes them directly reusable. Covers CVPN-2605 and CVPN-2606.

## How Has This Been Tested?

- New unit tests covering each `Config::validate()` rule: default config, `sndbuf`/`rcvbuf` in TCP mode, device GUID parsing, PMTUD in TCP mode, and wintun ring capacity bounds.
- Existing e2e tests, with the TCP single-threaded test updated to stop passing `--enable-pmtud` (now rejected by validation).

## Types of changes

  - `config::Config::server_dn` is now `String` instead of `Option<String>`.
  - Configurations that previously enabled PMTUD over TCP, or overrode `sndbuf`/`rcvbuf` in TCP mode, are now rejected at startup.

## Checklist:


lightway-client/src/config.rs
```diff
@@ -62,7 +65,7 @@ pub struct Config {
     #[patch(attribute(clap(long)))]
     #[patch(attribute(doc = r#"Server domain name
     Only used if `servers` is empty"#))]
-    pub server_dn: Option<String>,
+    pub server_dn: String,
```
kp-mariappan-ramasamy, line 68: Why is this change needed ?

lightway-client/src/lib.rs
```diff
@@ -273,7 +273,8 @@ pub struct ClientConfig<ExtAppState: Send + Sync> {
 impl<ExtAppState: Send + Sync> ClientConfig<ExtAppState> {
     pub fn try_from_reload_sig_and_config(
         config_reload_signal: Option<mpsc::Receiver<ReloadableClientConfig>>,
-        config: config::Config) -> Result<ClientConfig<ExtAppState>> {
+        config: config::Config,
+    ) -> Result<ClientConfig<ExtAppState>> {
```
kp-mariappan-ramasamy, line 277: This change should be part of previous commit
`client: add try_from_event_handler_and_connection_config of ClientConnectionConfig`

lightway-client/src/main.rs
```diff
@@ -185,9 +153,14 @@ async fn main() -> Result<()> {
 
     let servers = config.take_servers()?;
 
-    let client_config = lightway_client::ClientConfig::<()>::try_from_reload_sig_and_config(config_reload_signal, config)?;
+    let client_config = lightway_client::ClientConfig::<()>::try_from_reload_sig_and_config(
+        config_reload_signal,
+        config,
+    )?;
```
kp-mariappan-ramasamy, line 159: Same as prev comment. This has to be moved to prev commit

lightway-client/src/main.rs
```diff
@@ -203,7 +176,9 @@ async fn main() -> Result<()> {
         }
     };
 
-    client(client_config, ctrlc_rx, conn_confs).await.map(|_| ())
+    client(client_config, ctrlc_rx, conn_confs)
+        .await
+        .map(|_| ())
```
kp-mariappan-ramasamy, line 181: Please move it to commit where it touches this line

lightway-client/src/main.rs
kp-antonio-yang: Thank you

lightway-client/src/lib.rs
kp-antonio-yang: Thank you

lightway-client/src/config.rs
kp-antonio-yang: This `Option` is redundant, because everything in patch already wrapped with option.
And also explain more details to you in direct message.

lightway-client/src/config.rs
```diff
@@ -384,6 +388,47 @@ impl Config {
         }
         Ok(std::mem::take::<Vec<ConnectionConfig>>(&mut self.servers))
     }
+
+    /// Ensure the config is validated, and alerted when there's a conflict in the settings.
+    pub fn validate(&self) -> anyhow::Result<()> {
+        #[cfg(not(macos))]
+        if self.sndbuf != DEFAULT_SNDBUF {
+            // TODO: with is_udp, when pr/451 ready
```
kp-mariappan-ramasamy: since this pr is merged, you can use is_udp and get rid of todo comment here
https://github.com/expressvpn/lightway/pull/451/changes#diff-d9550a479768c6c1a9d21fddc9a29a2240f747603d6ddbd6f17a76a461fdaea9R28

lightway-client/src/config.rs
```diff
@@ -384,6 +388,47 @@ impl Config {
...
+            anyhow::ensure!(
+                !self.mode.is_tcp(),
+                "Cannot be applied sndbuf; this is preserved in the kernel for autotuning purposes."
+            )
+        }
+        #[cfg(not(macos))]
+        if self.rcvbuf != DEFAULT_RCVBUF {
+            // TODO: with is_udp, when pr/451 ready
```
kp-mariappan-ramasamy: Same as above

lightway-client/src/config.rs
```diff
@@ -384,6 +388,47 @@ impl Config {
...
+        #[cfg(windows)]
+        if let Some(guid) = &self.device_guid {
+            anyhow::ensure!(
+                uuid::Uuid::parse_str(guid).is_ok(),
+                "device_guid must be a valid UUID (e.g. 550e8400-e29b-41d4-a716-446655440000)"
+            );
+        }
+        if self.enable_pmtud {
+            anyhow::ensure!(
+                !self.mode.is_tcp(),
```
kp-mariappan-ramasamy: Use is_udp here too

kp-mariappan-ramasamy (commented):
LGTM with minor cleanup comments

lightway-client/src/config.rs
kp-antonio-yang: Thank you.

kp-mariappan-ramasamy (approved):
LGTM thanks

PR #459, opened 2026-06-29
CVPN-2505: Ignore tunnel's own address in Windows network-change monitor

kp-weiwen-goh:
## Description

Ignore tunnel IP when detecting network change.

Take the tunnel address(es) in NetworkChangeMonitor::spawn and, on Windows, snapshot the system's relevant unicast addresses (excluding loopback, link-local and the tunnel itself) on each notification, signalling only when that filtered set actually changes. This mirrors the prefix==0 default-route filter already applied to route changes.

## Motivation and Context

The Windows NotifyAddrChange fallback fired on any address change, including the client's own TUN interface address being assigned during tunnel bring-up. Consumers that feed the signal into the client tear down Stream (TCP) connections on a network change, so on Windows every TCP connect was followed ~2s later by a spurious reconnect before any traffic flowed.

## How Has This Been Tested?

Verified network change sent correctly

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/Cargo.toml
+++ b/lightway-app-utils/Cargo.toml
@@ -52,6 +52,7 @@ windows-sys = { workspace = true, features = [
     "Win32_Foundation",
     "Win32_NetworkManagement",
     "Win32_NetworkManagement_IpHelper",
+    "Win32_NetworkManagement_Ndis",
     "Win32_Networking_WinSock",
     "Win32_System_IO",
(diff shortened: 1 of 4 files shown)


kp-anuz-tomar (commented):
I am not seeing any major issues from my limited experience.

kp-anuz-tomar (approved):

PR #461, opened 2026-07-02
server: give recv_gso a default Unsupported impl

kp-thomas-yau:
## Description

Make InsideIORecv::recv_gso a provided method that returns Unsupported instead of a required one. Inside IO backends that don't offer GSO reads (e.g. alternate/external ones) no longer have to implement it, while Tun keeps its real implementation. Update the doc comment to describe the default and frame the contract as what GSO-capable backends must uphold.

## Motivation and Context

Make recv_gso() fail by default if we don't implement it

## How Has This Been Tested?

NA

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/io/inside.rs
+++ b/lightway-server/src/io/inside.rs
@@ -15,6 +15,10 @@ pub trait InsideIORecv: Sync + Send {
 
     /// Receive a GSO superpacket into `buf`.
     ///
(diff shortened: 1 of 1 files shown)


kp-mariappan-ramasamy (approved):
LGTM

kp-samuel-tam (approved):
Thanks for the fix

PR #462, opened 2026-07-06
core: avoid losing inner frames on blocked TCP sends

xv-thomas-leong:
## Description
When the outside socket would block on a Stream (TCP) connection, inner frames were dropped with no recovery path. Queue them instead up to 64 packets, i.e. ~96 KiB at 1500 MTU, and flush on the next send.

Frames carrying inner TCP packets, and all frames on Datagram connections, retain only the in-flight head buffer (the TLS library requires retrying a blocked write with the same buffer); further frames are dropped as inner TCP retransmits the payloads anyway.

## How Has This Been Tested?

Manually tested

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -11,6 +11,7 @@ use bytes::{Buf, BufMut, Bytes, BytesMut};
 use dplpmtud::BASE_PLPMTU;
 use rand::distr::{Distribution, StandardUniform};
 use std::borrow::Cow;
+use std::collections::VecDeque;
 use std::net::AddrParseError;
 use std::num::{NonZeroU16, Wrapping};
 use std::{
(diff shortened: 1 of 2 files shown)


lightway-core/src/connection.rs
```diff
@@ -1348,30 +1361,54 @@ impl<AppState: Send> Connection<AppState> {
...
+    /// Everything else (Datagram connections, frames carrying inner
+    /// TCP packets) retains only the in-flight head buffer (the TLS
+    /// library requires retrying a blocked write with the same
+    /// buffer); further frames are dropped — inner TCP retransmits the
+    /// payloads anyway.
+    fn send_frame_or_queue(&mut self, frame: wire::Frame) -> ConnectionResult<()> {
+        let queue_limit = match self.connection_type {
+            ConnectionType::Stream => match &frame {
+                wire::Frame::Data(d) if ipv4_is_tcp(d.data.as_ref()) => 1,
+                wire::Frame::DataFrag(f) if f.offset == 0 && ipv4_is_tcp(f.data.as_ref()) => 1,
```
kp-mariappan-ramasamy: In Lightway TCP, DataFrag is not possible since we dont enable PMTUD, could we remove it  and handle only for Data frame?

xv-thomas-leong: Good catch, and updated 🙏

kp-mariappan-ramasamy (approved):
LGTM

PR #463, opened 2026-07-06
CVPN-2623: batch inside IO and flush UDP sends with sendmmsg

kp-thomas-yau:
## Description

Adds an opt-in batched send path to the UDP server, gated behind a new `--enable-batch-send` flag (default off):

- `InsideIORecvBatch` — a capability trait mirroring `InsideIORecvGso`: backends that can pop multiple packets per call (up to `MAX_IO_BATCH_SIZE`) override `as_batch`, and the batched loop accepts only the upgraded handle, so the capability check happens once at startup. Each backend sizes its own receive buffers from its MTU. The io_uring TUN backend batches via `mpsc::Receiver::recv_many`; the direct TUN backend reads one packet per call.
- A scoped send queue on the UDP server: while the inside loop processes a receive batch, per-connection sends are queued instead of issuing one `sendmsg` each, then flushed together — one `sendmmsg` covering the whole batch on Linux, a `sendmsg` loop elsewhere. A full send buffer is treated as backpressure: the flush waits for socket writability and resumes from the first unsent datagram, pausing the inside loop rather than dropping already-encrypted packets. A hard IO error drops only the datagram that caused it; the drop-guard fallback (taken only if the loop exits without flushing) stays best-effort single-attempt and is counted separately.
- The batched loop runs only for UDP servers — stream transports have no batched send path, so on TCP the flag is a no-op and the default loop runs. Combining it with `--enable-tun-offload` is rejected at startup, since the GSO loop already aggregates packets.
- New metrics, documented in `docs/logs_and_metrics.md`: `tun_recv_batch_size` and `udp_send_batch_size` histograms; `udp_send_batch_blocked`, `udp_send_batch_dropped` and `udp_send_batch_missed_flush` counters.
- E2E variants exercising the flag (UDP, UDP + batch receive, TCP no-op).

Also includes a fix for a pre-existing issue the new tests surfaced: `send_to_socket` always attached a control buffer, and macOS rejects a non-null `msg_control` with `msg_controllen == 0`, failing sends that carry no cmsg.

## Motivation and Context

The download path (tun → encrypt → UDP) costs one `sendmsg` syscall per packet. The receive direction already batches (`--enable-batch-receive`, `recvmmsg`/`recvmsg_x`); this mirrors that on the send side. Batches only form under backlog — `recv_buf_many` returns whatever is immediately available and waits only when idle — so batching adds no latency. The flag defaults off so the change can be A/B tested; the new metrics are the instruments for that comparison.

## How Has This Been Tested?

- Unit tests for the send queue (real socket pairs: window atomicity, in-order flush, batches larger than `MAX_IO_BATCH_SIZE`, window close, per-message pktinfo on Linux), the queue-aware `UdpSocket::send`, the `as_batch` capability upgrade, and the batched loop (scripted inside-IO mock).
- `cargo test -p lightway-server -p lightway-app-utils`, `cargo fmt --check`, and `cargo clippy --all-targets` all clean, developed on macOS; `cargo check`/`cargo clippy` additionally cross-checked for `aarch64-unknown-linux-gnu` to cover the Linux-only code.
- The Linux-only paths (`sendmmsg` flush, its pktinfo test, and the `io-uring` feature combination) need this repo's Linux CI to compile and run them.
- New e2e targets: `run-udp-server-batch-send-test`, `run-udp-server-all-batching-test`, `run-tcp-server-batch-send-test`.
- Throughput/syscall-count comparison under load is still to be done before considering a default change; the flag stays off.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/tun.rs
+++ b/lightway-app-utils/src/tun.rs
@@ -319,6 +319,24 @@ impl Tun {
         }
     }
 
+    /// Recv up to `max` packets from Tun, appending to `pkts` (each
+    /// buffer sized to the interface MTU by the backend). Waits only
+    /// when nothing is immediately available. Only the io_uring backend
+    /// returns more than one packet per call; the direct backend reads
+    /// a single packet.
+    #[cfg_attr(not(feature = "io-uring"), allow(unused_variables))]
+    pub async fn recv_buf_many(
+        &self,
+        pkts: &mut Vec<BytesMut>,
+        max: usize,
+    ) -> IOCallbackResult<usize> {
+        match self {
+            Tun::Direct(t) => t.recv_buf_many(pkts).await,
+            #[cfg(feature = "io-uring")]
+            Tun::IoUring(t) => t.recv_buf_many(pkts, max).await,
+        }
+    }
+
     /// Recv a GSO frame from `Tun` into `buf`, stripping and decoding the
     /// leading `virtio_net_hdr`.
     ///
@@ -453,6 +471,24 @@ impl TunDirect {
                 IOCallbackResult::WouldBlock
             }
             Err(err) => IOCallbackResult::Err(err),
+        }
+    }
+
+    /// Recv one packet from Tun, appending it to `pkts` as a buffer
+    /// sized to the interface MTU. The direct backend has no batched
+    /// read, so this is the single-packet counterpart of the io_uring
+    /// backend's `recv_buf_many`.
+    pub async fn recv_buf_many(&self, pkts: &mut Vec<BytesMut>) -> IOCallbackResult<usize> {
+        let mtu = self.mtu();
+        let mut buf = BytesMut::with_capacity(mtu);
+        buf.resize(mtu, 0);
+        match self.recv_buf(&mut buf).await {
+            IOCallbackResult::Ok(_n) => {
+                pkts.push(buf);
+                IOCallbackResult::Ok(1)
+            }
+            IOCallbackResult::WouldBlock => IOCallbackResult::WouldBlock,
+            IOCallbackResult::Err(e) => IOCallbackResult::Err(e),
         }
     }
 
@@ -652,6 +688,18 @@ impl TunIoUring {
                 *buf = pkt;
                 IOCallbackResult::Ok(len)
             }
+            Err(e) => IOCallbackResult::Err(std::io::Error::other(e)),
+        }
+    }
+
+    /// Recv up to `max` packets from Tun via the io_uring rx queue.
+    pub async fn recv_buf_many(
+        &self,
+        pkts: &mut Vec<BytesMut>,
+        max: usize,
+    ) -> IOCallbackResult<usize> {
+        match self.tun_io_uring.recv_many(pkts, max).await {
+            Ok(n) => IOCallbackResult::Ok(n),
             Err(e) => IOCallbackResult::Err(std::io::Error::other(e)),
         }
     }
--- a/lightway-server/src/io/inside.rs
+++ b/lightway-server/src/io/inside.rs
@@ -13,6 +13,12 @@ use std::sync::Arc;
 pub trait InsideIORecv: Sync + Send {
     async fn recv_buf(&self, buf: &mut bytes::BytesMut) -> IOCallbackResult<usize>;
 
+    /// Upgrade to the batch-receive interface, when this instance can
+    /// pop multiple packets per call. Default: not supported.
+    fn as_batch(self: Arc<Self>) -> Option<Arc<dyn InsideIORecvBatch>> {
+        None
+    }
+
     /// Upgrade to the GSO-capable interface, when this instance supports
     /// GSO reads. Default: not supported. Capability is per-instance —
     /// a TUN opened without offload returns `None`.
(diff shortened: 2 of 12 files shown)


lightway-server/src/io/outside/udp/send_queue.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+impl Drop for SendBatchGuard {
+    fn drop(&mut self) {
+        // Fallback for paths that exit without calling flush(), which
+        // takes the queue and leaves None here.
+        let msgs = self.queue.queue.lock().unwrap().take();
+        let Some(msgs) = msgs else { return };
+        metrics::udp_send_batch_flush(msgs.len());
+        if !msgs.is_empty() {
```
kp-mariappan-ramasamy, line 120: We could add a counter here to monitor whether we miss flush in some path ?

lightway-server/src/metrics.rs
```diff
@@ -57,6 +57,12 @@ static METRIC_UDP_RECV_INVALID_ADDR: LazyLock<Counter> =
     LazyLock::new(|| counter!("udp_recv_invalid_addr"));
 static METRIC_UDP_RECV_MISSING_PKTINFO: LazyLock<Counter> =
     LazyLock::new(|| counter!("udp_recv_missing_pktinfo"));
+static METRIC_UDP_SEND_BATCH_SIZE: LazyLock<Histogram> =
+    LazyLock::new(|| histogram!("udp_send_batch_size"));
+static METRIC_UDP_SEND_BATCH_DROPPED: LazyLock<Counter> =
+    LazyLock::new(|| counter!("udp_send_batch_dropped"));
+static METRIC_UDP_SEND_BATCH_BLOCKED: LazyLock<Counter> =
+    LazyLock::new(|| counter!("udp_send_batch_blocked"));
```
kp-mariappan-ramasamy, line 65: Update docs about these new metrics
https://github.com/expressvpn/lightway/blob/main/docs/logs_and_metrics.md

lightway-server/src/lib.rs
```diff
@@ -258,6 +258,11 @@ pub struct ServerConfig<SA: for<'a> ServerAuth<AuthState<'a>>> {
     /// Enable batch receive (`recvmsg_x` on macOS, `recvmmsg` on Linux)
     pub enable_batch_receive: bool,
 
+    /// Batch inside-path receives and flush the resulting UDP datagrams
+    /// with a single syscall where the platform supports it (UDP mode,
+    /// Linux). Default off.
+    pub enable_inside_io_batching: bool,
```
kp-mariappan-ramasamy: Would be consistent to rename it to `enable_batch_send`since the ideal goal is to use sendmmsg in UDP socket

lightway-server/src/io/outside/udp.rs
```diff
@@ -110,11 +111,20 @@ struct UdpSocket {
     sock: Arc<tokio::net::UdpSocket>,
     peer_addr: RwLock<(SocketAddr, SockAddr)>,
     reply_pktinfo: Option<libc::in_pktinfo>,
+    /// When inside-IO batching is enabled, sends are diverted into this
+    /// queue while a batch window is open. `None` when the feature is
+    /// off, keeping the existing path untouched.
```
kp-mariappan-ramasamy: This is implicit, guess its LLM thing. We could remove this ?

lightway-server/src/io/inside.rs
```diff
@@ -13,6 +13,34 @@ use std::sync::Arc;
...
+    /// available; when packets are already queued they return whatever
+    /// is ready without waiting, so batching adds no latency.
+    ///
+    /// The default implementation receives a single packet via
+    /// [`Self::recv_buf`], so existing backends work unchanged.
+    async fn recv_buf_many(
+        &self,
+        pkts: &mut Vec<bytes::BytesMut>,
+        _max: usize,
+        mtu: usize,
```
kp-mariappan-ramasamy: MTU is a tun property. We should consider removing from recv_many api ?
Or is it needed for any corner cases?

lightway-app-utils/src/tun.rs
```diff
@@ -316,6 +316,35 @@ impl Tun {
...
+                let mut buf = BytesMut::with_capacity(mtu);
+                buf.resize(mtu, 0);
+                match t.recv_buf(&mut buf).await {
+                    IOCallbackResult::Ok(_n) => {
+                        pkts.push(buf);
+                        IOCallbackResult::Ok(1)
+                    }
+                    IOCallbackResult::WouldBlock => IOCallbackResult::WouldBlock,
+                    IOCallbackResult::Err(e) => IOCallbackResult::Err(e),
+                }
```
kp-mariappan-ramasamy: Think we should move this part inside Tun::Direct instead of inlining ?

lightway-server/src/io/inside/tun.rs
```diff
@@ -56,6 +56,24 @@ impl InsideIORecv for Tun {
...
+    async fn recv_buf_many(
+        &self,
+        pkts: &mut Vec<BytesMut>,
+        max: usize,
+        mtu: usize,
+    ) -> IOCallbackResult<usize> {
+        let start = pkts.len();
+        match self.0.recv_buf_many(pkts, max, mtu).await {
+            IOCallbackResult::Ok(n) => {
+                for pkt in &pkts[start..] {
```
kp-mariappan-ramasamy: This `start` was tricky. 
Note to myself: Since pkts is a vector, could contain elements from previous batch too, so start from start.

Wonder whether we need to use static size instead of Vec. Will revisit after full review

lightway-server/src/metrics.rs
```diff
@@ -83,6 +83,8 @@ static METRIC_TUN_REJECTED_NO_CLIENT_IP: LazyLock<Counter> =
 // Traffic volume
 static METRIC_TUN_FROM_CLIENT: LazyLock<Counter> = LazyLock::new(|| counter!("tun_from_client"));
 static METRIC_TUN_TO_CLIENT: LazyLock<Counter> = LazyLock::new(|| counter!("tun_to_client"));
+static METRIC_TUN_RECV_BATCH_SIZE: LazyLock<Histogram> =
+    LazyLock::new(|| histogram!("tun_recv_batch_size"));
```
kp-mariappan-ramasamy, line 89: Same as above. Add the metrics in docs

lightway-server/src/io/outside/udp/send_queue.rs
kp-mariappan-ramasamy, line 1: Think we could squash `server: flush the whole send queue with one sendmmsg` into first commit where it added the SendQueue

lightway-server/src/io/outside/udp.rs
kp-mariappan-ramasamy, line 1: Squash `server: drop redundant send_queue field comment` into original commit pls

lightway-server/src/lib.rs
```diff
@@ -752,11 +751,21 @@ mod tests {
...
+        }
+    }
+
+    #[async_trait::async_trait]
+    impl InsideIORecvBatch for ScriptedInsideIO {
         async fn recv_buf_many(
             &self,
             pkts: &mut Vec<BytesMut>,
             _max: usize,
-            _mtu: usize,
```
kp-mariappan-ramasamy: Would be nicer to move this into the orignal commit

kp-mariappan-ramasamy (approved):
LGTM, Thanks

PR #467, opened 2026-07-08
client: config: warn per-server instead of erroring for sndbuf/rcvbuf…

kp-antonio-yang:
## Description
`Config::validate` now iterates over all configured servers (entries in
`servers` plus the top-level `server` field when non-empty) for `sndbuf`,
`rcvbuf`, and `enable_pmtud` compatibility checks. For each server using
TCP that cannot apply the setting, a `tracing::warn!` is emitted with the
server address. `validate()` always returns `Ok(())` for these checks so
the client can downgrade gracefully.

## Motivation and Context
Previously the client failed hard when `sndbuf`, `rcvbuf`, or
`enable_pmtud` were configured alongside a TCP server, even though the
client can simply skip those features for TCP connections and continue.
The old check also only looked at the top-level `mode` field, ignoring
entries in the `servers` list entirely.

## How Has This Been Tested?
Added unit tests in `config.rs` using `#[tracing_test::traced_test]`:
- `validate_sndbuf_on_tcp_server` — warns with server address, returns `Ok`
- `validate_rcvbuf_on_tcp_server` — warns with server address, returns `Ok`
- `validate_pmtud_on_tcp_server` — warns with server address, returns `Ok`
- `validate_default_config` — no warnings, returns `Ok`

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -391,19 +391,36 @@ impl Config {
 
     /// Ensure the config is validated, and alerted when there's a conflict in the settings.
     pub fn validate(&self) -> anyhow::Result<()> {
+        let mut all_servers: Vec<(&str, ConnectionType)> = self
+            .servers
+            .iter()
+            .map(|s| (s.server.as_str(), s.mode))
+            .collect();
+        if !self.server.is_empty() {
+            all_servers.push((self.server.as_str(), self.mode));
+        }
+
         #[cfg(not(macos))]
         if self.sndbuf != DEFAULT_SNDBUF {
-            anyhow::ensure!(
-                self.mode.is_udp(),
-                "Cannot be applied sndbuf; this is preserved in the kernel for autotuning purposes."
-            )
+            for (server, mode) in &all_servers {
+                if mode.is_tcp() {
+                    tracing::warn!(
+                        server,
+                        "sndbuf is set but cannot be applied to this TCP connections"
+                    );
+                }
+            }
         }
         #[cfg(not(macos))]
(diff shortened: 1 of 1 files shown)


lightway-client/src/config.rs
```diff
@@ -391,19 +391,46 @@ impl Config {
...
+                    );
+                }
+            }
             anyhow::ensure!(
-                self.mode.is_udp(),
-                "Cannot be applied sndbuf; this is preserved in the kernel for autotuning purposes."
-            )
+                all_servers.iter().any(|(_, m)| m.is_udp()),
+                "sndbuf cannot be applied; no server uses UDP"
+            );
```
kp-mariappan-ramasamy: We don't have to error out here. just display warning above should be fine.

kp-mariappan-ramasamy: Same for rcvbuf and pmtud.

lightway-client/src/config.rs
```diff
@@ -391,19 +391,46 @@ impl Config {
...
+            .map(|s| (s.server.as_str(), s.mode))
+            .collect();
+        if !self.server.is_empty() {
+            all_servers.push((self.server.as_str(), self.mode));
+        }
+
         #[cfg(not(macos))]
         if self.sndbuf != DEFAULT_SNDBUF {
+            for (server, mode) in &all_servers {
+                if !mode.is_udp() {
```
kp-mariappan-ramasamy: we have mode.is_tcp api available

kp-mariappan-ramasamy (approved):

PR #468, opened 2026-07-09
CVPN-2621: Fix race condition where expresslane self keys becomes None

kp-mariappan-ramasamy:
## Description

When expresslane config ack message is received, handler flow tries to promote next self key to current. But this it not idempotent. i.e when a duplicate message is received, it replaced the valid key with a null one.

Fixed it by checking whether it is a valid key 

## Motivation and Context

Some clients went to Connect Cannot Browse with following errors:

`Failed to process outside data: Protocol Error: Insufficient data`

## Types of changes

## Checklist:


kp-anuz-tomar (approved):
Looks like simple checks.

PR #472, opened 2026-07-10
ci: build and publish Android AAR nightly via GitHub Actions

kp-antonio-yang:
## Description

Adds an automated Android AAR build and nightly release to CI:

- **New `build-android-aar` job in `ci.yaml`** — runs only on pushes to `main`, and only when mobile-related paths changed (`lightway-client`, `lightway-core`, `lightway-app-utils`, `uniffi-bindgen`, `Cargo.toml`, `Cargo.lock`, `Makefile.toml`), as determined by a `paths-filter` job. It frees runner disk space, installs Nix (Determinate installer + magic-nix-cache), builds the AAR inside the Android devshell (`nix develop .#android -c build`: cargo-ndk for x86_64 / x86 / armeabi-v7a / arm64-v8a, uniffi Kotlin bindgen, Gradle AAR bundling), generates the lightway-client config JSON schema, uploads both as workflow artifacts, and publishes them to the `android-nightly` prerelease.
- **CI gate rework** — the `ci` gate job now requires `nix-mobile-check` directly and tolerates `skipped` results, since the AAR build is conditional and must not fail the gate when filtered out. `nix-mobile-check` (check / clippy / test with mobile features) still runs on every PR.
- **Trimmed the Android Nix devshell** (`nix/modules/devshells.nix`) to fit GitHub-hosted runners: `includeEmulator = false`, dropped the duplicate `androidenv.androidPkgs` SDK/NDK packages and the rust-analyzer/rust-src extensions, pointed `ANDROID_NDK_HOME` at the composed SDK's `ndk-bundle`, and updated the pinned `sdk.dir` store path in `local.properties`.
- **`lightway-app-utils/build.rs`** — temporary `#![allow(semicolon_in_expressions_from_macros)]` for cfg_aliases 0.2.1 tripping a nightly deny-by-default lint, which broke the test-miri job.

## Motivation and Context

Android developers currently build the AAR by hand from a local devshell. This automates it: every mobile-related change merged to `main` refreshes the `android-nightly` prerelease with a downloadable AAR + JSON schema that downstream apps can consume directly. The devshell trim is required because the full Android SDK closure (emulator + system images) does not fit on `ubuntu-latest` runners.

## How Has This Been Tested?

- Built the AAR locally with `nix develop .#android -c build` and verified `lightway-release.aar` is produced at the expected path.
- The AAR job is gated to `main`, so it does not run on this PR; earlier revisions of this branch ran it as a PR job and produced the `.aar` and `.schema.json` artifacts successfully. The first `main`-gated run should be verified after merge.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   ci:
-    needs: [earthly, e2e, coverage, nix-mobile-check, nix-build, nix-tests, windows-build]
+    needs: [earthly, e2e, coverage, nix-mobile-build, nix-build, nix-tests, windows-build]
     runs-on: ubuntu-latest
     if: always()
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-android-aar.yaml
```diff
@@ -0,0 +1,99 @@
...
+jobs:
+  build-android-aar:
+    name: Build Android AAR
+    if: github.repository == 'expressvpn/lightway'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@v1.3.1
```
xv-thomas-leong: For all reusable actions that do not use an immutable release, please pin them to a commit hash instead of a version tag.

.github/workflows/build-android-aar.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - '.github/workflows/build-android-aar.yaml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref_name != 'main' }}
+
+jobs:
+  build-android-aar:
+    name: Build Android AAR
+    if: github.repository == 'expressvpn/lightway'
```
xv-thomas-leong:
```suggestion

```

.github/workflows/build-android-aar.yaml
xv-thomas-leong, line 1: If there's no special reason, please merge this file into `ci.yaml`

.github/workflows/build-android-aar.yaml
kp-antonio-yang: Just follow what we are doing in other existing CI task.
https://github.com/expressvpn/lightway/blob/c3fca39f8f75ba13952339b6e4eaf93239481fe6/.github/workflows/nix-build.yaml#L213

Why we need the hash specially here?  

A lot actions using version in current CI, Just list some here


```
 * ci.yaml
│ 186:       - uses: jwalton/gh-find-current-pr@v1                                                                                                            │
│ 192:         uses: peter-evans/find-comment@v3                                                                                                              │
│ 200:         uses: peter-evans/create-or-update-comment@v4                                                                                                 
```

.github/workflows/build-android-aar.yaml
kp-antonio-yang: Yes, that is doable. There will be a `mobile-related-changes` step in place, so the `.aar` will always be released whenever a relevant change is detected.

.github/workflows/build-android-aar.yaml
xv-thomas-leong: Fair point. In general, we would like reusable actions to be pinned. We aren't doing that now in other places but we should probably start doing so. I'm okay with what you have here, but can you do me a favor and pin them in a separate PR?

kp-antonio-yang:
It is integrated to CI.yml, so we can easy to see if there is a mobile release or not.

.github/workflows/ci.yaml
```diff
@@ -386,11 +383,81 @@ jobs:
...
+        with:
+          filters: |
+            android:
+              - 'lightway-client/**'
+              - 'lightway-core/**'
+              - 'lightway-app-utils/**'
+              - 'uniffi-bindgen/**'
+              - 'Cargo.toml'
+              - 'Cargo.lock'
+              - 'Makefile.toml'
```
xv-thomas-leong: Please remove this if there is no special reason behind. Typically we want a build for every commit

.github/workflows/ci.yaml
```diff
@@ -424,6 +491,13 @@ jobs:
           cp artifacts/lightway-client-x86_64-pc-windows-msvc/lightway-client.exe release-assets/lightway-client-x86_64-pc-windows-msvc.exe
           cp artifacts/lightway-client-aarch64-pc-windows-msvc/lightway-client.exe release-assets/lightway-client-aarch64-pc-windows-msvc.exe
 
+          if [ -d "artifacts/lightway-android-aar" ]; then
+            cp artifacts/lightway-android-aar/lightway-release.aar release-assets/lightway-release.aar
+          fi
+          if [ -d "artifacts/lightway-client-jsonschema" ]; then
+            cp artifacts/lightway-client-jsonschema/client.schema.json release-assets/client.schema.json
+          fi
```
xv-thomas-leong: This won't be necessary since we will have a build per commit

xv-thomas-leong (approved):

PR #479, opened 2026-07-17
ci: pin GitHub Actions to commit hashes instead of mutable tags

kp-antonio-yang:
## Description

Replace all GitHub Actions version tag references (e.g. `@v4`, `@stable`) with exact commit SHAs or immutable version tags across all workflow files. Original version tags are retained as inline comments for readability and maintainability.

Affected workflows:
- `.github/workflows/ci.yaml`
- `.github/workflows/nightly-cargo-deny.yaml`
- `.github/workflows/nightly-rust-update-check.yaml`
- `.github/workflows/nix-build.yaml`
- `.github/workflows/weekly-cargo-update.yaml`

## Motivation and Context

Mutable tags (e.g. `actions/checkout@v4`) can be silently redirected to a different commit by a compromised or malicious maintainer, introducing supply chain attack vectors into CI. Pinning to immutable commit SHAs ensures the exact code that was reviewed is always what runs, regardless of upstream tag changes.  Ref: https://github.com/expressvpn/lightway/pull/472#discussion_r3592550874

## How Has This Been Tested?

The workflow files were updated and verified to reference valid commit SHAs for each action. No logic changes were made — only the action reference format changed — so existing CI behavior is preserved.

## Types of changes

- [X] No Breaking change
> Note: Not a breaking change in the traditional sense, but any future attempts to use mutable tags will be inconsistent with this convention.

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,52 +48,52 @@ jobs:
     env:
       FORCE_COLOR: 1
     steps:
-      - uses: earthly/actions-setup@v1
+      - uses: earthly/actions-setup@43211c7a0eae5344d6d79fb4aaf209c8f8866203 # v1
         with:
           version: ~v0.8
           github-token: ${{ secrets.GITHUB_TOKEN }}
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4.3.1
       - name: Run +${{ matrix.target }} on Earthly
         run: earthly --ci --output ${{ matrix.privileged == true && '--allow-privileged' || '' }} +${{ matrix.target }}
 
       - name: Upload lightway-client artifact (x86_64)
         if: matrix.target == 'build'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v4.6.2
         with:
           name: lightway-client-x86_64-unknown-linux-gnu
           path: target/release/lightway-client
 
       - name: Upload lightway-server artifact (x86_64)
         if: matrix.target == 'build'
-        uses: actions/upload-artifact@v4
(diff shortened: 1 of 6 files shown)


kp-thomas-yau (commented):
Is any of the github action here is using immutable release? If it's using immutable release, I think it's ok to use the version number/tag for those

kp-antonio-yang:
Pinning the hash of a CI action is a recognized practice for preventing supply chain attacks, and I agree with it in principle. Version tags are also acceptable to me.

That said, the real security benefit comes from actually reading and auditing the scripts inside the actions after pinning — simply pinning the hash moves us toward the goal but does not provide any concrete security benefit on its own until that review is done.

kp-thomas-yau:
> Pinning the hash of a CI action is a recognized practice for preventing supply chain attacks, and I agree with it in principle. Version tags are also acceptable to me.
> 
> That said, the real security benefit comes from actually reading and auditing the scripts inside the actions after pinning — simply pinning the hash moves us toward the goal but does not provide any concrete security benefit on its own until that review is done.

I know, but GitHub released something called [immutable release](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) that causes Git tags cannot be moved, effectively pinning the Git tag to the same hash. So if the GitHub Action workflow is using this immutable release, we can just use the tag. (and we can of course use commit hash for workflows that isn't using this feature)

kp-antonio-yang:
Yeh, now all the actions pin with immutable version tags or immutable hashes.

kp-thomas-yau (dismissed):
noice

kp-thomas-yau (approved):

PR #481, opened 2026-07-20
build: pin stable Rust version via rust-toolchain.toml

kp-thomas-yau:
## Description

The stable Rust version lived only in the Earthfile's base image tag. Local builds were left unpinned, and the nix flake floated on rust-overlay's latest stable, so the three could silently drift onto different compilers.

Add a rust-toolchain.toml that rustup honours for both local builds and the Earthfile image, and pin the nix stable toolchain to that same channel (rustLatest -> rustStable) rather than tracking latest. The nightly updater now rewrites the toml alongside the Earthfile and refreshes the rust-overlay input in flake.lock, so nix can resolve the freshly pinned version within the same PR.

## Motivation and Context

Earthly, local build and nix build uses the same rust compiler

## How Has This Been Tested?

Passed CI and tested building lightway-client/core locally

## Types of changes

## Checklist:


kp-anuz-tomar (approved):
I don't see anything here which looks out of place. Seems like regular update stuff.

PR #482, opened 2026-07-20
client: Get socket send & receive buffer size

kp-thomas-yau:
## Description

Get outside socket's send/receive buffer size via socket2's function.

## Motivation and Context

Easier debugging + compare differences when setting different buffer size (to verify that setting the socket buffer size ACTUALLY changes the size at all)

## How Has This Been Tested?

NA

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -86,6 +86,15 @@ impl OutsideIO for Udp {
         Ok(())
     }
 
+    fn send_buffer_size(&self) -> Result<usize> {
+        let socket = socket2::SockRef::from(&self.sock);
+        Ok(socket.send_buffer_size()?)
+    }
+    fn recv_buffer_size(&self) -> Result<usize> {
+        let socket = socket2::SockRef::from(&self.sock);
+        Ok(socket.recv_buffer_size()?)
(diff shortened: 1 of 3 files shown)


xv-thomas-leong (approved):

lightway-client/src/io/outside/udp.rs
```diff
@@ -86,6 +86,15 @@ impl OutsideIO for Udp {
         Ok(())
     }
 
+    fn send_buffer_size(&self) -> Result<usize> {
+        let socket = socket2::SockRef::from(&self.sock);
```
kp-mariappan-ramasamy, line 90: Could also be nicer if we print/warn the set value differs from actual value.
something like:

```rust
fn set_send_buffer_size(&self, size: usize) -> Result<()> {
        let socket = socket2::SockRef::from(&self.sock);
        match socket.set_send_buffer_size(size) {
        Err(e) => {
            tracing::warn!("Failed to set UDP send buffer size to {size}: {e}");
        }
        Ok(_) => if self.send_buffer_size() != size {
           tracing::warn!("UDP send buffer size truncated to {self.send_buffer_size()}");
        }
}
        Ok(())
    }
```

PR #487, opened 2026-07-24
client: cli: allow disable keepalive_continuous via CLI

kp-antonio-yang:
# client: cli: allow disable keepalive_continuous via CLI

## Description

Renames the `keepalive_continuous` config field to `disable_keepalive_continuous` and exposes it as a `--disable-keepalive-continuous` CLI flag. The internal `continuous_keepalive` value is derived by negating this field, so the runtime behaviour is unchanged when the flag is not set.

## Motivation and Context

Continuous keepalive is enabled by default. The previous field name (`keepalive_continuous`) implied an opt-in, but since the feature is always on there was no CLI way to turn it off. Inverting the flag to `--disable-keepalive-continuous` makes the default state implicit and gives operators a clear opt-out when they need to disable the feature.

**Why not `--keepalive-continuous=false`?**

The `LW_CLIENT_*` environment variable convention is the reason. When `LW_CLIENT_KEEPALIVE_CONTINUOUS` is unset, it is generally interpreted as false, which would inadvertently disable continuous keepalive. For a feature that is enabled by default, a `--disable-*` flag is the more correct and predictable design.

## How Has This Been Tested?

- Verified that the `Default` impl still produces `continuous_keepalive = true` (i.e. `disable_keepalive_continuous = false`).
- Manually confirmed that passing `--disable-keepalive-continuous` on the CLI sets `continuous_keepalive = false` in the resulting `ClientConfig`.
- Existing unit and integration test suites pass with no changes required.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -151,8 +151,8 @@ pub struct Config {
     #[patch(attribute(clap(long)))]
     #[patch(empty_value = false)]
     #[patch(attribute(serde(default)))]
-    #[patch(attribute(doc = "Enable continuous Keepalive"))]
-    pub keepalive_continuous: bool,
+    #[patch(attribute(doc = "Disable continuous keepalive"))]
+    pub disable_keepalive_continuous: bool,
 
     #[patch(attribute(clap(long)))]
     #[patch(attribute(doc = r#"Time it takes to trigger a tracer packet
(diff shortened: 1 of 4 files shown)


kp-mariappan-ramasamy:
@kp-antonio-yang Do we need this, since we are planning to move away from clap and use config file ?

kp-antonio-yang:
If there is already an open PR moving away from clap and using config file/environment variable only, we can simply reference it here and close this one. It is good practice to clearly point to the solution that supersedes this PR, making it easy for anyone to verify that the issue has been fully resolved.

kp-antonio-yang:
This PR might be closed after config grouping, else the issue still here, and better not ignored.

PR #489, opened 2026-07-27
CVPN-2671: Downgrade inside packet codec if stalled

kp-weiwen-goh:
## Description

Adds a client-side data-plane stall guard for the inside packet codec.
New `ClientConfig::inside_pkt_codec_stall_timeout` param defines the timeout.

When the codec is active (encoding enabled) but no decoded packet has reached the inside/TUN path within a configurable timeout, the client requests disabling the codec (EncodingRequest{enable: false}) so traffic falls back to unencoded, instead of tearing the tunnel down. The connection keeps running and the data plane recovers.

- New Connection::downgrade_inside_pkt_codec_if_stalled(timeout) (lightway-core) encapsulates the check + downgrade. It self-guards: it only acts while encoding is enabled and no encoding request is already in flight, so a stall episode issues a single disable request.
- New ConnectionActivity::last_data_delivered_to_inside, updated only when a decoded packet is actually delivered to the inside path. Control frames (keepalive ping/pong) deliberately do not touch it, so it reflects real data-plane delivery.
- inside_io_task (lightway-client) calls the guard each iteration with the configurable ClientConfig::inside_pkt_codec_stall_timeout (Duration::ZERO disables the check).

## Motivation and Context

A failing inside packet codec can silently black-hole the data plane: it accepts outbound packets but never emits them. Keepalive and the tracer cannot detect this, because keepalive ping/pong are sent as raw `wire::Frame::Ping` control frames that bypass the codec. The tunnel looks perfectly healthy while every user packet is dropped.

Rather than tear down the tunnel (which false-positives on legitimately one-way / bursty-outbound traffic), it gracefully downgrades to unencoded and lets traffic recover. Idle tunnels are never affected: the check only runs while outbound traffic is flowing, since it is evaluated in the inside-read loop.

## How Has This Been Tested?

Tests with a blackhole packet adapter (drops every packet)

## Types of changes

## Checklist:

### DIFF
--- a/docs/inside_packet_codec.md
+++ b/docs/inside_packet_codec.md
@@ -142,3 +142,16 @@ to encoding requests with its id higher or equal to the id of the most recently 
 
 The number of re-transmission is limited by a constant to avoid wasting network resources in case the server does not have encoding enabled. If the client
 cannot receive a response after the maximum number of re-transmissions has been attempted, the request is considered as failed.
+
+## Data-Plane Stall Detection
+A faulty codec can silently black-hole the data plane: its encoder accepts inside packets (`PacketAccepted`) but never emits the encoded packets, so nothing
+reaches the tunnel. Keepalive cannot detect this, because keepalive ping/pong are control frames that bypass the codec and keep flowing, so the connection
+still looks alive while every inside packet is dropped.
+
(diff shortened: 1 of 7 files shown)


kp-antonio-yang (approved):
Looks good for me.

PR #491, opened 2026-07-28
Add Boringssl Integration

kp-ben-hung:
## Description
Added new feature flags, backend and CI changes required for support for boringssl.
The boringssl crate mirrors wolfssl's APIs, and re-export them. The rest of the crate imports the appropriate APIs from the feature-gated facade crate.  

## Motivation and Context

Add boringssl integration to the project.

## How Has This Been Tested?

Cross-comp. tests in E2E tests are added.

## Types of changes

Previously, Datagram traffic accepts both DTLS1.2 and DTLS1.3, this PR removed the support for DTLS1.2.

## Checklist:


kp-thomas-yau (commented):
I think we might need to update README or a doc to explain Lightway supports WolfSSL  (primarily) and BoringSSL as well, and explain how it works.

kp-ben-hung:
Thanks for the suggestion. I have updated the docs, explaining more on the alternative TLS backend. (and how the compile time switch works).

lightway-core/src/context.rs
```diff
@@ -396,8 +396,7 @@ impl<AppState> ServerContextBuilder<AppState> {
     ) -> ContextBuilderResult<Self> {
         let protocol = match connection_type {
             ConnectionType::Stream => crate::tls::Method::TlsServerV1_3,
-            // `crate::tls::Method::DtlsServer` supports both DTLS 1.2 and 1.3
-            ConnectionType::Datagram => crate::tls::Method::DtlsServer,
```
kp-mariappan-ramasamy, line 400: Split this into a separate commit. 
This is breaking change and needs to be highlighted separately

kp-ben-hung: Sure, I have separated it into another commit. I still want the original commit to be compilable so i put a feature gate there instead, and only fully remove it in the latest commit. I also removed mentions of DTLS1.2 support from the docs. Thanks for the suggestion.

lightway-boring/src/context.rs
```diff
@@ -0,0 +1,395 @@
...
+            }
+            Method::DtlsClientV1_3 | Method::DtlsServerV1_3 => {
+                SslContext::builder(SslMethod::dtls())
+            }
+        }
+        .map_err(|e| NewContextBuilderError(e.to_string()))?;
+
+        if matches!(method, Method::DtlsClientV1_3 | Method::TlsClientV1_3) {
+            builder.set_verify(SslVerifyMode::PEER);
+        }
```
kp-antonio-yang, line 52: Do we need a warn message or `unreachable!` in a else here?  Just prevent if we do in some mode in the future and without verify.

lightway-boring/src/context.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        Ok(self)
+    }
+
+    /// Set the cipher list (no-op for TLS 1.3 / DTLS 1.3).
+    ///
+    /// BoringSSL uses secure defaults for TLS 1.3 ciphersuites.
+    /// This method exists only for API compatibility with the WolfSSL backend.
+    pub fn with_cipher_list(self, _cipher_list: &str) -> Result<Self> {
+        Ok(self)
```
kp-antonio-yang, line 209: It might be good to have log here, rather than just `OK()`.

If there are something in `cipher_list` as expected, 
do info log "BoringSSL uses  TLS 1.3 ciphersuites",
else do warn log "BoringSSL ONLY uses own TLS 1.3 ciphersuites"

kp-antonio-yang (changes requested):

lightway-boring/src/context.rs
kp-ben-hung: Setting `SslVerifyMode::PEER` on server is the *incorrect* behavior here. If set on the server, the server will ask the client for certificate, which means a mTLS setup, which is not what we are doing. By the way, you are looking at the old change, in the newer change i refactored the line to `method.is_client()`. This will ensure future changes to the `Method` enum will not cause silent misclassification here. (the `matches!` is not exhaustive here after all)

lightway-boring/src/context.rs
kp-ben-hung: Agreed, i have updated the function to emit a warning on no-op. Additionally, since we added shims for TLS1.2, this function will not be no-op anymore on TLS1.2. I have added the logic for that as well.

kp-antonio-yang (dismissed):
I have no objections to this PR, though I'll admit my knowledge in this area is limited. It is a large change, so just want to make sure everyone who can review it has had a chance to do so.

kp-antonio-yang (approved):
This is a large PR that also adds a CI task, so there will likely be conflicts with [#508](https://github.com/expressvpn/lightway/pull/508). The PR looks good to me — and since #508 is still in draft, there is no rush to resolve the conflicts just yet. 😄 Happy to help sort out the CI side if needed.

PR #493, opened 2026-07-29
CVPN-2682: connected outside UDP socket on Apple platforms

kp-thomas-yau:
## Description

Adds `--enable-connected-udp` (off by default, Apple platforms only): `connect()` the outside UDP socket to the server so sends and receives skip the per-packet route lookup and PCB work. The socket-side machinery is gated on `apple` (macOS, iOS/iPadOS incl. Mac Catalyst, tvOS); the transition monitoring below is macOS-specific.

A connected socket pins the route and source address chosen at connect time, so the series also covers recovery:

- Network-change reactions run in a single coordinator task, in order: refresh the /32 server route, re-`connect()` the socket (same fd and local port, so fds handed to embedders stay valid), then — on a genuine transition, or (Apple, Datagram) whenever the server route was actually replaced — trigger the network-change keepalives, so the probe rides the freshly pinned path instead of racing it.
- The network-change monitor publishes *route* and *transition* signals separately. On macOS a transition is an address change (watched via [`netwatcher`](https://crates.io/crates/netwatcher)) plus a default-route *arrival* (add/change — the down-leg delete of a roam leaves nothing to probe) within a short window, so route-only churn updates routes without firing keepalives.
- If a send reports `EADDRNOTAVAIL` or a re-connect fails, the socket drops back to unconnected `send_to` (dissolving the association with `disconnectx(2)`) until the next network change restores the fast path. ICMP-derived errors on a connected receive are treated as transient instead of tearing the connection down.
- Swallowed outside-send errors are now logged (rate-limited to one line per second), and connect/downgrade/reconnect log the pinned local address, so a silently dead path shows up in debug logs.

The option works in every route mode, including `none`: the coordinator drives the reconnect regardless. On targets without the desktop route manager (iOS and friends), the network-change signal forwarder re-pins the socket before the keepalive probe instead. The uniffi mobile wrapper does not wire the option yet.

## Motivation and Context

Connecting the single-peer outside socket is a measurable throughput win on macOS. Without the recovery work, a network switch would silently blackhole the pinned socket until keepalive timeout — so the feature ships together with it, and stays opt-in.

Stacked on `client-keepalive` (the base branch): its resend-on-network-change fix plus the ordered refresh-then-probe here is what lets roams recover promptly.

## How Has This Been Tested?

- Unit tests: connected send/recv round-trip, re-connect keeps fd/port, downgrade and restore, ICMP unreachable non-fatal on receive, transition-detector pairing and rate-limiting, coordinator lifecycle (routes cleared exactly once on stop).
- Kernel behaviour verified with standalone probes on macOS: re-`connect()` re-selects route and source address on the same fd/port; `send_to` on a connected socket fails `EISCONN`, hence the explicit dissolve on fallback.
- Real Ethernet↔Wi-Fi roams with debug logging pinned down the ordering race this series fixes (the post-transition keepalive going out before the re-pin and dying silently); a roam re-test of the final ordering is pending.
- `cargo test`, `cargo clippy --all-targets` and `cargo fmt --check` clean on macOS (aarch64); `cargo check` clean for `aarch64-apple-ios` (including `--features mobile`) and Mac Catalyst (`aarch64-apple-ios-macabi`). Every commit in the series compiles standalone.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/network_change_monitor.rs
+++ b/lightway-app-utils/src/network_change_monitor.rs
@@ -1,24 +1,40 @@
 //! Reusable network-change monitor.
 //!
-//! Spawns the platform route/address listeners and republishes their events as a coalescing
-//! [`tokio::sync::watch`] signal. Consumers can [`subscribe`](NetworkChangeMonitor::subscribe) to
-//! network changes
+//! Spawns the platform route/address listeners and republishes their events as
+//! two coalescing [`tokio::sync::watch`] signals: *routes* (an applicable
+//! default route changed — the trigger for route-table fixups) and
+//! *transitions* (the machine moved networks: on macOS an address change and
+//! an applicable route *arrival* — add or change, not delete — within a short
+//! window; on Windows any detected change, since route changes are not always
+//! published there; never fired on platforms without an address listener).
 
 #[cfg(windows)]
 mod addr_monitor;
 
 use anyhow::Result;
-use route_manager::{AsyncRouteListener, RouteChange};
+use route_manager::{AsyncRouteListener, Route, RouteChange};
 use tokio::sync::watch;
 use tokio::task::JoinHandle;
 
 #[cfg(windows)]
 use addr_monitor::AsyncAddrListener;
 
-/// Monitors network changes (route and, on Windows, address changes) and
-/// publishes a notification each time one is detected.
+#[cfg(any(macos, test))]
+use std::time::{Duration, Instant};
+
+/// Generous pairing window: joining a network can assign the address seconds
+/// before the default route is installed.
+#[cfg(macos)]
+const TRANSITION_WINDOW: Duration = Duration::from_secs(15);
+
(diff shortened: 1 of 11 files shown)


lightway-app-utils/src/network_change_monitor.rs
```diff
@@ -74,11 +97,22 @@ impl NetworkChangeMonitor {
...
+                                   // roam) leaves nothing to probe yet.
+                                   #[cfg(macos)]
+                                   if !matches!(route_change, RouteChange::Delete(_))
+                                       && detector.on_route(Instant::now())
+                                   {
+                                       tracing::info!("Network transition detected");
+                                       let _ = transition_tx.send(());
+                                   }
+                                   #[cfg(windows)]
+                                   let _ = transition_tx.send(());
```
kp-antonio-yang, line 115: nit: might be good to have the same info log for windows to know if is_applicable_route or not.

kp-antonio-yang (approved):

PR #495, opened 2026-08-03
client: apply SO_MARK to outside UDP socket on Linux

kp-antonio-yang:
## Description

This is a focused split from #492 that introduces only the `SO_MARK` socket support, without the `RouteMode::Fwmark` policy routing mode or the associated `ip rule` management.

Changes:
- `lightway-app-utils`: new `sockopt::set_so_mark` / `get_so_mark` wrappers around the Linux `SO_MARK` socket option (Linux only, `CAP_NET_ADMIN` required)
- `lightway-client`: adds an optional `fwmark: u32` field to both `Config` and `ClientConfig` (Linux only, defaults to unset with 0 as convention)
- `lightway-client`: applies the mark to the outside UDP socket before it is used for anything, so no packet can escape unmarked; tolerates `setsockopt` failure with a warning; verifies the mark was accepted in `debug` builds

## Motivation and Context

Without a firewall mark the tunnel's own encrypted packets are steered by the host route to the VPN server. That host route is managed by `route_manager` and can be transiently absent during a network change (e.g. Wi-Fi roam), leaving a window where the server's IP matches the tunnel's catch-all route, causing an encapsulation loop.

Marking the outside socket with `SO_MARK` lets a policy routing rule (`ip rule fwmark <mark> lookup main`) keep the tunnel's own traffic on the WAN interface regardless of whether a host route is present. This PR ships the socket primitive needed for that, as a self-contained, non-breaking change that can be merged and deployed independently while the full policy routing mode (#492) is still under review.

## How Has This Been Tested?

- Confirmed `SO_MARK` is applied and read back correctly on Linux with `CAP_NET_ADMIN`.
- Confirmed graceful fallback (warning log, tunnel establishes) when running without `CAP_NET_ADMIN`.
- Confirmed no behavioural change on non-Linux platforms or with `feature = "mobile"` (parameter compiles out entirely).

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -212,6 +212,19 @@ pub struct Config {
     #[schemars(extend("x-cfg" = "desktop"))]
     pub route_mode: RouteMode,
 
+    #[cfg(linux)]
+    #[patch(attribute(serde(default)))]
+    #[patch(attribute(clap(long)))]
+    #[patch(
+        attribute(doc = r#"Firewall mark (SO_MARK) applied to the outside socket.
+        The tunnel's own encrypted packets carry this mark so policy routing
+        rules can keep them out of the tunnel.
+        In Linux networking, a fwmark of 0 represents the default unmarked state
+        and will not apply SO_MARK to the socket."#)
+    )]
+    #[schemars(extend("x-cfg" = "linux"))]
+    pub fwmark: u32,
+
     #[cfg(desktop)]
     #[patch(attribute(clap(long, value_enum)))]
     #[patch(attribute(doc = r#"DNS configuration mode
@@ -491,6 +504,8 @@ impl Default for Config {
             enable_batch_receive: false,
             #[cfg(desktop)]
             route_mode: RouteMode::default(),
+            #[cfg(linux)]
+            fwmark: 0,
             #[cfg(desktop)]
             dns_config_mode: DnsConfigMode::default(),
             log_level: LogLevel::Info,
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -21,7 +21,11 @@ pub struct Udp {
 }
 
 impl Udp {
-    pub async fn new(remote_addr: SocketAddr, sock: Option<UdpSocket>) -> Result<Self> {
+    pub async fn new(
+        remote_addr: SocketAddr,
+        sock: Option<UdpSocket>,
+        #[cfg(all(linux, not(feature = "mobile")))] fwmark: u32,
+    ) -> Result<Self> {
         let peer_addr = tokio::net::lookup_host(remote_addr)
             .await?
             .next()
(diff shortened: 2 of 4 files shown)


lightway-app-utils/src/sockopt/so_mark.rs
kp-thomas-yau, line 1: socket2 already have this built-in already: https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.set_mark

kp-antonio-yang:
@kp-weiwen-goh Some context for you.  This PR is request by @kp-mariappan-ramasamy, and split from the existing one, such that we can do quick response to a ticket.

lightway-client/src/io/outside/udp.rs
```diff
@@ -37,6 +41,38 @@ impl Udp {
...
+            let socket = socket2::SockRef::from(&sock);
+            if socket
+                .set_mark(fwmark)
+                .map_err(|e| {
+                    tracing::warn!("Fail to set so mark on socket: {}", e);
+                })
+                .is_ok()
+            {
+                tracing::info!("Applied firewall mark to outside socket");
+            }
```
kp-mariappan-ramasamy: Using match would have been less verbose for this case.

Also i think, we could also enable the next check without debug feature, so that it exits if it is not set properly.

kp-antonio-yang: Agreed, match is good.

kp-antonio-yang: Keeping the `check after write` only under the `debug` feature makes sense. 

The socket is created by this program and the mark is set by the same program, so there is no real need to verify it on a release build. Having the check in debug builds is a nice sanity guard, but enabling it in release would be like checking whether a file exists immediately after opening it ourselves.

kp-mariappan-ramasamy (approved):
LGTM

lightway-client/src/config.rs
```diff
@@ -212,6 +212,19 @@ pub struct Config {
...
+    #[patch(attribute(clap(long)))]
+    #[patch(
+        attribute(doc = r#"Firewall mark (SO_MARK) applied to the outside socket.
+        The tunnel's own encrypted packets carry this mark so policy routing
+        rules can keep them out of the tunnel.
+        In Linux networking, a fwmark of 0 represents the default unmarked state
+        and will not apply SO_MARK to the socket."#)
+    )]
+    #[schemars(extend("x-cfg" = "linux"))]
+    pub fwmark: u32,
```
kp-thomas-yau, line 226: very smol nit: Prefer this to be Option<NonZeroU32>, but should be fine still

kp-antonio-yang: There is a main PR for the issue, and we will think twice in that PR.

kp-antonio-yang: By the way, the user intent still be clearly reflected from Cli/ConfigFile/Envars in this PR.

The fields in these surface are already wrapped with `Option<>`, the `Config` is not a directly surface anymore,
so it is allowed if user want to leave it blank or apply a blank setting with `Some(0)`.

PR #497, opened 2026-08-04
client: apply SO_MARK to outside TCP socket on Linux

kp-antonio-yang:
## Description

Apply `SO_MARK` to the outside **TCP** socket on Linux (non-mobile), mirroring the UDP support added in #495.

For fresh connections the mark is set via `socket2` **before** `connect()` is called, so the SYN packet itself carries the mark — ensuring policy-routing rules take effect from the very start of the TCP handshake. For caller-supplied sockets the mark is applied immediately after hand-off.

An Earthfile `run-tcp-fwmark-test` target is added and included in `run-all-tests`.

## Motivation and Context

Without this change, using `--fwmark` in TCP mode left the outside socket unmarked, so policy-routing rules (e.g. those installed by `RouteMode::Fwmark`) had no effect on TCP tunnel traffic. The fix brings TCP to parity with UDP.

## How Has This Been Tested?

- `+run-tcp-fwmark-test` Earthly target exercises the new path end-to-end.
- Existing `+run-udp-fwmark-test` continues to pass.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/tcp.rs
+++ b/lightway-client/src/io/outside/tcp.rs
@@ -5,14 +5,61 @@ use tokio::net::TcpStream;
 
 use super::{OutsideIO, OutsideSocket};
 use lightway_core::{IOCallbackResult, OutsideIOSendCallback, OutsideIOSendCallbackArg};
+use socket2::{Domain, Protocol, Socket, Type};
 
 pub struct Tcp(tokio::net::TcpStream, SocketAddr);
 
 impl Tcp {
-    pub async fn new(remote_addr: SocketAddr, maybe_sock: Option<TcpStream>) -> Result<Self> {
+    pub async fn new(
+        remote_addr: SocketAddr,
+        maybe_sock: Option<TcpStream>,
+        #[cfg(all(linux, not(feature = "mobile")))] fwmark: u32,
+    ) -> Result<Self> {
         let sock = match maybe_sock {
-            Some(s) => s,
-            None => tokio::net::TcpStream::connect(remote_addr).await?,
+            Some(s) => {
+                #[cfg(all(linux, not(feature = "mobile")))]
+                if fwmark != 0 {
+                    let socket = socket2::SockRef::from(&s);
+                    match socket.set_mark(fwmark) {
+                        Ok(_) => tracing::info!("Applied firewall mark to outside TCP socket"),
+                        Err(e) => tracing::warn!("Failed to set SO_MARK on TCP socket: {}", e),
+                    }
+                }
+                s
+            }
+            None => {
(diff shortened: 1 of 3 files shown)


lightway-client/src/io/outside/tcp.rs
```diff
@@ -9,16 +9,73 @@ use lightway_core::{IOCallbackResult, OutsideIOSendCallback, OutsideIOSendCallba
...
+                s
+            }
+            None => {
+                #[cfg(all(linux, not(feature = "mobile")))]
+                {
+                    if fwmark != 0 {
+                        return Self::connect_with_fwmark(remote_addr, fwmark).await;
+                    }
+                }
+                TcpStream::connect(remote_addr).await?
```
kp-mariappan-ramasamy: It is error-prone to have 2 return paths for fwmark and not. i.e It is possible we add one socket option below and forget in other place.

Better to remove this `TcpStream::connect(remote_addr).await?` and call  `connect_with_fwmark` from here. Set fwmark only if it is set in the config. This would allow to keep setting in sync and avoid future tpos

kp-mariappan-ramasamy (approved):

kp-antonio-yang:
Thanks for reviewing.

lightway-client/src/io/outside/tcp.rs
```diff
@@ -5,14 +5,61 @@ use tokio::net::TcpStream;
...
+                match socket.connect(&addr) {
+                    Ok(()) => {}
+                    Err(e) if e.raw_os_error() == Some(libc::EINPROGRESS) => {}
+                    Err(e) => return Err(e.into()),
+                }
+
+                let std_stream: std::net::TcpStream = socket.into();
+                let stream = TcpStream::from_std(std_stream)?;
+                stream.writable().await?;
+                stream
```
xv-thomas-leong, line 61: Instead of manually creating a socket, a simpler way would be to create the socket with tokio::TcpSocket

```rs
                let socket = if remote_addr.is_ipv6() {
                    TcpSocket::new_v6()?
                } else {
                    TcpSocket::new_v4()?
                };

                // Set SO_MARK before connect() so that the SYN packet is also
                // marked, ensuring all traffic (including connection setup) is
                // subject to policy routing rules based on the mark.
                #[cfg(all(linux, not(feature = "mobile")))]
                if fwmark != 0 {
                    let socket = socket2::SockRef::from(&socket);
                    match socket.set_mark(fwmark) {
                        Ok(_) => tracing::info!("Applied firewall mark to outside TCP socket"),
                        Err(e) => tracing::warn!("Failed to set SO_MARK on TCP socket: {}", e),
                    }
                }

                socket.connect(remote_addr).await?
```

PR #499, opened 2026-08-05
ci: reduce time on nix build

kp-antonio-yang:
## Description

Reduce CI wall-clock time for nix-build jobs by combining the client and server builds into a single `nix build` invocation per target.

Previously, each matrix target ran two separate `nix build` steps — one for `lightway-client` and one for `lightway-server` — along with an additional MSRV pair, totalling four build steps on Linux. These are now collapsed into one step per target using `nix build -j 2 --cores 2 .#lightway-client-<nix-target> .#lightway-server-<nix-target>`, which lets Nix parallelise shared dependency builds and avoids evaluating the flake twice.

The MSRV matrix row is also split out as a separate `x86_64-linux-gnu` entry with `build_msrv: true` instead of being an additional flag on the same row, keeping the conditional logic cleaner.

## Motivation and Context

Each nix-build CI run was burning two separate Nix evaluations and two separate daemon cold-starts per target. Combining the builds into a single invocation halves that overhead and makes better use of parallel CPU cores on the runner.

## How Has This Been Tested?

- [ ] CI pipeline observed to produce passing `nix-build` jobs with combined client+server steps
- [ ] Artifact upload paths (`result/bin/lightway-client`, `result/bin/lightway-server`) verified against the new default result symlink
- [ ] MSRV build confirmed to still run on `x86_64-unknown-linux-gnu` only

## Types of changes

- [x] CI / infrastructure change (no production code affected)

## Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -266,21 +266,18 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - target: x86_64-unknown-linux-gnu
+          - target: x86_64-linux-gnu
             os: ubuntu-latest
-            package_client: lightway-client
-            package_server: lightway-server
+            build_msrv: false
+          - target: x86_64-linux-gnu
+            os: ubuntu-latest
             build_msrv: true
-          - target: aarch64-apple-darwin
+          - target: aarch64-darwin
             os: macos-26
-            package_client: lightway-client
-            package_server: lightway-server
             build_msrv: false
           # Cross-compile x86_64 Darwin from Apple Silicon (much faster than Intel Mac)
-          - target: x86_64-apple-darwin
+          - target: x86_64-darwin
             os: macos-26
-            package_client: lightway-client-x86_64-darwin
-            package_server: lightway-server-x86_64-darwin
             build_msrv: false
     steps:
       - uses: actions/checkout@v4.3.1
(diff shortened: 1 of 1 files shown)


kp-antonio-yang:
Some CI timing comparisons for full builds (when cache is at its limit or missing): build time reduces from `42m6s` to `30m28s`. Since the client and server share similar dependencies, overall cache downloads are reduced, which also improves cache hit rate across CI tasks.

Screenshot-2026-08-06 16-10-24
Screenshot-2026-08-06 16-10-51

.github/workflows/ci.yaml
```diff
@@ -267,20 +267,21 @@ jobs:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
+            nix_target: x86_64-linux-gnu
```
kp-mariappan-ramasamy: I think we don't need nix_target and can reuse target. i don't see target is anywhere else.

kp-antonio-yang: Thank you.

kp-antonio-yang: The `target` is used for the upload name
```
        uses: actions/upload-artifact@v4.6.2
        with:
          name: lightway-client-${{ matrix.target }}
          path: result/bin/lightway-client
```
so it will change from `lightway-client-x86_64-unknown-linux-gnu` to `lightway-client-x86_64-linux-gnu`.
I think we still can do that, the name is still clear.

kp-mariappan-ramasamy (approved):

PR #500, opened 2026-08-06
Split Expresslane into separate crate

kp-mariappan-ramasamy:
## Description

With expresslane, data processing can be offloaded to process external to lightway. This in turn makes lightway process simpler and handle only control traffic.

This PR attempts to split the expresslane data into a separate crate. This makes it easier for any external offload application to use this directly.

## Motivation and Context

Make expresslane more accessible from downstream consumers.

## How Has This Been Tested?

Lightway core still uses the same expresslane primitives, so it exercises the split.

## Types of changes

## Checklist:


lightway-core/src/connection.rs
kp-thomas-yau: I found out that we're mapping any error we received in the socket to `AccessDenied` while debugging something locally. Is this done on purpose?

kp-mariappan-ramasamy: No, it is a typo. I think added as a a placeholder and then forgot to fix it.
I will update it

kp-mariappan-ramasamy: Fixed it by adding new error variant. PTAL

xv-thomas-leong (approved):
LGTM

PR #510, opened 2026-08-11
CVPN-2682: swallow EPIPE on connected UDP send via dead-socket hook

kp-thomas-yau:
## Description

Adds an opt-in dead-socket callback to the client's outside UDP socket, registered via `Udp::on_dead_socket`. Gated to iOS/tvOS (where the network extension hits this), plus Apple test builds so the unit tests run on macOS. When a callback is registered and a connected send fails with `EPIPE`, the error is swallowed like the other transition errors (fake success; DTLS retransmission covers the packet) and the callback fires once per socket instance so the embedding application can arrange a replacement socket. With no callback registered, `EPIPE` stays on the fatal path — behaviour is unchanged for existing users.

## Motivation and Context

On iOS a network transition (e.g. cellular → WiFi) can defunct a `connect()`-ed UDP socket's flow, after which every send on the fd fails with `EPIPE` permanently. That error currently reaches the fatal error path and tears the session down. The fd is unrecoverable — `send_to` fails identically, so downgrading doesn't help — and only the embedding application can provide a replacement socket, hence the hook. Keepalive timeout remains the liveness backstop.

## How Has This Been Tested?

New unit tests in `udp.rs` against real sockets, using `shutdown(SHUT_WR)` to make Darwin return `EPIPE` from send: with a hook registered, sends report success and the callback fires exactly once across repeated failures; without one, send returns a fatal `BrokenPipe` error. `cargo test -p lightway-client` (75 passed) and clippy are clean on macOS. Also verified the hook compiles out of the macOS production build (`cargo check`) and into the iOS one (`cargo check --lib --target aarch64-apple-ios`).

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/io/outside/udp.rs
+++ b/lightway-client/src/io/outside/udp.rs
@@ -3,6 +3,8 @@ use anyhow::{Result, anyhow};
 use async_trait::async_trait;
 use lightway_app_utils::sockopt;
 use lightway_core::{IOCallbackResult, OutsideIOSendCallback, OutsideIOSendCallbackArg};
+#[cfg(any(ios, tvos, all(test, apple)))]
+use std::sync::OnceLock;
 #[cfg(apple)]
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::{AtomicU64, Ordering};
@@ -30,6 +32,15 @@ pub struct Udp {
     /// [`OutsideIO::reconnect`] on the next network-change event.
     #[cfg(apple)]
     connected: AtomicBool,
+    /// Callback registered via [`Udp::on_dead_socket`], fired when the
+    /// connected socket turns out to be permanently dead. iOS/tvOS only
+    /// (compiled into Apple test builds so the unit tests can run on macOS).
+    #[cfg(any(ios, tvos, all(test, apple)))]
+    dead_socket_cb: OnceLock<Arc<dyn Fn() + Send + Sync>>,
+    /// Latch ensuring the dead-socket callback fires at most once per socket
+    /// instance.
+    #[cfg(any(ios, tvos, all(test, apple)))]
+    dead: AtomicBool,
     /// Consecutive swallowed send errors, reset by a successful send; see
     /// [`Udp::note_swallowed_send`].
     swallowed_sends: AtomicU64,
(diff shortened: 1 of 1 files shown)


lightway-client/src/io/outside/udp.rs
```diff
@@ -119,6 +134,22 @@ impl Udp {
...
+    /// swallowing those sends — without one they stay on the fatal path,
+    /// which is the enforcement that a recovery handler exists.
+    ///
+    /// The callback runs inside the send callback, under the connection
+    /// lock: it must be non-blocking and must not re-enter the connection.
+    /// It is invoked at most once per socket instance and is expected to
+    /// arrange, elsewhere, for this socket to be replaced.
+    #[cfg(any(ios, tvos, all(test, apple)))]
+    pub fn on_dead_socket(&self, cb: Arc<dyn Fn() + Send + Sync>) {
+        let _ = self.dead_socket_cb.set(cb);
```
kp-ben-hung: Maybe we can do a `expect()` here so a second call to this function will fail closed? (assuming a second call to this function is not expected after the `cb` is set) Or at least return the Err(...)) to the caller? As this is a `OnceLock`.

lightway-client/src/io/outside/udp.rs
```diff
@@ -416,6 +447,28 @@ impl OutsideIOSendCallback for Udp {
...
+                // iOS defuncts the connected flow on an interface change
+                // (e.g. cellular -> WiFi), after which every send on this fd
+                // fails with EPIPE: the fd is unrecoverable, and `send_to`
+                // fails identically so downgrading would not help. Fire the
+                // registered callback (once) to request a replacement socket
+                // and swallow the error — DTLS retransmission covers the
+                // packet and the keepalive timeout remains the liveness
+                // backstop.
+                if !self.dead.swap(true, Ordering::Relaxed)
+                    && let Some(cb) = self.dead_socket_cb.get()
```
kp-ben-hung, line 467: nits: i think the `is_some()` and  `let Some(cb)` could be collapsed. (the inner check is not needed)

kp-thomas-yau: if-let guard in match statement is not available on our MSRV, so back to original behaviour for now....

kp-ben-hung (approved):
LGTM

PR #508, opened 2026-08-07
Ci improvements

kp-mariappan-ramasamy:
## Description

Merge both client and server compilation to save CI time
Also disabled nix build of x64 mac which takes too long

## Motivation and Context

To improve CI times

## How Has This Been Tested?

CI passed

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] The correct base branch is being used, if not `main`

PR #511, opened 2026-08-11
client: move `#[cfg(feature="mobile")]` to `#[cfg(mobile)]`

kp-antonio-yang:
## Description

This PR bundles related improvements to `lightway-client`.   We need #508 to let the CI all green to prevent out of disk space issue, so this PR based on top of it.

#### Replace `mobile` Cargo feature with built-in `mobile` cfg predicate

Replace the hand-rolled `mobile` Cargo feature flag with Rust's built-in `mobile` cfg predicate across `lightway-client`.

Key mechanical changes:
- **`Cargo.toml`**: Remove the `mobile` feature entirely. `mobile-test` is narrowed to only enabling `dep:uniffi` for integration-test scaffolding. `tracing-core` and `tracing-panic` become unconditional dependencies (lightweight, needed on all targets). `uniffi` moves to a platform-specific `[target.'cfg(any(target_os = "android", target_os = "ios", target_os = "tvos"))'.dependencies]` section.
- **`src/lib.rs`, `src/config.rs`**: All `#[cfg(feature = "mobile")]` guards replaced with `#[cfg(mobile)]` or `#[cfg(any(mobile, feature = "mobile-test"))]` as appropriate.
- **`src/io/outside/tcp.rs`, `udp.rs`**: fwmark exclusion guard changed from `not(feature = "mobile")` to `not(feature = "mobile-test")` to keep firewall marking active on real mobile targets.
- **`ci.yaml`**: `--features=mobile` removed from check/clippy/JsonSchema steps; replaced by `--lib` where target-based detection is sufficient.
- **`Makefile.toml`**: `--features mobile` dropped from the Android NDK build task.

## Motivation and Context

- The `mobile` feature flag was fragile: forgetting to pass it produced a silently incorrect binary (no UniFFI scaffolding, wrong cfg branches). Using the compiler's native `mobile` predicate means the correct code paths are selected automatically based on the actual target triple, with no manual flag required.

## How Has This Been Tested?
- [X] CI pipeline green on the updated `ci.yaml` steps

## Types of changes

- [x] Refactor (internal restructuring, no behaviour change for end users)

## Checklist:

### DIFF
--- a/lightway-client/Cargo.toml
+++ b/lightway-client/Cargo.toml
@@ -17,8 +17,7 @@ crate-type = ["cdylib", "rlib"]
 
 [features]
 default = ["wolfssl"]
-mobile = ["uniffi", "tracing-core", "tracing-panic"]
-mobile-test = [ "mobile" ]  # Only mock struct, no uniffi structs
+mobile-test = ["dep:uniffi", "dep:tracing-core", "dep:tracing-panic"]  # Only mock struct, no uniffi exports
 boringssl = ["lightway-app-utils/boringssl"]
 wolfssl = ["lightway-app-utils/wolfssl"]
 debug = ["lightway-core/debug","lightway-app-utils/debug"]
@@ -54,10 +53,9 @@ tokio.workspace = true
 tracing = { workspace = true, features = ["attributes"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
-# mobile feature
+tracing-core = { version = "0.1.33", optional = true }
+tracing-panic = { version = "0.1.2", optional = true }
 uniffi = { workspace = true, optional = true }
-tracing-core = { version = "0.1.33", optional =true }
-tracing-panic = { version = "0.1.2", optional= true }
 
 [dev-dependencies]
 more-asserts.workspace = true
@@ -69,6 +67,11 @@ serial_test.workspace = true
 tempfile = "3.24.0"
 tracing-test = "0.2.6"
 tun-rs.workspace = true
+
+[target.'cfg(any(target_os = "android", target_os = "ios", target_os = "tvos"))'.dependencies]
+tracing-core = "0.1.33"
+tracing-panic = "0.1.2"
+uniffi = { workspace = true }
 
 [target.'cfg(target_os = "android")'.dependencies]
 tracing-android = "0.2.0"
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -24,7 +24,6 @@ const DEFAULT_RCVBUF: ByteSize = ByteSize::mib(8);
 // values follow the Rust convention in Default trait, such that we are able
 // to serialized out any kind of configure from the Config::default(), such
 // that our library will be nice and easy to further do integrations.
-#[allow(dead_code)]
 #[derive(Clone, Debug, PartialEq, Deserialize, JsonSchema, Serialize, Patch, Substrate)]
 #[patch(attribute(derive(Clone, Deserialize, Parser)))]
 #[patch(attribute(command(about = "A lightway client")))]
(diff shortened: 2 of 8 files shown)


lightway-client/src/config.rs
```diff
@@ -341,7 +342,6 @@ pub struct Config {
     pub enable_dpapi: bool,
 
     /// SNI header for TLS connections
-    #[cfg(feature = "mobile")]
     #[patch(attribute(clap(skip)))]
     #[schemars(extend("x-cfg" = "mobile"))]
     pub sni_header: String,
```
kp-thomas-yau: why do this setting have to exist on desktop? I think desktop have server_dn already?

lightway-client/Cargo.toml
```diff
@@ -55,10 +54,9 @@ tokio.workspace = true
 tracing = { workspace = true, features = ["attributes"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
-# mobile feature
+tracing-core = "0.1.33"
+tracing-panic = "0.1.2"
 uniffi = { workspace = true, optional = true }
```
kp-thomas-yau: why not move them under
`[target.'cfg(any(target_os = "android", target_os = "ios", target_os = "tvos"))'.dependencies]`?

kp-thomas-yau (commented):
Btw, the commit message in a66e06a4500ab70022689bb736d7448a2d8dd057 is wrong. We are using cfg_alias to define custom `mobile` cfg flag. It is not built in Rust at all..... Please update the commit message.

and also I think this PR's scope is broader than it should be, not sure why there's 2 nix-related changes here.

lightway-client/src/config.rs
kp-antonio-yang: JsonSchema is built from linux in our CI, and also possible built from other desktop.  If the mobile is not the feature gate anymore.  The config will have all the mobile fields, such that developer can generate the schema by `cargo run -p lightway-client -- -c client.schema.json -g jsonschema`

Because it does not make sense to build a cli on `android` to generate a JsonSchema for android, it is a trade-off for extra field in the Config.

kp-thomas-yau: Then we need a way to make sure `sni_header` is never used on desktop. It's simply a big risk factor IMO since we now have 2 places to set SNI header... and make things less confusing as well

kp-antonio-yang: I see, let me merge the field, it should be one.  Thanks for reminding me.

kp-antonio-yang: Regarding the sni_header in desktop, it is possible to setup globally and apply to all the servers when the sni_header of each server is empty, and the follow the current flow handling the global auth for all servers.

kp-thomas-yau: For reference: I originally thought sni_header + server_dn are the same thing, which is entirely different, so this should not be an issue.

lightway-client/src/config.rs
```diff
@@ -26,6 +25,7 @@ const DEFAULT_RCVBUF: ByteSize = ByteSize::mib(8);
 // values follow the Rust convention in Default trait, such that we are able
 // to serialized out any kind of configure from the Config::default(), such
 // that our library will be nice and easy to further do integrations.
+#[allow(dead_code)]
```
kp-thomas-yau: Why do we allow a dead code here?

lightway-client/src/config.rs
```diff
@@ -341,9 +340,9 @@ pub struct Config {
     #[schemars(extend("x-cfg" = "windows"))]
     pub enable_dpapi: bool,
 
-    /// SNI header for TLS connections
-    #[patch(attribute(clap(skip)))]
-    #[schemars(extend("x-cfg" = "mobile"))]
+    #[patch(attribute(clap(long)))]
+    #[patch(attribute(doc = r#"SNI header for TLS connections"#))]
+    /// ex: example.com
```
kp-thomas-yau, line 351: We should put a comment here to explicitly say it will be overrode by server-level config. and should we make this as option<string>?

kp-antonio-yang: I change the behavior, so it is the same behavior as `server_dn` only for single server mode.   Now the behaviors of two field about server domain name are the same.  

Thanks for pointing out,  I did overwrite, but I should not let `server_dn` override other servers config.  It is better to be the same as `server_dn`.

kp-thomas-yau (approved):

kp-antonio-yang:
Thanks for review.

PR #512, opened 2026-08-11
server: client: config: handle unknown fields from config file for client and server

kp-antonio-yang:
## Description

Detect unknown fields in config files for both `lightway-client` and `lightway-server`, and fail at validation time with a clear error message. An `--accept-unknowns` flag is added to both binaries to downgrade the error to a logged warning, allowing the process to continue.

### Change details

Previously, deserialising a config file with a typo or a removed field would succeed silently, making it easy to unknowingly run with an ineffective setting. This PR surfaces those mistakes at startup.

Key mechanical changes:

- **Unknown-field collection**: `#[patch(attribute(serde(deny_unknown_fields)))]` is replaced with a `#[patch(attribute(serde(flatten)))]` `HashMap<String, serde_json::Value>` field (`unknowns`) that absorbs any key not matched by a known field. This shifts detection from parse time to validation time, which is needed because `struct-patch`'s `deny_unknown_fields` and `flatten` cannot coexist.
- **`Config::validate()`** (client and server): if `unknowns` is non-empty, either bail with an error listing the unknown keys (default) or emit a `tracing::error!` / `tracing::warn!` and continue when `--accept-unknowns` is set.
- **`--accept-unknowns` / `-a` flag**: added to both `ConfigPatch` structs via `#[patch(attribute(clap(short, long)))]`. Lets operators migrate config files incrementally without blocking startup.
- **`struct-patch` bump 0.12.5 → 0.13.2**: required for `skip_wrap` + `apply_by` support used by the `unknowns` field merge strategy.
- **Test fixture**: removes the stale `bind_attempts` field from `tests/server/server_config.yaml`, which would now trigger the new error.
- **Unit tests** (client and server):
  - `config_unknown_fields_cause_error` – parses a YAML with an unknown key, applies the patch, and asserts `validate()` returns an error.
  - `config_unknown_fields_with_accept_flag_only_warns` – same setup but with `accept_unknowns = true`; asserts `validate()` succeeds and the expected log message is emitted (via `tracing-test`).

## Motivation and Context

Config files with typos or stale keys (e.g. from a field that was renamed or removed) were silently accepted, giving users no feedback that their setting was being ignored. Rejecting unknown fields at validation time surfaces these mistakes immediately at startup, while the `--accept-unknowns` escape hatch avoids hard-blocking operators who need time to clean up their config files.

## How Has This Been Tested?
- [x] add new testcase with unknow fields

## Types of changes

- [ ] Refactor (internal restructuring, no behaviour change for end users)
- [ ] CI / infrastructure change (no production code affected)

## Checklist:

### DIFF
--- a/lightway-client/src/config.rs
+++ b/lightway-client/src/config.rs
@@ -12,6 +12,7 @@ use lightway_app_utils::args::{
 use lightway_core::{AuthMethod, MAX_OUTSIDE_MTU, Version};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::time::Duration as StdDuration;
 use std::{net::Ipv4Addr, path::PathBuf};
 use struct_patch::{Patch, Substrate};
@@ -345,6 +346,18 @@ pub struct Config {
     #[patch(attribute(clap(skip)))]
     #[schemars(extend("x-cfg" = "mobile"))]
     pub sni_header: String,
+
+    #[patch(attribute(clap(short, long)))]
+    #[patch(empty_value = false)]
+    #[patch(attribute(serde(default)))]
+    #[patch(attribute(doc = r#"Accept unknown inputs with error message without quiting"#))]
+    pub accept_unknowns: bool,
+
+    /// The unknown options from config file
+    #[patch(attribute(clap(skip)))]
+    #[patch(attribute(serde(flatten)))]
+    #[patch(skip_wrap, apply_by(std::collections::HashMap::extend))]
+    pub unknowns: HashMap<String, serde_json::Value>,
 }
 
 impl Config {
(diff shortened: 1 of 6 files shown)


kp-antonio-yang:
I think there is some benefit to accept unknown field when developing new features.  Let me add an option to accept unknown fields and only show warning.

lightway-client/src/config.rs
```diff
@@ -416,6 +429,15 @@ impl Config {
 
     /// Ensure the config is validated, and alerted when there's a conflict in the settings.
     pub fn validate(&self) -> anyhow::Result<()> {
+        if !self.unknowns.is_empty() {
+            let fields: Vec<&str> = self.unknowns.keys().map(String::as_str).collect();
+            if self.accept_unknowns {
+                tracing::error!(fields = ?fields, "unknown config fields will be ignored");
```
kp-mariappan-ramasamy: Let's make this as a warning, instead of error.

kp-mariappan-ramasamy (commented):
LGTM. I think 3 commits can be squashed, since there is no additional benefit splitting it now.

kp-mariappan-ramasamy (approved):

PR #513, opened 2026-08-11
tests: replace checked-in certs with on-the-fly generated certs

kp-ben-hung:
## Description

Replace the checked-in certs with a module at `lightway-core/tests/common/certgen.rs`, generating certs during tests-time. The test generation logic now live outside of the time-out window to prevent flaky tests.

## Motivation and Context

Currently, certs used in tests are checked-in. Replace them with test-time generated certs. This allows better control of certs tests parameters in the future, and prevent tests failing in the future due to expiry.

## How Has This Been Tested?

Ran `cargo nextest run` and returned green. And modifying `example.com` to `example2.com` will fail the `test_server_dn` tests, confirming the tests are working against the generated certs.

## Types of changes

## Checklist:


lightway-core/tests/connection.rs
```diff
@@ -382,6 +385,10 @@ async fn test_server_dn(server_dn: Option<&str>) {
     let _ = client_sock.writable().await;
 
     let auth = Arc::new(TestAuth::default());
+
+    // Generate the shared test PKI before the timed window (see run_test).
+    test_pki();
```
kp-antonio-yang: Could we rename this function? 

The current name reads like the function itself is running a test, which may confuse a first-time reader?  
It actually just builds the shared PKI used by tests. 

Something like `gen_shared_testing_pki()` would make the intent clear, and then the extra doc line probably isn't needed anymore.

kp-ben-hung: Agreed, renamed the function. The doc line is still needed i think because it's highlighting the cert-gen process lives *outside* the `tokio::time::timeout` window. If it's included it will cause flaky tests (the cert-gen process takes time as well).

kp-antonio-yang (approved):

PR #515, opened 2026-08-13
boring: add chacha20 poly1305

kp-ben-hung:
## Description

Add chacha20 poly1305 support on BoringSSL backend.

## How Has This Been Tested?

CI passing. The test vectors in the changes are directly from RFC 8439. (only change is the AAD being empty).
Also verified manually that the calculation is correct (using the same parameters in the test case).

## Types of changes

## Checklist:

### DIFF
--- a/lightway-boring/src/chacha20_poly1305.rs
+++ b/lightway-boring/src/chacha20_poly1305.rs
@@ -1,0 +1,218 @@
+//! ChaCha20-Poly1305 AEAD cipher (RFC 8439)
+
+use boring::aead::{AeadCtx, Algorithm};
+use bytes::BytesMut;
+use zeroize::Zeroizing;
+
+/// ChaCha20-Poly1305 authenticated encryption cipher.
+pub struct Chacha20Poly1305Aead {
+    ctx: Option<AeadCtx>,
+}
+
+impl Chacha20Poly1305Aead {
+    /// Size of key in bytes
+    pub const KEY_SIZE: usize = 32;
+
+    /// Size of initialisation vector / nonce in bytes
+    pub const IV_SIZE: usize = 12;
+
+    /// Size of authentication tag in bytes
+    pub const AUTHTAG_SIZE: usize = 16;
+
+    /// Creates a new `Chacha20Poly1305Aead` instance bound to `key`.
+    pub fn new(key: [u8; Self::KEY_SIZE]) -> Self {
+        let key = Zeroizing::new(key);
+        let ctx = AeadCtx::new_default_tag(&Algorithm::chacha20_poly1305(), key.as_ref()).ok();
+        Self { ctx }
+    }
+
+    /// Encrypt plaintext using ChaCha20-Poly1305.
+    ///
+    /// Returns the ciphertext and the 16-byte Poly1305 authentication tag.
+    ///
+    /// The nonce must be unique for every call made with the same key.
+    pub fn encrypt(
+        &self,
+        iv: [u8; Self::IV_SIZE],
+        plain_text: &[u8],
+    ) -> Result<(BytesMut, [u8; Self::AUTHTAG_SIZE]), String> {
+        let ctx = self
+            .ctx
+            .as_ref()
+            .ok_or_else(|| String::from("Encrypt failed"))?;
+
+        let mut out = BytesMut::from(plain_text);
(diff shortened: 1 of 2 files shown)


lightway-boring/src/chacha20_poly1305.rs
```diff
@@ -0,0 +1,224 @@
...
+#[derive(Error, Debug)]
+pub enum Chacha20Poly1305Error {
+    #[error("Chacha20Poly1305 Init Failed")]
+    InitFailed,
+    #[error("Auth tag mismatch")]
+    AuthTagMismatch,
+    /// BoringSSL error
+    #[error("Fatal: {0}")]
+    Fatal(#[from] ErrorStack),
+}
```
kp-antonio-yang: If I understand correctly, `lightway-boring` exposes the same API surface as `wolfssl-rs`, but the error types differ. See for comparison: 
https://github.com/expressvpn/wolfssl-rs/blob/main/wolfssl/src/chacha20_poly1305.rs

Since both crates are ours, I think the cleaner approach is to define a `TlsBackend` trait covering the shared functions and a common error type, and have `lightway-boring` and `wolfssl-rs` each implement it. That way callers depend on the trait rather than on either backend's concrete error type.

kp-antonio-yang (changes requested):

lightway-boring/src/chacha20_poly1305.rs
kp-ben-hung: First on the `TlsBackend` trait, i don't think it's worth it for now. 

First we are not expecting a runtime switch, so polymorphism is unnecessary.  The feature gate we have now already gave us compiler-checked conformance *in practice* due to the CI. And sometimes we decide that the features are really not needed (or should be excluded) in one side, and we feature-gate them.

Secondly, backend quirks are security-critical and i think unifying hides them too much. A common trait will have to pick the *weakest common contract (losing information a security-conscious caller needs)*, or *paper over soundness-relevant differences*. (for example no-ops in some of the functions). And i think feature-gating keeps the differences vocal and clear.

Thirdly, if we do a trait object, we still have to have an associated type `Error` on the trait, and each backend could implement it differently, so compiler cannot force the same API for the Error on both backend anyway (by using the trait alone). The errors will only surface on the caller side (which is basically the same now with feature-gate). I don't think it's useful.

Lastly, if we do trait, we must consume it using generics. If we use Box<dyn ...>, it will likely not compile. The trait will likely not be **dyn-compatible** since i think we need associated constants.

However, I agree that the error types should be unified, i will make it into `Result<_, String>`  (at least for now), since that's what wolfssl-side is doing.

kp-antonio-yang: One of Rust's core features is zero-cost abstraction. Traits are zero-cost when used with static dispatch (i.e. without `dyn`):

**Static dispatch**: when traits are used through generics or `impl Trait`, the compiler generates specialized code for each concrete type at compile time — a process called monomorphization. This produces direct function calls that are usually inlined(not at `opt-level=0`, of course), so there is no runtime overhead compared to hand-written code.
**No vtable overhead**: unlike dynamic dispatch, static dispatch does not use vtables or fat pointers, so there is no runtime lookup cost or indirection.

I'm fine with merging this into main quickly to get the support in, but let's not forget there is still room for improvement here. There's also a draft demo from the AI that uses static dispatch instead of `dyn`. I'm happy to take over the refactor later on.
AI Demo: https://github.com/expressvpn/lightway/pull/516

kp-antonio-yang (approved):

lightway-boring/src/chacha20_poly1305.rs
kp-ben-hung: I completely agree we can use monomorphization via `impl` or generics, and it would be zero performance cost (that's why i never included that argument in my response (and i said we must use generics). 

However, that's not what my concern now. I am more concerned about potentially hiding security details, and make the codebase even more complicated for (at least i see for now, no major benefits that couldn't be done with feature-flag), and even if we do that, there would likely be ripple effect throughout the codebase (or even between crates) that will likely need more discussion. *(Assuming you meant putting *all* backend-specific structs, enums and functions behind a trait, not just the chacha20 part).*

I think we can discuss this more in the future when we have time, I am also interested in deep-diving more on this too, glad to have this discussion open in the future.

PR #519, opened 2026-08-17
client: Re-pin server route on every network change

kp-antonio-yang:
## Description

This PR improves the server route management and re-pinning logic in the lightway client. The changes refactor the route update process, consolidate duplicate retry logic, and implement more robust re-pinning behavior on network changes.
Ticket: CVPN-2731

### Key Changes:

1. **Route Update Consolidation**: Extract duplicate route deletion and addition logic into a `update_server_route()` helper method to reduce code duplication and prevent potential inconsistency bugs.

2. **Unified Retry Logic**: Previously, exponential back-off retry logic was only applied under `RepinMode::Always` on Apple targets, while `RepinMode::OnRouteChange` used simpler 1s fixed retry with warning-level logs. Both modes now share consistent retry behavior with log-level escalation.

3. **In-Loop Retry for Re-pins**: On Apple platforms, re-pin the server route unconditionally on every network event. Failed re-pins are retried from inside the `network_event_coordinator` select! loop with exponential back-off (500ms → 1s → 2s → 4s → 8s → 10s), allowing the loop to continue servicing other events while retry deadlines are pending.

4. **RepinMode Abstraction**: Introduce `RepinMode` and `RepinState` to `route_manager/repin.rs` with a single variant (`OnRouteChange`) that re-pins only when gateway or interface index actually changed. Wire `repin_mode` into `RouteManagerInner` and delegate the update decision to `repin_mode.needs_repin()`.

5. **MSRV Bump**: Bump Minimum Supported Rust Version to 1.95.0 for `cfg_select!` support.

## Motivation and Context

These changes improve resilience to within-subnet roaming scenarios on macOS where routing identifiers are identical but the physical network path has changed. By reliably re-pinning the outside socket on every network transition and properly retrying failed attempts, the client can handle seamless roaming without connection drops.

## How Has This Been Tested?
- Follow current testcase

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ repository = "https://github.com/expressvpn/lightway"
 edition = "2024"
 authors = ["lightway-developers@expressvpn.com"]
 license = "AGPL-3.0-only"
-rust-version = "1.91.0"
+rust-version = "1.95.0"
 
 [workspace.lints.rust]
 unsafe_code = "deny"
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -856,7 +856,8 @@ async fn handle_network_change<ExtAppState: Send + Sync>(
 /// nudge the connection-level network-change handler. On Apple platforms the
 /// nudge also fires when the server route was actually replaced (Datagram
 /// wiring only). Failed route refreshes retry from inside the select! loop so
-/// fresh wake-ups are never blocked; a fresh event restarts the retry budget.
+/// fresh wake-ups are never blocked, and a fresh event inherits the retry it
+/// supersedes, so consecutive failures keep backing off rather than resetting.
 /// Owns the [`RouteUpdater`], so aborting the task removes the installed routes.
 #[cfg(desktop)]
 async fn network_event_coordinator(
@@ -914,12 +915,11 @@ async fn network_event_coordinator(
                 if route_repin_state.is_some() => None,
         };
 
-        // A fresh event restarts the retry clock but keeps any nudge owed
-        // from the interrupted retry.
+        // A fresh event serves immediately but inherits the superseded retry's
+        // owed nudge, failure clock and back-off position, so a re-pin that
+        // keeps failing keeps escalating across events.
         let mut state = match event {
-            Some(nudge) => {
-                RepinState::new(nudge || route_repin_state.take().is_some_and(|p| p.nudge))
-            }
+            Some(nudge) => RepinState::for_event(nudge, route_repin_state.take()),
             None => route_repin_state.take().expect("branch guarded on is_some"),
         };
 
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -925,10 +925,20 @@ async fn network_event_coordinator(
 
         match route_updater.check_and_update_server_route().await {
             Err(e) => {
-                route_updater.on_repin_failure(&mut state, &e);
-                // Reconnect and nudge wait for the terminal outcome.
-                route_repin_state = Some(state);
-                continue;
+                match &e {
+                    route_manager::RoutingTableError::ServerRouteAddFailed(_) => {
+                        tracing::error!(
+                            "Fatal error updating server route: {e}. Shutting down VPN."
+                        );
+                        break;
+                    }
+                    _ => {
+                        route_updater.on_repin_failure(&mut state, &e);
+                        // Reconnect and nudge wait for the terminal outcome.
+                        route_repin_state = Some(state);
+                        continue;
+                    }
+                }
             }
             // A replaced server route is ground truth that the path to the
             // server moved; probe it so the session floats promptly.
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -42,7 +42,7 @@ use crate::debug::WiresharkKeyLogger;
 use crate::dns_manager::{DnsConfigMode, DnsManager, DnsManagerError, DnsSetup};
 use crate::keepalive::Config as KeepaliveConfig;
 #[cfg(desktop)]
-use crate::route_manager::{RouteManager, RouteMode, RouteUpdater};
+use crate::route_manager::{RepinState, RouteManager, RouteMode, RouteUpdater};
 #[cfg(batch_receive)]
 use lightway_core::MAX_IO_BATCH_SIZE;
 pub use lightway_core::{
@@ -855,8 +855,9 @@ async fn handle_network_change<ExtAppState: Send + Sync>(
 /// socket (Apple platforms) and, when the wake-up represents a network transition,
 /// nudge the connection-level network-change handler. On Apple platforms the
 /// nudge also fires when the server route was actually replaced (Datagram
-/// wiring only). Owns the [`RouteUpdater`], so aborting the task removes the
-/// installed routes.
+/// wiring only). Failed route refreshes retry from inside the select! loop so
+/// fresh wake-ups are never blocked; a fresh event restarts the retry budget.
+/// Owns the [`RouteUpdater`], so aborting the task removes the installed routes.
 #[cfg(desktop)]
 async fn network_event_coordinator(
     mut route_updater: RouteUpdater,
@@ -868,15 +869,17 @@ async fn network_event_coordinator(
     network_change_signal: mpsc::Sender<()>,
 ) {
     tracing::info!("Reacting to network change events...");
-    loop {
-        let mut nudge = nudge_on_route_event;
+    let mut route_repin_state: Option<RepinState> = None;
 
-        tokio::select! {
+    loop {
+        // `Some(nudge)` is a fresh network event; `None` is a retry tick.
+        let event = tokio::select! {
             changed = route_rx.changed() => {
                 if changed.is_err() {
                     // Wake source gone; dropping the updater clears routes.
                     break;
                 }
+                let mut nudge = nudge_on_route_event;
                 // Fold a transition that fired alongside the route event into
                 // this iteration instead of waking a second time.
                 if let Some(rx) = transition_rx.as_mut()
@@ -885,6 +888,7 @@ async fn network_event_coordinator(
                     rx.mark_unchanged();
                     nudge = true;
                 }
+                Some(nudge)
             }
             changed = async {
                 match transition_rx.as_mut() {
@@ -897,22 +901,40 @@ async fn network_event_coordinator(
                     transition_rx = None;
                     continue;
                 }
-                nudge = true;
                 // A transition can complete on the address leg alone; consume
                 // any simultaneous route event so the pair is one iteration.
                 if route_rx.has_changed().unwrap_or(false) {
                     route_rx.mark_unchanged();
                 }
+                Some(true)
             }
-        }
+            // The deadline is absolute, so recreating the sleep future each
+            // iteration is fine.
+            _ = async { tokio::time::sleep_until(route_repin_state.as_ref().unwrap().next_at).await },
+                if route_repin_state.is_some() => None,
+        };
+
+        // A fresh event restarts the retry clock but keeps any nudge owed
+        // from the interrupted retry.
+        let mut state = match event {
+            Some(nudge) => {
+                RepinState::new(nudge || route_repin_state.take().is_some_and(|p| p.nudge))
+            }
+            None => route_repin_state.take().expect("branch guarded on is_some"),
+        };
 
         match route_updater.check_and_update_server_route().await {
+            Err(e) => {
+                route_updater.on_repin_failure(&mut state, &e);
+                // Reconnect and nudge wait for the terminal outcome.
+                route_repin_state = Some(state);
+                continue;
+            }
             // A replaced server route is ground truth that the path to the
             // server moved; probe it so the session floats promptly.
             #[cfg(apple)]
-            Ok(true) if nudge_on_route_update => nudge = true,
+            Ok(true) if nudge_on_route_update => state.nudge = true,
             Ok(_) => {}
-            Err(e) => tracing::warn!("Updating server route failed: {:?}", e),
         }
 
         // The connected outside socket pins the route resolved at connect()
@@ -922,7 +944,9 @@ async fn network_event_coordinator(
             io.reconnect();
         }
 
-        if nudge && let Err(e) = network_change_signal.send(()).await {
+        if state.nudge
+            && let Err(e) = network_change_signal.send(()).await
+        {
             tracing::error!("Failed to send network_change_signal: {e}");
         }
     }
--- a/lightway-client/src/route_manager.rs
+++ b/lightway-client/src/route_manager.rs
@@ -200,7 +200,7 @@ impl RouteUpdater {
     /// Delegate a failed re-pin to the mode handler, which updates the retry
     /// deadline on `state` and logs at the appropriate level.
     pub fn on_repin_failure(&self, state: &mut RepinState, error: &RoutingTableError) {
-        self.inner.repin_mode.on_failure(state, error);
+        RepinMode::on_failure(state, error);
     }
 }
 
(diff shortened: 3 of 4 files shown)


lightway-client/src/route_manager.rs
```diff
@@ -1079,9 +1116,12 @@ mod tests {
         assert!(inner.server_route.is_some());
         let initial_server_route = inner.server_route.as_ref().unwrap().clone();
 
-        // Test check_and_update_server_route when no change is needed
+        // Test check_and_update_server_route when no change is needed but repin in MacOS
         let result = inner.check_and_update_server_route().await;
+        #[cfg(not(macos))]
```
kp-weiwen-goh: Let's use `cfg_select` here:
```
cfg_select! {
    macos => { ... }
    _ => { ... }
}
```

lightway-client/src/route_manager.rs
```diff
@@ -1202,9 +1242,12 @@ mod tests {
         assert_eq!(initial_server_route.destination(), EXTERNAL_IP_V6);
         assert_eq!(initial_server_route.prefix(), Ipv6Addr::BITS as u8);
 
-        // Test check_and_update_server_route when no change is needed
+        // Test check_and_update_server_route when no change is needed but repin in MacOS
         let result = inner.check_and_update_server_route().await;
+        #[cfg(not(macos))]
```
kp-weiwen-goh: Same here

lightway-client/src/route_manager.rs
```diff
@@ -511,14 +530,32 @@ impl RouteManagerInner {
...
-                );
+            let route_changed =
+                server_gateway != current_gateway || server_if_index != current_if_index;
+
+            // When roaming within the same subnet/gateway/if_index identical,
+            // we need re-pin unconditionally to rebind.
+            #[cfg(apple)]
+            let needs_update = true;
+            #[cfg(not(apple))]
+            let needs_update = route_changed;
```
kp-weiwen-goh: Same here,
```
let needs_update = cfg_select! { ... };
```

kp-weiwen-goh (commented):
LGTM, some nits

kp-weiwen-goh (dismissed):

Cargo.toml
```diff
@@ -24,7 +24,7 @@ repository = "https://github.com/expressvpn/lightway"
 edition = "2024"
 authors = ["lightway-developers@expressvpn.com"]
 license = "AGPL-3.0-only"
-rust-version = "1.91.0"
+rust-version = "1.95.0"  # for cfg_select!
```
kp-mariappan-ramasamy: Its not very useful to add cfg_select as comment here. Just mention in commit message. Should be enough

lightway-client/src/route_manager.rs
```diff
@@ -192,6 +192,25 @@ impl RouteUpdater {
         }
         self.inner.check_and_update_server_route().await
     }
+
+    /// Re-pin the server route, retrying while the network is still settling.
+    ///
+    /// Sleeps 500ms after the first failure, then 1s per retry (29.5s total).
+    /// The ceiling covers PEAP phase 2 stalls which can hold for up to 25s.
+    pub async fn update_server_route_with_retry(&mut self) -> Result<bool, RoutingTableError> {
```
kp-mariappan-ramasamy: Looking at the logic before approving once, this seems little tricky.
This async fn might take 30s to complete in case of failure, which is a significant amount of time. In that time interval, it is possible there are some route message received and missed.


I think it would be cleaner to add this loop inside `network_event_coordinator::loop::switch` ?
WDYT?

kp-antonio-yang: I have reproduced this scenario many times and it consistently completes in under `6s`(~5.3s). The `30s` figure came from a surveyed case on a Linux laptop. I will adjust the log threshold to warn when it exceeds `5.5s`, so users have visibility into what happened if the worst case occurs in a rare corner case.

Regarding `select!`, it is designed for waiting, not processing. When the route is broken, there is nothing to do except retry, so this is the correct approach here.

Is this make sense to you?

kp-antonio-yang: Okay, now the `update_server_route_with_retry` is inside the `network_event_coordinator`

lightway-client/src/lib.rs
```diff
@@ -816,6 +813,26 @@ async fn network_event_coordinator(
                     rx.mark_unchanged();
                     nudge = true;
                 }
+
+                match route_updater.update_server_route_with_retry().await {
```
kp-mariappan-ramasamy: No, I was not suggesting to inline and duplicate.

I was suggesting something like this:
https://github.com/expressvpn/lightway/commit/8790fc34ab5fa24d68fe6852a58be61c1be014be

kp-antonio-yang: Btw, I just change the log from warn to error, because it fails after `30s`, I test again on Mac also.  I believe this is good to go.

kp-mariappan-ramasamy: This keeps me thinking whether we want to close the connection if server route is not updated even after 30s and failed.

kp-antonio-yang: I disagree here. There is a valid use case for this, and error logs are not inherently bad. Clear log status gives the user the information they need to decide whether to kill the application themselves, which is better than silently killing it on their behalf.

Even sitting at an office desk, I have observed roaming times of around 30s. Why not accommodate the following common cases?

To give a concrete example: in many homes, especially multi-floor houses without mesh networking, people set up two routers, one on the ground floor and one in the master bedroom on the top floor. If a user picks up their laptop and walks from the living room to the bedroom upstairs, that journey could easily take more than 30 seconds. The same applies to moving between buildings on a campus, or from office to office in a manufacturing facility. In these cases, a longer transition time is entirely normal and expected user behavior.

kp-antonio-yang: Do we remove the MAX_RETRY, and let it keep try in `10s` interval, when the time over `30s`?

kp-mariappan-ramasamy: Yes, can

lightway-client/src/lib.rs
```diff
@@ -838,16 +797,32 @@ async fn network_event_coordinator(
...
+        next_at: tokio::time::Instant,
+        nudge: bool,
+    }
+
+    // Failed re-pins retry indefinitely: 1s interval for the first 30s,
+    // then 10s interval. Log level escalates to error after 6s.
+    const REPIN_SHORT_INTERVAL: std::time::Duration = std::time::Duration::from_secs(1);
+    const REPIN_LONG_INTERVAL: std::time::Duration = std::time::Duration::from_secs(10);
+    const REPIN_ERROR_AFTER: std::time::Duration = std::time::Duration::from_secs(6);
+    const REPIN_SLOW_RETRY_AFTER: std::time::Duration = std::time::Duration::from_secs(30);
```
xv-thomas-leong: Introducing a new state machine makes sense here, but `lib.rs` is starting to feel a bit bloated. Could we extract this mechanism into its own file?

lightway-client/src/route_manager/repin.rs
```diff
@@ -40,6 +47,28 @@ impl RepinMode {
...
+                    tracing::error!(
+                        "Server route update failed ({error:?}) more than 6s, retrying"
+                    );
+                }
+                state.next_at = tokio::time::Instant::now()
+                    + if elapsed >= SLOW_RETRY_AFTER {
+                        LONG_INTERVAL
+                    } else {
+                        SHORT_INTERVAL
+                    };
```
xv-thomas-leong: should we do exponential retry here?

lightway-client/src/route_manager.rs
xv-thomas-leong, line 535: It seems the retry mechanism may not work the way we imagined.

If the server route is deleted but adding failed below (under `if let Some(old_route) = self.server_route.take()`), then a retry won't happen because `self.server_route` is already `None`.

lightway-client/src/route_manager/repin.rs
```diff
@@ -25,12 +25,19 @@ impl RepinState {
 pub enum RepinMode {
     /// Only re-pin when the gateway or interface index actually changed.
     OnRouteChange,
+    /// Always re-pin on every network event — handles within-subnet roaming on
+    /// Apple platforms where routing identifiers are identical but the physical
+    /// path has changed and the outside socket needs rebinding.
+    #[cfg(apple)]
```
xv-thomas-leong: Do we need a condition compile here? What's the harm of other platform having this mode?

kp-antonio-yang: Yes, we can do that, and expose this mode to config file.

Currently the mode is internal, so it is not configurable.  That is why platform gate on it.

lightway-client/src/route_manager.rs
kp-antonio-yang: Yes, I also can fix the other potential bug in the PR.

lightway-client/src/route_manager/repin.rs
xv-thomas-leong: I guess we can still hardcode a specific mode to a specific platform, but we don't have to platform gate it here

kp-mariappan-ramasamy: Yep agree, lets do platform / feature gate only when necessary

lightway-client/src/route_manager/repin.rs
```diff
@@ -0,0 +1,61 @@
...
+        let now = tokio::time::Instant::now();
+        Self {
+            started_at: now,
+            next_at: now,
+            nudge,
+            retry_count: 0,
+        }
+    }
+
+    pub fn elapsed_since_start(&self) -> std::time::Duration {
```
xv-thomas-leong, line 22: Is this function, along with `started_at` used anywhere?

lightway-client/src/route_manager/repin.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl RepinMode {
+    pub fn needs_repin(&self, route_changed: bool) -> bool {
+        match self {
+            RepinMode::OnRouteChange => route_changed,
+            RepinMode::Always => true,
+        }
+    }
+
+    pub fn on_failure(&self, state: &mut RepinState, error: &RoutingTableError) {
```
xv-thomas-leong: nit: `&self` not used

lightway-client/src/route_manager/repin.rs
kp-antonio-yang: Yes, this time should show in the log for debug.  :pray:

lightway-client/src/route_manager.rs
```diff
@@ -395,6 +409,18 @@ impl RouteManagerInner {
...
     }
 
+    /// Updates the server route by deleting the old one and adding a new one
+    async fn update_server_route(&mut self, new_route: Route) -> Result<(), RoutingTableError> {
+        if let Some(old_route) = self.server_route.clone()
+            && self.route_manager_async.delete(&old_route).await.is_ok()
+        {
+            self.server_route = None;
+        }
+        self.add_route(&new_route).await?;
```
xv-thomas-leong: It seems with this function, we might trigger another bug when `add_route` failed:

```
attempt 1:  find_best_default_route  OK
            if let Some(..)          ENTER
            delete(old)              OK   → server_route = None
            add_route(new)           ERR  → `?` returns, memory stays None
            → coordinator: Err ⇒ retry ✅

attempt 2:  find_best_default_route  OK
            if let Some(..) = None   SKIP  ⛔
            return Ok(false)
            → coordinator: Ok(_) ⇒ "success", drops retry state, parks
            → /32 gone permanently; outside traffic matches 0.0.0.0/1 via tun
            → io.reconnect() then pins the outside socket INTO the tunnel
```

<details>

<summary>Suggested fix</summary>

```rs
// pseudo
async fn check_and_update_server_route(&mut self) -> Result<bool, RoutingTableError> {
    match &self.server_route {
        // No pin at all: an earlier re-pin deleted the old one and failed to
        // install its replacement. `needs_repin` has nothing to compare against,
        // and a missing pin is never something to skip.
        None => {
            warn!("Server route missing - reinstalling with ({if_index}, {gateway})");
            // fall through to install
        }
        Some(server_route) => {
            let route_changed = server_route.gateway()  != current_gateway
                             || server_route.if_index() != current_if_index;
    
            if !self.repin_mode.needs_repin(route_changed) {
                return Ok(false);          // the ONLY "nothing to do" exit
            }
    
            // fall through to install
        }
    }
    
    // single install path, reached by both arms
    build new_server_route from current_gateway / current_if_index
    self.update_server_route(new_server_route).await?;
    info!("Updated server route for network change");
    Ok(true)
}
```

</details>

xv-thomas-leong (commented):
One more thing to fix, otherwise looks good

lightway-client/src/route_manager.rs
kp-antonio-yang: The suggestion applied.

The another issue we need to assume the server route add without error here (The add again did not work will be CCB, only add after delete works). 
So there is an Error `RoutingTableError::ServerRouteAddFailed` raised now.  And make sure the client can stop.

```
Network Event Occurs
         ↓
network_event_coordinator() loop (line 875)
         ↓
check_and_update_server_route() (line 926)
         ↓
update_server_route() (line 417)
         ↓
add_route() fails
         ↓
ERROR: ServerRouteAddFailed created (line 424)
    └─ "Server route add failed: {e:?}"
         ↓
Error propagates back to network_event_coordinator (line 926)
         ↓
Match on error (line 928-929)
         ↓
    ServerRouteAddFailed matched ✓
         ↓
Log error message (line 930-931)
    "Fatal error updating server route: {e}. Shutting down VPN."
         ↓
break; (line 933)
    └─ Exit the event loop
         ↓
network_event_coordinator() function ends
         ↓
RouteUpdater dropped
         ↓
Routes cleaned up (removed from system)
         ↓
STOPPED
```

lightway-client/src/route_manager.rs
```diff
@@ -511,41 +545,49 @@ impl RouteManagerInner {
...
-                self.add_route_server(new_server_route).await?;
+            self.update_server_route(new_server_route).await?;
 
-                tracing::info!("Updated server route for network change");
-                return Ok(true);
-            }
+            tracing::info!("Updated server route for network change");
+        } else {
+            warn!("Server route missing - reinstalling with {current_route:}");
+            self.add_route_server(current_route).await?;
```
xv-thomas-leong: `current_route` is the default route. Doing so will always fail because it already exists. Do you mean to do `add_route_server(new_server_route)`?

lightway-client/src/lib.rs
```diff
@@ -897,22 +902,49 @@ async fn network_event_coordinator(
...
+        };
 
         match route_updater.check_and_update_server_route().await {
+            Err(e) => {
+                match &e {
+                    route_manager::RoutingTableError::ServerRouteAddFailed(_) => {
+                        tracing::error!(
+                            "Fatal error updating server route: {e}. Shutting down VPN."
+                        );
+                        break;
```
xv-thomas-leong: The connection won't shut down here - the `break` will bring down this task only.

Also, by shutting down the connection upon server route add failure, we won't be retrying anymore. Is this intended?

lightway-client/src/route_manager.rs
kp-antonio-yang: Yes it is server_route, it defined by line 539, and just the variable named current_route.
 
`let current_route = self.find_best_default_route(&server_ip)?;`

When `check_and_update_server_route`, if it there is no `server_route` of `RouteManagerInner`, we should take this chance to added it back.  In general the `server_route` of  `RouteManagerInner` should be consist with the best_default_route. should not be None.

lightway-client/src/lib.rs
kp-antonio-yang: Yes, now by reusing the stop_signal, we can stop the service when fetal error happen.

I got cheated by AI https://github.com/expressvpn/lightway/pull/519#discussion_r3985412364, even I asked about did it really stop the service.

lightway-client/src/route_manager.rs
kp-antonio-yang: Yes, we should also raise a fatal error here, it should not be None, and it is nothing about self-healing.

PR #522, opened 2026-08-19
expresslane: fail closed under the all-zero degrade sentinel

kp-mariappan-ramasamy:
## Description

A degrade notice installs the all-zero sentinel key. promote_self_key and update_peer_key already clear has_valid_keys for it, but the data path never consulted those flags: encrypt_into/append_to_wire sealed under the sentinel and decrypt_into/try_from_wire opened under it, including via the prev_peer rotation-grace slot. A caller that does not gate every call on has_valid_keys - or that decrypts one-sided during the initial exchange, which is legitimate - would emit traffic under a publicly known key (effective plaintext) and accept attacker-forged frames.

Reject the sentinel inside the key lock in all four data paths, and skip a sentinel sitting in prev_peer. Checking key.is_invalid() under the same read guard that loads the cipher closes the check-then-use race a lock-free flag test would leave open.

## Motivation and Context

To avoid installing all 0 keys

## How Has This Been Tested?

Added unit tests

## Types of changes

## Checklist:

### DIFF
--- a/lightway-expresslane/src/session.rs
+++ b/lightway-expresslane/src/session.rs
@@ -283,6 +283,9 @@ impl<A: ExpresslaneAead> ExpresslaneSession<A> {
         let (cipher_text, auth_tag) = {
             let keys = self.keys_read();
             let current = keys.current_self.as_ref().ok_or(ExpresslaneError::NoKey)?;
+            if current.key.is_invalid() {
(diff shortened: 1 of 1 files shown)


kp-weiwen-goh (approved):
LGTM

PR #523, opened 2026-08-19
lightway-client: windows: modify wait timeout to 5 sec

kp-anuz-tomar:
## Description
Currently windows client wait indefinitely blocking for all process to terminate. However for some reason when the thread terminates, the waiting thread doesn't get the status update and the client hangs indefinitely.

We have modified this behaviour and we wait for 5 seconds and loop again. So if all threads are terminated by this time, we do not wait indefinitely

## How Has This Been Tested?
Tested locally on the windows vm

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


kp-antonio-yang (approved):
Thank you.

PR #524, opened 2026-08-20
[CVPN-2667] add option to set TUN interface txqueuelen

xv-piotr-haber:
## Description
Set default txqueuelen to 1000.
Bump tun-rs version.

## Motivation and Context
We want to set default txqueuelen on linux tun interface to 1000 to prevent packet drops on high-throughput tunnels.

## How Has This Been Tested?
Checked that txqueuelen is set accordingly when specified and defaults to 1000 when not.

## Types of changes

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ tokio-stream = "0.1.14"
 tokio-util = "0.7.10"
 tracing = "0.1.37"
 tracing-subscriber = "0.3.17"
-tun-rs = { version = "2.8.5", features = ["async"] }
+tun-rs = { version = "2.8.8", features = ["async"] }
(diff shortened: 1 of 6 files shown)


kp-antonio-yang (dismissed):
Nit: we could add a validation step for the user here:
https://github.com/expressvpn/lightway/blob/c22f534bc9902c6c049dd93191981d49e3ca5aad/lightway-client/src/config.rs#L431

At minimum, we can reject `0` to prevent behavior when set to `0` is undefined. Additionally, we could emit a warning if the value is lower than `1000`(lower than our default) to nudge users away from potentially problematic configurations.

kp-mariappan-ramasamy (approved):

PR #527, opened 2026-08-24
client: pin outside socket egress on Windows via `IP_UNICAST_IF`

kp-antonio-yang:
## Description

Pin the Windows outside socket to the physical egress interface using `IP_UNICAST_IF`/`IPV6_UNICAST_IF` so that routing-table changes (including the tunnel's own default route) cannot redirect outside traffic back into the tunnel. Re-pin on every network-change event and swallow transient send errors that occur mid-roam.  The encapsulation loop from outside packet back to tunnel cause issues in CVPN-2746, CVPN-2747 and CVPN-2748.  

This is possible to apply on Linux but won't do, because the roaming in `NoExec` mode from wifi only work in different subnet.  The deep reason for this is `IP_UNICAST_IF` is evaluated on every send call on Windows (Winsock), and it overrides routing table decisions per-packet, regardless of whether the socket has previously called `connect()`.  However, `IP_UNICAST_IF` sets `sk->sk_bound_dev_if`, only consulted during a route lookup on Linux. For a connected UDP socket, the route is cached in `sk_dst_cache` at `connect()` time, re-calling `setsockopt(IP_UNICAST_IF, same_index)` updates the flag but does not flush the route cache.  #492 still needed no mater we have have this feature on Linux or not.
 
### Change details

- **`platform/windows/egress.rs`** *(new)*: wraps `GetBestInterfaceEx` + `setsockopt(IP_UNICAST_IF/IPV6_UNICAST_IF)` into three helpers — `best_interface_index`, `set_unicast_if`, and `pin_to_peer_interface`.
- **`io/outside/udp.rs`**: pins the UDP socket at creation; stores the current `pinned_if_index` (AtomicU32) and implements `OutsideIO::pin_egress_interface` to re-pin on network change; swallows transient `WSAEADDRNOTAVAIL`/`WSAEHOSTUNREACH` send errors that arise mid-roam before the next send re-selects a source address.
- **`io/outside/tcp.rs`**: pins the TCP socket before `connect()` (the option has no effect afterwards, so this is the only opportunity).
- **`io/outside.rs`**: adds `OutsideIO::pin_egress_interface` (Windows only, default no-op).
- **`lib.rs`**: calls `pin_egress_interface` from `network_event_coordinator` after each successful route refresh; extends the `outside_io` weak reference from Apple-only to `any(apple, windows)`.
- **`route_manager.rs`**: adds `RouteUpdater::server_route_if_index()` to expose the interface index of the currently installed server route.

## Motivation and Context

On Windows without pinning, once the tunnel's default route is installed the OS may route outside packets back into the tunnel, causing a routing loop. Pinning the socket to the physical interface that currently reaches the server makes egress selection independent of the routing table. The pin is refreshed after each network-change event so that switching adapters (Wi-Fi → Ethernet, docking) is handled correctly.

## How Has This Been Tested?

- [X] Follow current CI testcases 
- [X] Manual Windows Wi-Fi roam test

## Types of changes

- [ ] Refactor (internal restructuring, no behaviour change for end users)
- [ ] CI / infrastructure change (no production code affected)

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -216,6 +216,13 @@ pub struct ClientConfig<ExtAppState: Send + Sync> {
     #[cfg(linux)]
     pub fwmark: u32,
 
+    /// Disable pinning the outside socket to the physical egress interface via
+    /// `IP_UNICAST_IF`/`IPV6_UNICAST_IF` (Windows desktop).
+    /// Pinning is on by default; set this to `true` only when external routing
+    /// policy handles egress selection.
+    #[cfg(windows)]
+    pub disable_pin_egress_interface: bool,
+
     /// DNS configuration mode
     #[cfg(desktop)]
     pub dns_config_mode: DnsConfigMode,
@@ -362,6 +369,8 @@ impl<ExtAppState: Send + Sync> ClientConfig<ExtAppState> {
             route_mode: config.route_mode,
             #[cfg(linux)]
             fwmark: config.fwmark,
+            #[cfg(windows)]
+            disable_pin_egress_interface: config.disable_pin_egress_interface,
             #[cfg(desktop)]
             dns_config_mode: config.dns_config_mode,
             enable_pmtud: config.enable_pmtud,
@@ -866,6 +875,7 @@ async fn network_event_coordinator(
     nudge_on_route_event: bool,
     #[cfg(apple)] nudge_on_route_update: bool,
     #[cfg(any(apple, windows))] outside_io: Weak<dyn OutsideIO>,
+    #[cfg(windows)] pin_egress_interface: bool,
     network_change_signal: mpsc::Sender<()>,
 ) {
     tracing::info!("Reacting to network change events...");
@@ -927,7 +937,8 @@ async fn network_event_coordinator(
         // index usually does not change (a Wi-Fi roam keeps the adapter), but
         // switching adapters entirely does change it.
         #[cfg(windows)]
-        if let Some(if_index) = route_updater.server_route_if_index()
+        if pin_egress_interface
+            && let Some(if_index) = route_updater.server_route_if_index()
             && let Some(io) = outside_io.upgrade()
         {
             io.pin_egress_interface(if_index);
@@ -1090,6 +1101,7 @@ impl<ExtAppState: Send + Sync> ClientConnection<ExtAppState> {
         transition_rx: Option<watch::Receiver<()>>,
         nudge_on_route_event: bool,
         #[cfg(apple)] nudge_on_route_update: bool,
+        #[cfg(windows)] pin_egress_interface: bool,
     ) -> Result<()> {
         let server_ip = self.outside_io.peer_addr().ip();
         let tun_index = self.inside_io.if_index()?;
@@ -1117,6 +1129,8 @@ impl<ExtAppState: Send + Sync> ClientConnection<ExtAppState> {
             nudge_on_route_update,
             #[cfg(any(apple, windows))]
             Arc::downgrade(&self.outside_io),
+            #[cfg(windows)]
+            pin_egress_interface,
             self.network_change_signal.clone(),
         )));
 
@@ -1836,6 +1850,8 @@ pub async fn client<
                 nudge_on_route_event,
                 #[cfg(apple)]
                 nudge_on_route_update,
+                #[cfg(windows)]
+                !config.disable_pin_egress_interface,
             )
             .await?;
     }
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -852,7 +852,8 @@ async fn handle_network_change<ExtAppState: Send + Sync>(
 
 /// Single consumer for network-change wake-ups. Reacts to each one with the
 /// steps in order: refresh the server route, re-`connect()` the outside
-/// socket (Apple platforms) and, when the wake-up represents a network transition,
+/// socket (Apple platforms), and re-pin its egress interface (Windows), when
+/// the wake-up represents a network transition,
 /// nudge the connection-level network-change handler. On Apple platforms the
 /// nudge also fires when the server route was actually replaced (Datagram
 /// wiring only). Owns the [`RouteUpdater`], so aborting the task removes the
@@ -864,7 +865,7 @@ async fn network_event_coordinator(
     mut transition_rx: Option<watch::Receiver<()>>,
     nudge_on_route_event: bool,
     #[cfg(apple)] nudge_on_route_update: bool,
-    #[cfg(apple)] outside_io: Weak<dyn OutsideIO>,
+    #[cfg(any(apple, windows))] outside_io: Weak<dyn OutsideIO>,
     network_change_signal: mpsc::Sender<()>,
 ) {
     tracing::info!("Reacting to network change events...");
@@ -920,6 +921,16 @@ async fn network_event_coordinator(
         #[cfg(apple)]
         if let Some(io) = outside_io.upgrade() {
             io.reconnect();
+        }
+
+        // Refresh the egress pin from the server route we just validated. The
+        // index usually does not change (a Wi-Fi roam keeps the adapter), but
+        // switching adapters entirely does change it.
+        #[cfg(windows)]
+        if let Some(if_index) = route_updater.server_route_if_index()
+            && let Some(io) = outside_io.upgrade()
+        {
+            io.pin_egress_interface(if_index);
         }
 
         if nudge && let Err(e) = network_change_signal.send(()).await {
@@ -1104,7 +1115,7 @@ impl<ExtAppState: Send + Sync> ClientConnection<ExtAppState> {
             nudge_on_route_event,
             #[cfg(apple)]
             nudge_on_route_update,
-            #[cfg(apple)]
+            #[cfg(any(apple, windows))]
             Arc::downgrade(&self.outside_io),
             self.network_change_signal.clone(),
         )));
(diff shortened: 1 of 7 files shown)


lightway-client/src/io/outside/tcp.rs
```diff
@@ -44,6 +44,24 @@ impl Tcp {
                     }
                 }
 
+                // the route lookup that binds the local address happens there,
+                // and IP_UNICAST_IF has no effect on an already-connected socket.
+                #[cfg(all(windows, not(feature = "mobile")))]
```
kp-mariappan-ramasamy: Why do we need to have !mobile ?
Is not windows only cover desktop clients ?

lightway-client/src/platform/linux/egress.rs
```diff
@@ -0,0 +1,277 @@
...
+//! egress to a single interface, preventing the tunnel's own routes from
+//! capturing outside traffic after they are installed.
+
+use std::io;
+use std::mem::size_of;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::os::fd::RawFd;
+
+/// Constrain `socket`'s egress to interface `if_index`.
+pub fn set_unicast_if(fd: RawFd, if_index: u32, ipv6: bool) -> io::Result<()> {
```
kp-mariappan-ramasamy: For sockopt, follow the existing apputils module, which has absrtactions for different os'es
https://github.com/expressvpn/lightway/blob/main/lightway-app-utils/src/sockopt/ip_mtu_discover.rs

lightway-client/src/io/outside/tcp.rs
```diff
@@ -13,7 +13,7 @@ impl Tcp {
         remote_addr: SocketAddr,
         maybe_sock: Option<TcpStream>,
         #[cfg(all(linux, not(feature = "mobile")))] fwmark: u32,
-        #[cfg(all(windows, not(feature = "mobile")))] pin_egress_interface: bool,
+        #[cfg(all(any(linux, windows), not(feature = "mobile")))] pin_egress_interface: bool,
```
kp-mariappan-ramasamy: i think i understood why you added !mobile with windows, to avoid android ??

I suggest to create more cfg in https://github.com/expressvpn/lightway/blob/ac8468c2eb812f6a72f72afb55a72be83f482986/lightway-client/build.rs#L7 like linux_desktop here and use it.

It will be clear and avoid mistakes.

lightway-client/src/io/outside/tcp.rs
```diff
@@ -44,6 +45,37 @@ impl Tcp {
                     }
                 }
 
+                // The route lookup that binds the local address happens at
+                // connect() and IP_UNICAST_IF has no effect on an already-connected
+                // socket, so pin must be applied before connect.
```
kp-mariappan-ramasamy, line 50: From windows documentation, i do not see the option should be set before connect
https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options

Can you add the reference to this claim ?
If this is not true, we can ignore setting this now, our route manager will set this when first default route is added.

lightway-client/src/platform/windows/egress.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use windows_sys::Win32::NetworkManagement::IpHelper::GetBestInterfaceEx;
+use windows_sys::Win32::Networking::WinSock::{
+    AF_INET, AF_INET6, IN_ADDR, IN_ADDR_0, IN6_ADDR, IN6_ADDR_0, IP_UNICAST_IF, IPPROTO_IP,
+    IPPROTO_IPV6, IPV6_UNICAST_IF, SOCKADDR, SOCKADDR_IN, SOCKADDR_IN6, SOCKADDR_IN6_0, SOCKET,
+    SOCKET_ERROR, setsockopt,
+};
+
+/// The interface index Windows would currently use to reach `dst`.
+pub fn best_interface_index(dst: IpAddr) -> io::Result<u32> {
```
kp-mariappan-ramasamy, line 23: We are trying to replicate the code from routw manager here
https://github.com/expressvpn/lightway/blob/main/lightway-client/src/route_manager.rs#L248

In worst case, both implementation deviates and server route is set with 1 interface and the socket is pinned to different interface and it might break the connection.

Better to use same implementation. 
Also if we confirm we can set UNICAST_IP after connection, we can get rid of this flow and pin egress only from route_manager.

lightway-client/src/io/outside/tcp.rs
kp-antonio-yang: If windows developer cross compiling the android, it might be a issue.  The mobile is not a platform gate,  we need use !mobile to ensure this will not go into mobile.

kp-mariappan-ramasamy: Per rust documentation, https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.target_os
target_os should be set to Android, if some dev compiles Android from Windows host os.

Are you seeing different behavior ?

kp-antonio-yang: The mobile is feature gate is not `target_os` to `android`.  We try to do it in #511, and then it introduced downstream issue with uniffi version confliction so we regated it in #534.

kp-antonio-yang: Agreed, it is not an issue from feature gate.

lightway-client/src/platform/windows/egress.rs
kp-antonio-yang: Yes, this restriction is documented only on Linux [here](https://github.com/ctrliq/kernel-src-tree/commit/0e4d354762cefd3e16b4cff8988ff276e45effc4)

> However, for connected sockets, the outbound interface
is chosen when connecting the socket, and in the route lookup process
which is done when a packet is being sent, the interface specified by
the IP_UNICAST_IF socket option is being ignored.

I want to restrict the implementation to windows only, since Linux can use fwmark to solve this issue.

lightway-client/src/lib.rs
```diff
@@ -207,6 +207,13 @@ pub struct ClientConfig<ExtAppState: Send + Sync> {
     #[cfg(linux)]
     pub fwmark: u32,
 
+    /// Disable pinning the outside socket to the physical egress interface via
+    /// `IP_UNICAST_IF`/`IPV6_UNICAST_IF` (Linux and Windows desktop).
+    /// Pinning is on by default; set this to `true` only when external routing
+    /// policy handles egress selection.
+    #[cfg(any(linux_desktop, windows_desktop))]
+    pub disable_pin_egress_interface: bool,
```
kp-mariappan-ramasamy, line 215: I think we dont need this flag. and can use `route_mode == noexec` to do the same ?

kp-antonio-yang: `pin_egress_interface` apply on `noexec` mode, it is not a new router mode.  We can remove this option, if it does not make any sense to turn off it.   I tend to keep it, just in cause some bug happened and we want to disabled it in quick.

lightway-client/src/platform/windows/egress.rs
kp-antonio-yang: It will be better not to merge these functions, it still different.

- `best_interface_index()` uses Windows' `GetBestInterfaceEx` API to query what **interface** Windows would use to reach any destination IP (more direct/accurate)
- find_best_default_route() queries the routing table for the default route with **the best metric**(higher level, more affected by metric calculations)

The using case and mechanism are mismatch, and we do not have Linux version for `best_interface_index` any more, so it is good to keep these function separated.

kp-antonio-yang:
Hi @kp-mariappan-ramasamy,
Could you help me to review this PR again? :pray: 

- The spec now shrink to windows, which have native API for the interface.
- I believe the gate issue handled in other PR will be better, because I will need to change `#[cfg(linux)]` to `#[cfg(linux_destktop)]` at the same time, and current PR is not linux related any more.

Thank you.

lightway-client/src/io/outside/tcp.rs
kp-mariappan-ramasamy: Reopened the comment. This is not resolved yet.

```
IP_UNICAST_IF has no effect on an already-connected
```

still is not correct. Windows documentation does not mention any

lightway-client/src/platform/windows/egress.rs
kp-mariappan-ramasamy: Yes that is correct. It is possible both values are mismatched,

Lets say there are 2 interfaces with defaut addresses and above mismatch happens.
In that case, we have route table which is pointing to one interface and socket which is pinned to different interface, which makes the traffic to fail cause it might be in different networks.

Which is why we need to use the same mechanism to pin and add routes. 

> we do not have Linux version for best_interface_index any more, so it is good to keep these function separated

What is linux has to do with separating the functions in windows ?

I don't understand why separating is better in this situation. I believe it will break the client if there is a mismatch

kp-antonio-yang: Linux eventually so_mark in #492, and no `IP_UNICAST_IF`, because the flag is intrinsic different on Windows and on Linux, and we can not use it to solve Linux issues easily.

In previous time, we tired using `IP_UNICAST_IF` (same behavior with windows) and also repin it (same behavior with macos).
However, this solution can not solve same subnet issues, unless we do extra monitoring on layer 2 events.

lightway-client/src/lib.rs
kp-mariappan-ramasamy: I see, make sense.

kp-antonio-yang:
@kp-mariappan-ramasamy 
You are correct, we do not need to pin the socket before connection.  :pray:   The restriction of Windows is less, we only need pin it when network switching, and the solution can be in a simple way.

PR #531, opened 2026-08-27
CVPN-2754: Support IPv6 outside socket in ip_mtu_discover

kp-thomas-yau:
## Description

Return IPv6 level and optname for Apple platform if the socket is using IPv6.

## Motivation and Context

Right now if we're using IPv6 socket on Apple devices, getsockopt(IPPROTO_IP, IP_DONTFRAG) will fail as it only works with IPv4 sockets.

## How Has This Been Tested?

Added unit tests to run on all platforms to verify getsockopt won't fail when we're using IPv6 sockets

## Types of changes

## Checklist:

### DIFF
--- a/lightway-app-utils/src/sockopt/ip_mtu_discover.rs
+++ b/lightway-app-utils/src/sockopt/ip_mtu_discover.rs
@@ -10,8 +10,7 @@ use std::mem::MaybeUninit;
 use libc::socklen_t;
 
 #[cfg(unix)]
-use std::os::fd::{AsRawFd, RawFd};
-
+use std::os::fd::{AsFd, AsRawFd, RawFd};
 #[cfg(windows)]
 use std::os::windows::io::AsRawSocket;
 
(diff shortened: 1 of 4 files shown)


lightway-app-utils/src/sockopt/ip_mtu_discover.rs
```diff
@@ -164,30 +164,45 @@ mod internal {
...
+
+/// On Apple platforms the don't-fragment option lives at a different
+/// level/name for IPv4 and IPv6 sockets, so inspect the socket's
+/// address family to pick the right pair.
+#[cfg(apple)]
+fn get_level_and_optname(sock: &impl AsGenericHandle) -> std::io::Result<(i32, i32)> {
+    // SAFETY: the fd belongs to the socket our caller holds a borrow of,
+    // so it remains open for the duration of this call
+    let fd = unsafe { std::os::fd::BorrowedFd::borrow_raw(sock.as_generic_handle()) };
+    let addr = socket2::SockRef::from(&fd).local_addr()?;
```
xv-thomas-leong: If we modify `AsGenericHandle` into impl<T: **AsFd**> AsGenericHandle instead of impl<T: **AsRawFd**> AsGenericHandle

Then we might be able to do 

```suggestion
    let addr = socket2::SockRef::from(sock).local_addr()?;
```

, without needing to introduce unsafe

xv-thomas-leong (dismissed):

lightway-app-utils/src/sockopt/ip_mtu_discover.rs
kp-thomas-yau: Done^

xv-thomas-leong (approved):

PR #536, opened 2026-09-01
tests: add cases for invalid, expired or corrupted certs

kp-ben-hung:
## Description

Add negative tests for certs validation. Negative tests include corrupt, expired and invalid certs, on root intermediate and leaf.

Note: These tests are currently disabled on RISCV, as generating certs on RISCV runner is taking too long (sometimes over 13 extra minutes). Since these tests already run on other platforms, it is not worth it to run them on RISCV now.

## Motivation and Context

Ensure certs validate correctly.

## How Has This Been Tested?

CI passing.

## Types of changes

## Checklist:

### DIFF
--- a/lightway-core/tests/certs.rs
+++ b/lightway-core/tests/certs.rs
@@ -1,4 +1,7 @@
 //! Certificate-chain validation tests for backend.
+//!
+//! On-the-fly generation of certs is too slow on RISCV, skip this suite.
+#![cfg(not(target_arch = "riscv64"))]
 
 use std::sync::Arc;
 use std::time::Duration;
--- a/lightway-core/tests/certs.rs
+++ b/lightway-core/tests/certs.rs
@@ -1,0 +1,83 @@
+//! Certificate-chain validation tests for backend.
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use test_case::test_matrix;
+use tokio::net::UnixStream;
+
+pub mod common;
+use crate::common::certgen::{Defect, Link, TestPki};
+use crate::common::connection::{PQCrypto, TestAuth, TestSock, TestStreamSock, client, server};
+use crate::common::get_test_timeout;
+use rcgen::RsaKeySize;
+
+async fn handshake(pki: &'static TestPki) -> Result<(), String> {
+    let attempt = tokio::spawn(async move {
+        let (client_sock, server_sock) = UnixStream::pair().expect("UnixStream");
+        let client_sock = Arc::new(TestStreamSock(client_sock));
+        let server_sock = Arc::new(TestStreamSock(server_sock));
+        let _ = client_sock.writable().await;
+
+        let auth = Arc::new(TestAuth::default());
+        let pqc = PQCrypto::default();
+        let (cert, key) = pki.server_secrets();
+        tokio::join!(
+            server(server_sock, auth, pqc, None, None, None, cert, key,),
+            client(
+                client_sock,
+                None,
+                pqc,
+                None,
+                false,
+                false,
+                false,
+                pki.root_ca(),
+            )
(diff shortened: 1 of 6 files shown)


lightway-core/tests/certs.rs
```diff
@@ -0,0 +1,194 @@
...
+                trust_bundle_pem: pem_bundle(&[&root.cert_der(), &intermediate.cert_der()]),
+                leaf: intermediate.issue_server_with(TEST_SERVER_DOMAIN, key_size, false),
+            }
+        }
+        _ => unreachable!("unsupported chain length"),
+    }
+}
+
+/// A PKI in which the certificate at `link` carries `defect`.
+fn defective_pki(defect: Defect, link: Link, key_size: RsaKeySize) -> Pki {
```
kp-antonio-yang: nit:
Could we do with `impl`?  Such that the constructors are grouped and easy know for reader.

```rust
Struct Pki {
}

impl Pki {
  pub new(chain_len: usize, key_size: RsaKeySize) -> Pki {}
  pub new_with_defect(defect: Defect, link: Link, key_size: RsaKeySize) -> Pki {}
}
```

lightway-core/tests/certs.rs
```diff
@@ -0,0 +1,194 @@
...
+    Corrupt,
+    Invalid,
+    Expired,
+}
+
+/// Everything one handshake attempt needs.
+struct Pki {
+    trust_bundle_pem: Vec<u8>,
+    leaf: CertifiedKey,
+}
```
kp-antonio-yang: Is it possible to reuse or extend the the `common::certgen::TestPki`?

At least we can do with enum, such that the reader can easy to know why we need more `TestPki`

```rust
enum TestPki {
   // A static shared test PKI before the timeout window RSA keygen is very slow on QEMU. (on RISCV).
   NoPem(StaticPki),
   // With chain details for hand shack testing 
   WithPem(FullChainPki) 
}

impl TestPki {

   // constructors
   pub fn new(chain_len: usize, key_size: RsaKeySize) -> TestPki::WithPem(FullChainPki),
   pub fn new_with_defect(defect: Defect, link: Link, key_size: RsaKeySize) -> TestPki::WithPem(FullChainPki),
   pub fn new_simple(key_size: RsaKeySize) -> TestPki::NoPem(StaticPki),

   //  methods, we can use `panic!` or `todo!` if some arms not really needed for the test case now.
   pub fn server_secrets(&self) 
   pub fn root_ca(&self) 

}
```

It will be better than `default_*` for readers.
How do you think about this?

kp-antonio-yang (changes requested):
A quick review, not complete read through all the commits.

lightway-core/tests/certs.rs
kp-ben-hung: First of all thanks a lot of the suggestion. I agree things could be improved in terms of readability. The reason i put `default_*` functions instead of an enum, is the `default_*` functions also handle caching. (making 10 different valid cert chain should NOT cause 10 keygens if the parameters are the same, just reuse the last one). This is implemented by having a global, static `LazyLock`. 

I also agree that the `TestPki` struct can actually be reused more, and the defect generation logic could be put into certgen.rs too (so other tests can reuse them).

How about we making a private struct `PkiSpec` to define what Pki we generate (the struct implements `Hash` so it could be put into a `HashMap` and cached). Then impl the `TestPki` with just 2 pub func: `get_valid` and `get_defective`. Future tests just need to put `let pki = TestPki::get_valid(2, RsaKeySize::_2048);` and it's done (with auto caching too).
This will be much more readable. See my latest change to see if this works better.

kp-antonio-yang (approved):
Looks good for me.

PR #541, opened 2026-09-02
CVPN-2761: Make OutsideIO injectable in server, client

kp-weiwen-goh:
## Description

Makes outside IO injectable on both the server and the client, the way inside IO already is. An application supplies a datagram transport; lightway keeps the receive loop and all of the dispatch, so a transport never touches ConnectionManager.

ServerConfig::outside_io and ClientConnectionConfig::outside_io take an OutsideIO: recv, recv_many (defaults to one packet, so a transport without a batch syscall implements recv alone), send_callback, send_unconnected, plus a RecvMeta { peer, local }. On the server UdpServer splits into UdpIo (the socket, still private) and DatagramServer (the loop).

## Motivation and Context

Inside IO has been injectable for a while, so a tun replacement can live outside this repository. Outside IO could not: UdpServer owned both the socket and the receive loop, and the packet handling was inlined into it, so a new transport meant either adding it here or duplicating the loop, the dispatch and the roaming rules.

## How Has This Been Tested?

Tested locally and with CI

## Types of changes

## Checklist:

### DIFF
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -130,8 +130,9 @@ pub enum LightwayError {
 /// to the live socket without reopening it.
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectionInfo {
-    /// The underlying socket, tagged with its transport type.
-    pub socket: io::outside::OutsideSocket,
+    /// The underlying socket, tagged with its transport type. `None` when
+    /// the outside IO has no single OS socket to name.
+    pub socket: Option<io::outside::OutsideSocket>,
     /// Remote peer address the connection is established to.
     pub peer_addr: std::net::SocketAddr,
 }
--- a/lightway-client/src/lib.rs
+++ b/lightway-client/src/lib.rs
@@ -78,6 +78,12 @@ use tracing::info;
 pub enum ClientConnectionMode {
     Stream(Option<TcpStream>),
     Datagram(Option<UdpSocket>),
+    /// A datagram transport supplied by the application, in place of the
+    /// socket the client otherwise creates. Every socket option in the
+    /// client config goes unused.
+    DatagramIo(Arc<dyn io::outside::OutsideIO>),
+    /// A stream transport supplied by the application; see `DatagramIo`.
+    StreamIo(Arc<dyn io::outside::OutsideIO>),
 }
 
 impl std::fmt::Debug for ClientConnectionMode {
@@ -85,6 +91,8 @@ impl std::fmt::Debug for ClientConnectionMode {
         match self {
             Self::Stream(_) => f.debug_tuple("Stream").finish(),
             Self::Datagram(_) => f.debug_tuple("Datagram").finish(),
+            Self::DatagramIo(_) => f.debug_tuple("DatagramIo").finish(),
+            Self::StreamIo(_) => f.debug_tuple("StreamIo").finish(),
         }
     }
 }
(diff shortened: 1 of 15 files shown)


lightway-server/src/io/outside/udp.rs
```diff
@@ -428,40 +448,32 @@ fn read_single_from_socket(
...
-    Ok((peer_addr, local_addr, reply_pktinfo))
-}
-
-/// Per-packet metadata produced by batched receive.
-struct BatchRecvMetadata {
-    peer: SocketAddr,
-    local: SocketAddr,
-    /// The `in_pktinfo` to echo back on replies, when the bind mode needs it.
-    reply_pktinfo: Option<libc::in_pktinfo>,
+    Ok((peer_addr, local_addr))
```
kp-mariappan-ramasamy: Could have removed just the reply_pktinfo and leave BatchRecvMetadata as it is ?
easier to read that (socketaddr, socketaddr)

lightway-server/src/io/outside/udp.rs
```diff
@@ -511,8 +502,96 @@ fn read_multiple_from_socket(
...
+                .unwrap();
+
+        assert_eq!(&reply[..len], b"rejected");
+        assert_eq!(
+            from, server_addr,
+            "reply must come from 127.0.0.2, which only an applied IP_PKTINFO achieves"
+        );
+    }
 }
+// Tests END -> panic, unwrap, expect allowed
```
kp-mariappan-ramasamy: We dont use this convention in our repo. lets remove it

lightway-server/src/lib.rs
```diff
@@ -164,6 +165,18 @@ pub struct ServerConfig<SA: for<'a> ServerAuth<AuthState<'a>>> {
...
+    ///
+    /// Supplying it replaces the socket the server would have created, so
+    /// every socket option on this struct goes unused, `enable_batch_send`
+    /// included: the batched inside loop needs the UDP send queue that
+    /// only the built-in socket provides.
+    ///
+    /// Datagram connections only, and a pre-bound socket in `mode` is
+    /// rejected: only one of the two can be honoured.
+    #[educe(Debug(ignore))]
+    pub outside_io: Option<Box<dyn OutsideIO>>,
```
kp-mariappan-ramasamy: I think instead of separate config, we could consider using ??
ServerConnectionMode::DatagramExt<Box<dyn OutsideIO>>

lightway-client/src/lib.rs
```diff
@@ -380,6 +380,17 @@ pub struct ClientConnectionConfig<EventHandler: 'static + Send + EventCallback>
...
+    /// Alternate Outside IO to use.
+    ///
+    /// Supplying it replaces the socket the client would have created, so
+    /// every socket option on this struct goes unused.
+    ///
+    /// `mode` still selects [`ConnectionType`], so it must match this
+    /// transport, and a pre-bound socket in `mode` is rejected: only one
+    /// of the two can be honoured.
+    #[educe(Debug(ignore))]
+    pub outside_io: Option<Arc<dyn io::outside::OutsideIO>>,
```
kp-mariappan-ramasamy: Same as server comment

lightway-server/src/io/outside/udp.rs
```diff
@@ -215,7 +213,7 @@ impl UdpServer {
             if lightway_app_utils::recvmsg_x::is_batch_receive_available() {
                 true
             } else {
-                warn!(
+                tracing::warn!(
```
kp-mariappan-ramasamy, line 224: this is an unnecessary churn

kp-weiwen-goh: `tracing::warn` is only used in `#cfg(macos)` now, so it is removed from top level import.

lightway-server/src/io/outside/udp.rs
kp-weiwen-goh: Added commit to remove all such comments

kp-mariappan-ramasamy (approved):
LGTM

PR #543, opened 2026-09-03
CVPN-2583: server: bound PROXY protocol header reads with a timeout

kp-mariappan-ramasamy:
## Description

handle_proxy_protocol() awaited both read_exact() calls with no deadline. Those reads run before create_streaming_connection(), so a socket parked there is never covered by the 60s stale-connection reaper. An unauthenticated peer could open thousands of TCP sessions, send one byte each and pin a tokio task plus an fd per socket indefinitely.

Wrap both reads in tokio::time::timeout and drop the socket if the header does not arrive within 5 seconds.

## Motivation and Context

Bug found by a security researcher.

## How Has This Been Tested?

Added unit test to cover the case

## Types of changes

## Checklist:

### DIFF
--- a/lightway-server/src/io/outside/tcp.rs
+++ b/lightway-server/src/io/outside/tcp.rs
@@ -40,6 +40,24 @@ impl OutsideIOSendCallback for TcpStream {
     }
 }
 
+/// The PROXY header is written by the proxy as soon as it accepts, so this
+/// only needs to cover the round trip. A socket which does not complete the
+/// header in time is dropped: these reads happen before the connection (and
+/// therefore the stale connection reaper) exists.
+const PROXY_PROTOCOL_READ_TIMEOUT: Duration = Duration::from_secs(5);
+
+async fn read_exact_with_timeout(
+    sock: &mut tokio::net::TcpStream,
+    buf: &mut [u8],
+    ctx: &'static str,
+) -> Result<()> {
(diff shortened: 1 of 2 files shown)


lightway-server/src/io/outside/tcp.rs
```diff
@@ -40,16 +40,37 @@ impl OutsideIOSendCallback for TcpStream {
     }
 }
 
+/// The PROXY header is written by the proxy as soon as it accepts, so this
+/// only needs to cover the round trip. A socket which does not complete the
+/// header in time is dropped: these reads happen before the connection (and
+/// therefore the stale connection reaper) exists.
+const PROXY_PROTOCOL_READ_TIMEOUT: Duration = Duration::from_secs(5);
```
kp-weiwen-goh, line 47: Should this be configurable in config?

kp-mariappan-ramasamy: This isn't a new feature, it's the fix for a pre-auth DoS. The change adds no config surface and no behaviour change for a well-behaved proxy, which writes the PROXY header immediately on accept. The only sockets affected are the ones that would otherwise hang forever. So I'd rather keep it as a default-on 5s deadline rather than push as a feature.

goweiwen (approved):
LGTM

kp-weiwen-goh (approved):
LGTM

PR #545, opened 2026-09-07
core: allow explicit outside MTU for datagram carriers

kp-weiwen-goh:
## Description

A datagram connection sends each inside packet in one outside datagram, so the outside MTU sets a
limit on the inside MTU. This change makes both sides obey that limit:

- `ServerConnectionBuilder::with_outside_mtu()` sets the outside MTU for a carrier whose
  per-frame budget is less than the wire MTU. `accept()` does a range check on it, and the DTLS
  record size follows it.
- `Connection::advertised_inside_mtu()` decreases the advertised inside MTU to the value that
  the outside path can send. This decrease is a ceiling only. A stream carrier, and a datagram
  carrier that already fits, keep their value.
- `ClientConnectionBuilder::with_strict_mtu()` marks a carrier that cannot use IP fragmentation.
  Such a carrier stops with an error on an MTU that it cannot send.
- Error `ConnectionError::PathMtuDiscoveryRequired` becomes `OutsideMtuTooSmallForInsideMtu`,
  and it gives the actual outside MTU next to the required one.
- Helpers `max_dtls_mtu()` and `max_dtls_outside_mtu()` now saturate. An outside MTU of less
  than 81 bytes went below zero before, and gave a value near `usize::MAX`.

`docs/pmtu_discovery.md` explains the limit and calls `MIN_INSIDE_MTU` advisory.

## Motivation and Context

The server advertised the raw MTU of the inside IO. With the Tun driver and no explicit MTU, the TUN
keeps the kernel default of 1500 bytes. That value needs an outside MTU of 1584 bytes, which is more
than `MAX_OUTSIDE_MTU`, so the client stopped right after authentication. A deployment that configures an
explicit inside MTU small enough for its outside path is not affected.

A correction for this existed on an earlier branch, but it never came to `main`. This is recovered
work, not a new defect.

## How Has This Been Tested?

Environment: Linux x86_64, `nix develop .#default`.

| Command                                                 | Result               |
| ------------------------------------------------------- | -------------------- |
| `cargo nextest run -p lightway-core --lib`              | 507 passed, 0 failed |
| `cargo nextest run -p lightway-core --tests`            | 570 passed, 0 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` | Clean                |

New tests: three unit tests for `advertised_inside_mtu()`, and four integration tests for the
handshake, the range check, the lenient path, and the strict path. The test harness gets a TUN MTU, a
PMTUD switch, a strict-MTU switch, and a sink for a fatal client error.

## Types of changes

Downstream consumers `lightway-core`:

- The renamed error variant has a new `outside_mtu` field. No crate in this workspace uses the old
  name.
- If its outside path is too small, a datagram server now advertises a decreased inside MTU.

## Checklist:

### DIFF
--- a/lightway-core/src/connection.rs
+++ b/lightway-core/src/connection.rs
@@ -208,15 +208,20 @@ pub enum ConnectionError {
     #[error("Invalid inside packet: {0}")]
     InvalidInsidePacket(InvalidPacketError),
 
-    /// Invalid Outside MTU
+    /// The outside MTU cannot carry the inside MTU the peer advertised, and
+    /// this connection has no way to shrink the gap: either PMTUD is running
+    /// (so the advertised MTU was never achievable) or the carrier is
+    /// strict-MTU and cannot fall back on IP fragmentation.
     #[error(
-        "PMTUD required: inside MTU {inside_mtu} needs at least {required_outside_mtu} outside MTU"
+        "inside MTU {inside_mtu} requires at least {required_outside_mtu} outside MTU, but outside MTU is {outside_mtu}"
     )]
-    PathMtuDiscoveryRequired {
+    OutsideMtuTooSmallForInsideMtu {
         /// The inside MTU
         inside_mtu: usize,
         /// The required outside MTU to support the inside MTU without PMTUD
         required_outside_mtu: usize,
+        /// The actual outside MTU the connection was built with
+        outside_mtu: usize,
     },
 
     /// Plugin returns a reply packet
@@ -287,7 +292,7 @@ impl ConnectionError {
                     PacketCodecDoesNotExist => true,
                     PacketCodecError(_) => true,
                     Disconnected => true,
-                    PathMtuDiscoveryRequired { .. } => true,
+                    OutsideMtuTooSmallForInsideMtu { .. } => true,
                     Tls(crate::tls::Error::Fatal(ErrorKind::DomainNameMismatch)) => true,
                     Tls(crate::tls::Error::Fatal(ErrorKind::DuplicateMessage)) => true,
                     Tls(crate::tls::Error::Fatal(ErrorKind::PeerClosed)) => true,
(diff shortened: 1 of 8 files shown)


lightway-core/src/connection.rs
```diff
@@ -2464,6 +2479,36 @@ impl<AppState: Send> Connection<AppState> {
...
+    fn advertised_inside_mtu(
+        connection_type: ConnectionType,
+        outside_mtu: usize,
+        inside_mtu: usize,
+    ) -> usize {
+        if !connection_type.is_datagram() {
+            return inside_mtu;
+        }
+
+        let advertised = max_dtls_mtu(outside_mtu).min(inside_mtu);
```
xv-thomas-leong: It seems inside of DTLS we still have the `Data` frame, we will have to account for that as well:

```suggestion
        let advertised = wire::Data::maximum_packet_size_for_plpmtu(max_dtls_mtu(outside_mtu))
        .min(inside_mtu);
```

kp-weiwen-goh: Good catch. `max_dtls_mtu` is the DTLS record payload, and the record still carries a `Data` frame around the inside packet. The frame costs 3 bytes, so the advertisement was 3 bytes above what one outside datagram can carry.

Applied in db151ce0d. The clamp now uses `Data::maximum_packet_size_for_plpmtu`, the same formula DPLPMTUD applies to its converged estimate. A full 1500-byte path now advertises 1416, which equals PMTUD's maximum data size.

The fix needed a second half. `dtls_required_outside_mtu` gains the same 3 bytes, so it stays the exact inverse of the clamp. Without that, the strict-MTU check stays 3 bytes lenient. `accept()` would also admit an outside MTU of 82 to 84. There the clamp panics on the `plpmtu too small` debug assert in debug builds. In release it wraps to a near-`usize::MAX` budget. The datagram floor is now 85: one inside byte in one datagram.

PR #553, opened 2026-09-17
Disable IPv4 DAD on Windows tunnel adapters

kp-john-mair:
While investigating slow Lightway connection and DNS readiness in Qt desktop on Windows, we saw the app report Connected before the tunnel address could be used for local binds. The roughly three-second delay before its local route appeared is consistent with IPv4 duplicate address detection (DAD). We have confirmed DAD was the cause, and with this change (tested locally) the delays have been eliminated.

This disables DAD on the tunnel adapter before assigning its IPv4 address, to avoid the initial DAD wait. [KDT-4475](https://polymoon.atlassian.net/browse/KDT-4475)

* keeps the existing interface settings and skips the update if DAD is already disabled
* logs a warning and carries on if it fails
* may also help with early DNS requests, but we haven't confirmed it resolves the longer DNS stalls we saw

WireGuard on Windows also [disables DAD on its tunnel interface](https://git.zx2c4.com/wireguard-windows/tree/tunnel/addressconfig.go#n137) with `DadTransmits = 0`. That's the same Windows setting used here.

Feel free to reject this PR or rewrite it to better suit your style. Happy for this to just serve as a starting point.

[KDT-4475]: https://polymoon.atlassian.net/browse/KDT-4475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kp-mariappan-ramasamy:
Thanks for the PR @kp-john-mair 

Changes looks good. 

One question.
Did you verify whether disabling dad fixes the tunnel readiness immediately after connected ? Not clear from description.

kp-john-mair:
> Thanks for the PR @kp-john-mair
> 
> Changes looks good.
> 
> One question. Did you verify whether disabling dad fixes the tunnel readiness immediately after connected ? Not clear from description.

Yes, it helps *dramatically*. Thanks!

kp-mariappan-ramasamy (approved):
LGTM, thanks

kp-john-mair:
Thanks for the approval @kp-mariappan-ramasamy!  FYI i can't merge this  (there's no button in the UI for me)

PR #555, opened 2026-09-18
ci: group the workflows and shrink the nix build matrix

kp-mariappan-ramasamy:
## Description

Both CI and nix-build workflows now delegate to reusable workflows, so GitHub collapses their jobs under one node per area and names them `<area> / <job>`.

- `Nix Build and Test` → 3 groups: `x86_64-linux`, `aarch64-linux`, `macos`
- `CI` → `build-linux`, `e2e`, `nix`, `windows`, plus coverage and the gate

Alongside the grouping, the nix matrix does less work:

| | before | after |
|---|---|---|
| Nix Build jobs | 32 | 15 |
| Nix Build wall clock | 59m | 23m |
| Nix Build compute | 440 job-min | 160 job-min |
| CI top-level rows | 32 | 7 |

- **Client and server build together.** The flake already exposed combined `lightway-<target>` packages that compile the shared dependency graph once. Using them halves the matrix, and each job is barely slower than building a single binary was.
- **The flake check runs once per platform**, not once per build job. It only evaluates the current runner's outputs, so it cannot collapse further without losing aarch64 and Darwin coverage.
- **The macOS cross builds moved to a weekly run** (`nix-cross-macos.yaml`).
- **Build jobs have a 90 minute timeout.**

## Motivation and Context

32 flat jobs in creation order, with the runner platform buried at the end of each name, made it hard to see what failed.

The nightly build also did unnecessary work: every target was built twice, once per binary, recompiling the shared dependency tree each time. Wall clock was set by macOS runner concurrency, where 12 jobs contended for 2-3 runners and queued up to 43 minutes. Four of those were cross-compiles that had never failed in 100 runs and whose targets are already built and verified on Linux the same day.

## How Has This Been Tested?

Full green runs on both workflows on this branch:

- `Nix Build and Test`: 16/16 jobs green, 23m
- `CI`: 31/31 green (1 expected skip), 21m

## Types of changes

## Checklist:


kp-mariappan-ramasamy:
**Nix Build and Test** — 32 jobs → 15, 59m → 23m, 440 → 160 job-min
- before: https://github.com/expressvpn/lightway/actions/runs/35069931924
- after:  https://github.com/expressvpn/lightway/actions/runs/35319815073

**CI** — 32 flat rows → 7 groups; compute unchanged (302 → 303 job-min), 23m → 21m
- before: https://github.com/expressvpn/lightway/actions/runs/35178576584
- after:  https://github.com/expressvpn/lightway/actions/runs/35544195688

kp-ben-hung (approved):
LGTM, manually verified the artifacts names actually correspond to the artifacts and nothing is accidentally removed.

